PR #1, opened 2016-11-11
Updates documentation: s/civis-client/civis-python

jjlueck-tock:
The repository name is incorrect, methinks.


keithing:
Yup, that's a typo.  Thanks for the fix!  LGTM.

PR #2, opened 2016-11-14
fixing a small typo in the user_guide

keithing:
Found a small error in the user_guide documentation.  This fixes that error.

In the process of this PR, a new version of `flake8` came out which caused the code to fail tests.  Those changes were addressed in #3.  However, that prompted me to pin the versions of dev requirements (which is also included in this PR).


keithing:
@elliothevel, care to give this a once over?

elliothevel:
Good catch. LGTM.

PR #3, opened 2016-11-15
MAINT Be less strict with requirements

stephen-hoover:
We don't need to pin to specific minor and micro versions of `requests` and `pyyaml`. Tell pip that it's okay to have newer versions of those libraries with the same major version.


stephen-hoover:
@keithing , how does this look?

keithing:
Thanks for changes and the style fix. The newest version of `flake8` is flagging some style issues that the previous version wasn't. I'm pinning the version in #2, but I think in the spirit of this PR I'll relax the pin to a major version rather than a specific version.

LGTM.

PR #4, opened 2016-11-16
Use 'all' resources when generating local docs

jcobian:
This will ensure the local documentation generated includes all the resources the current user has access to


keithing:
Great! Thanks for this.  LGTM.

PR #5, opened 2016-11-22
Adding "drop" option in docs for table uploads

keithing:
`drop` is an option in `client.imports.post_files` but wasn't explicitly listed in `dataframe_to_civis` or `csv_to_civis`.  This adds that documentation.


beckermr:
LGTM!

PR #7, opened 2016-12-05
Avoiding roundtrip to disk in read_civis_sql

### LINKED ISSUE #6: Speed up read_civis by avoiding round trip to disk
Initially pointed out by @cmmalone.  Let's try to speed up `read_civis` by avoiding a write to disk followed by a read from disk.

keithing:
Resolves #6.  This should be slightly faster since we'll be reading directly from memory rather than from disk.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1,16 +1,21 @@
 import csv
 import tempfile
 
+from civis import APIClient
+from civis._utils import maybe_get_random_name
+from civis.base import EmptyResultError
+from civis.polling import PollableResult, _DEFAULT_POLLING_INTERVAL
+import requests
+
+try:
+    from io import StringIO
+except ImportError:
+    from cStringIO import StringIO
 try:
     import pandas as pd
     NO_PANDAS = False
 except ImportError:
     NO_PANDAS = True
-import requests
-
-from civis import APIClient
-from civis._utils import maybe_get_random_name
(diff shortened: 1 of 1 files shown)


civis/io/_tables.py
```diff
@@ -172,17 +177,28 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
...
+                          (script_id, run_id),
+                          polling_interval)
+    if archive:
 
+        def f(x):
+            return client.scripts.put_sql_archive(script_id, True)
+
+        poll.add_done_callback(f)
+    poll.result()
+    url = client.scripts.get_sql_runs(script_id, run_id)["output"][0]["path"]
```
elliothevel: This line will raise an `IndexError` if the script doesn't return any rows. (`client.scripts.get_sql_runs(script_id, run_id)["output"]` will be an empty list). What about something like:
```
outputs = client.scripts.get_sql_runs(script_id, run_id)["output"]
if not outputs:
    raise EmptyResultError
url = outputs[0]["path"]
```

keithing: Good catch.  I like your solution. Implemented here: https://github.com/civisanalytics/civis-python/pull/7/commits/dd8e40fcfa5ee878bfdc02840ce151c85d2c0552

elliothevel:
LGTM!

PR #8, opened 2016-12-06
fix windows io issue in NamedTemporaryFile

knanne:
Proposing fix for bug in io operations on Windows platform. Currently windows throws `PermissionError: [Errno 13]` on temp file when using `civis.io.read_civis()`, `civis.io.read_civis_sql()` and `civis.io.dataframe_to_civis()`

According to Python docs for [`NamedTemporaryFile()`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile)
> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later). If delete is true (the default), the file is deleted as soon as it is closed.

Passing `delete=False` seems to fix the issue.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -172,7 +172,7 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
     """
     if use_pandas and NO_PANDAS:
         raise ImportError("use_pandas is True but pandas is not installed.")
-    with tempfile.NamedTemporaryFile(mode="w+") as f:
+    with tempfile.NamedTemporaryFile(mode="w+", delete=False) as f:
         csv_poll = civis_to_csv(f.name, sql=sql, database=database,
                                 job_name=job_name, credential_id=credential_id,
                                 polling_interval=polling_interval,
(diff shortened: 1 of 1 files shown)


keithing:
Thanks for pointing this out. I think this PR has a merge conflict due to #7.  I very recently finished a small optimization in that PR to avoid writing to disk when using `read_civis_sql` (and by extension `read_civis`).  That change removed one of the `NamedTemporaryFile`s you're fixing here, so #7 is in effect solving part of this issue in a different way. 

 @elliothevel is working on removing the other `NamedTemporaryFile` in `dataframe_to_civis`.  If you would like to resolve the merge conflict on this PR, we can merge your change in the meantime.  Otherwise, we can wait until after @elliothevel finishes his changes on `dataframe_to_civis` to solve the issue raised here.

I opened https://github.com/civisanalytics/civis-python/issues/10 for the broader issue of Windows compatibility.

keithing:
Closing this as the problematic `NamedTemporaryFile`s were removed in #7 and #11.

PR #11, opened 2016-12-08
Do not write to disk in dataframe_to_civis

elliothevel:
Use an in-memory stream instead of writing to a temporary file.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1,5 +1,5 @@
 import csv
-import tempfile
+import io
 
 from civis import APIClient
 from civis._utils import maybe_get_random_name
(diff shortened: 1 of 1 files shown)


civis/io/_tables.py
```diff
@@ -422,8 +432,12 @@ def csv_to_civis(filename, database, table, api_key=None,
...
 
     import_job = client.imports.post_files(**kwargs)
-    with open(filename, "rb") as data:
-        put_response = requests.put(import_job.upload_uri, data)
+    if hasattr(filename_or_buf, 'read'):
+        put_response = requests.put(import_job.upload_uri, filename_or_buf)
+    else:
+        with open(filename_or_buf, "rb") as data:
+            put_response = requests.put(import_job.upload_uri, data)
+
```
keithing: Instead of making a breaking change (i.e. change parameter `filename` to `filename_or_buf`), what do think about creating another function like `_import_bytes` that both `dataframe_to_civis` and `csv_to_civis` call?  We could avoid the somewhat awkward `if` / `else` here.

elliothevel: Yeah, that works too. I just pushed an update.

civis/io/_tables.py
```diff
@@ -1,5 +1,5 @@
 import csv
-import tempfile
+import io
```
keithing: `io` is imported below (it's in the weird `try` `except` to lead the way towards python2 compatibility).

civis/io/_tables.py
```diff
@@ -328,17 +328,17 @@ def dataframe_to_civis(df, database, table, api_key=None,
...
-                            api_key=api_key, max_errors=max_errors,
-                            distkey=distkey, sortkey1=sortkey1,
-                            sortkey2=sortkey2, delimiter=',',
-                            headers=headers, credential_id=credential_id,
-                            polling_interval=_DEFAULT_POLLING_INTERVAL,
-                            existing_table_rows=existing_table_rows,
-                            archive=archive)
-    return poll
+    buf = io.BytesIO()
+    # NOTE: TextIOWrapper is not thread-safe
```
keithing: Hmm.  This could be problematic since this function returns a `PollableResult` and is therefore possibly async.  Could we get around this with something like

```python
csv_str = df.to_csv(encoding='utf-8', index=False, **kwargs) # to_csv returns a string if no file is specified
csv_bytes = str.encode(csv_str)
return _import_bytes(csv_bytes, ...)
```

civis/io/_tables.py
elliothevel: Actually, what is the purpose of the try/except around that import? `from io import StringIO` works in 2.7 and 3.5.

http://python-future.org/compatible_idioms.html#stringio

keithing: hoho, nice.

civis/io/_tables.py
keithing: Summary of offline convo:

thread-safety shouldn't create any problems here and writing to the buffer could prevent an unnecessary copy.

keithing:
Ok, great. LGTM.  Could use a rebase before merge.

PR #12, opened 2016-12-08
Return a valid Response object on None responses

skomanduri:
### Issue

If a 204/205 response is returned from the API, and the client is set to the default return type of `'snake'`, an exception would be thrown.

### Solution

This PR returns a valid civis.response.Response object from these responses.  The `json_data` attribute is set to `None` to disambiguate empty JSON responses from no content responses.

An alternative is to propagate the `None` value back to the user, instead of returning a `Response` object.  This is significantly easier, but loses the response headers.  I'm not opposed to switching to that approach if we think it would be better.

### DIFF
--- a/civis/response.py
+++ b/civis/response.py
@@ -22,19 +22,23 @@ def _response_to_json(response):
 
     Returns
     -------
-    dict
-        The data in the response body.
+    dict | None
+        The data in the response body or None if the response has no
+        content.
 
     Raises
     ------
     CivisClientError
         If the data in the raw response cannot be parsed.
     """
-    try:
-        return response.json()
-    except ValueError:
-        raise CivisClientError("Unable to parse JSON from response",
-                               response)
+    if response.status_code in [204, 205]:
+        return None
+    else:
+        try:
+            return response.json()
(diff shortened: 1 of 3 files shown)


civis/response.py
```diff
@@ -22,19 +22,23 @@ def _response_to_json(response):
...
     CivisClientError
         If the data in the raw response cannot be parsed.
     """
-    try:
-        return response.json()
-    except ValueError:
-        raise CivisClientError("Unable to parse JSON from response",
-                               response)
+    if response.status_code in [204, 205]:
+        return None
```
keithing: What if this returns an empty dict instead of `None`?

skomanduri: Returning an empty dict here would mean that a list endpoint with no results is not distinguishable from an endpoint that returns no content.  I think we should keep these distinguishable, and wrap the original response as thinly as possible, but I'm open to other opinions.

keithing: Yeah, that's a convincing reason to return `None` instead of `{}`.

keithing:
LGTM

PR #16, opened 2016-12-09
ENH Switch "archive" to "hidden"

stephen-hoover:
For the `civis.io` functions, change the `archive` parameter (this ran a callback which archived the script after completion) to `hidden`. Instead of archiving scripts when they finish, hide them on creation. This is what the `archive` parameter had been trying to accomplish. Setting `hidden` keeps the scripts from cluttering the UI while leaving them accessible to e.g. status queries from the API. We also save the latency of an extra API call on script completion.

This PR also adds the `hidden` option to the `query_civis` function, which didn't have it before.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -6,6 +6,7 @@ from civis._utils import maybe_get_random_name
 from civis.base import EmptyResultError
 from civis.polling import PollableResult, _DEFAULT_POLLING_INTERVAL
 import requests
+import warnings
 
 try:
     from io import StringIO
@@ -28,7 +29,7 @@ DELIMITERS = {
 def read_civis(table, database, columns=None, use_pandas=False,
                job_name=None, api_key=None, credential_id=None,
                polling_interval=_DEFAULT_POLLING_INTERVAL,
-               archive=True, **kwargs):
+               archive=False, hidden=True, **kwargs):
     """Read data from a Civis table.
 
     Parameters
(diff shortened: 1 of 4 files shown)


stephen-hoover:
@keithing , this is ready for code review!

keithing:
Remove the parameter `archive` will be a breaking change.  Instead of removing it, we should add a deprecation warning that the parameter will be ignored in favor of `hidden` and will be removed after version `1.2.0`.

stephen-hoover:
@keithing , good point. Done. That was a big enough change that I rewrote the commit.

civis/io/_tables.py
```diff
@@ -55,9 +56,10 @@ def read_civis(table, database, columns=None, use_pandas=False,
         will be used.
     polling_interval : int or float, optional
         Number of seconds to wait between checks for query completion.
-    archive : bool, optional
-        If ``True`` (the default), archive the export job as soon as it
-        completes.
+    archive : bool, optional (deprecated)
+        If ``True``, archive the import job as soon as it completes.
```
keithing: Would it be more clear here to say that archive is just ignored in favor of hidden?

keithing: Oh, I see you're not actually removing the archiving functionality. That makes the most sense.

keithing:
One last thing, have you tested to see what happens if both `archive` and `hidden` are true?

stephen-hoover:
I just tested that. It works and does what you'd expect -- the object is created hidden and then archived immediately after completion. The two flags are actually orthogonal. "hidden" means that it doesn't show up in the UI (but it's still there), while "archived" objects are (mostly) inaccessible.

With client-controlled requests like this, you (at least I) often want things hidden so that you don't get dozens of machine-generated queries crowding out your activity list. But archiving these objects immediately after they run is not a best practice. (Hence this PR.)

civis/io/_tables.py
```diff
@@ -97,19 +99,22 @@ def read_civis(table, database, columns=None, use_pandas=False,
     """
     if use_pandas and NO_PANDAS:
         raise ImportError("use_pandas is True but pandas is not installed.")
+    if archive:
+        warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
+                      "Use `hidden` instead.", FutureWarning)
```
keithing: Since this calls `read_civis_sql`, a user will get two deprecation warnings if they set `archive=True` in `read_civis`.  If it's removed here, the drawback is that the warning may be a bit more confusing since it will be raised in `read_civis_sql`.  What do you think is better?

stephen-hoover: My preference would be to leave it in both places. The first warning will be clear. The second may be a bit confusing, but that seems like less confusion than not having a warning from the calling function.

PR #21, opened 2016-12-20
Update documentation: docs/index

EllieSA200:
Without the `export ` it doesn't work.  John Means and @dianecarroll encountered this difficulty when trying to set up from docs.


skomanduri:
Thanks for the update!

LGTM

PR #22, opened 2017-01-05
Pubnub notifications

menglewis:
This allows `PollableResult` to use Pubnub to subscribe to messages to get realtime notifications which is particularly useful for determining when a Job finishes. This functionality requires the `pubnub` library to be installed and the `pubnub` feature flag in Platform. If the library is not installed or the user doesn't have the feature flag, then it will gracefully degrade to using polling. It will also use polling in addition to Pubnub so if the connection to Pubnub is lost, then it will still poll for job status.

I also added a memoized feature_flags attribute to the client so it wouldn't do an API request to check for the feature flag every time it runs a job.


menglewis:
@elliothevel Could you give this PR a 30% review? I'm still testing some edge cases and need to write unit tests for this functionality and I might change a few things about how this code is organized.

civis/polling.py
```diff
@@ -104,6 +117,64 @@ def __repr__(self):
...
+                    self.job_id = job_id
+                    self.callback_function = callback_function
+
+                def message(self, pubnub, message):
+                    try:
+                        result = message.message
+                        if result['object']['id'] == self.job_id \
+                                and result['run']['state'] in DONE:
+                            self.callback_function()
+                    except:
```
elliothevel: This should catch a specific exception.

civis/polling.py
```diff
@@ -104,6 +117,64 @@ def __repr__(self):
...
+        if self.poller_args:
+            class JobCompleteListener(SubscribeCallback):
+                def __init__(self, job_id, callback_function):
+                    self.job_id = job_id
+                    self.callback_function = callback_function
+
+                def message(self, pubnub, message):
+                    try:
+                        result = message.message
+                        if result['object']['id'] == self.job_id \
```
elliothevel: Does the pubnub message include `run_id`? If so, it would be nice to check against both job and run ids to be as precise as possible.

civis/polling.py
```diff
@@ -104,6 +117,64 @@ def __repr__(self):
...
+                    pass
+
+            channel = self._client.channels.list()
+            channels = [chan['name'] for chan in channel['channels']]
+            pnconfig = PNConfiguration()
+            pnconfig.subscribe_key = channel['subscribe_key']
+            pnconfig.cipher_key = channel['cipher_key']
+            pnconfig.auth_key = channel['auth_key']
+            pnconfig.ssl = True
+            pnconfig.reconnect_policy = True
```
elliothevel: Most of this config could probably go in a separate function.

civis/polling.py
```diff
@@ -104,6 +117,64 @@ def __repr__(self):
                                             self._civis_state)
             return out
 
+    def _subscribe(self):
+        pubnub = None
+        if self.poller_args:
+            class JobCompleteListener(SubscribeCallback):
```
menglewis: Makes sense. I'll put it into a separate module.

civis/polling.py
menglewis: Yeah. The messages have `run_id`. The blocker right now is from Platform because `POST /jobs/:id/runs` currently returns a 204 with an empty response. I have a PR open to return the run that is created, so once that is merged (probably early next week) I'll update this to use run and also change the polling to use the runs endpoint instead of jobs.

civis/civis.py
```diff
@@ -309,3 +311,21 @@ def __init__(self, api_key=None, return_type='snake',
...
+    @property
+    def feature_flags(self):
+        if self._feature_flags:
+            return self._feature_flags
+        me = self.users.list_me()
+        self._feature_flags = tuple(flag for flag, value
+                                    in me['feature_flags'].items() if value)
+        return self._feature_flags
+
+    def run_job(self, job_id):
```
menglewis: I was kind of intending to use this as a convenience function like the `io` functions, but I wasn't really sure where to put it since it doesn't really fall under the umbrella of `io`.

civis/polling.py
menglewis: After thinking about it some more, I think that it might make more sense to leave `PollableResult` alone and create a separate class that implements the same interface for the Pubnub integration. Then we could create a wrapper that would connect to Pubnub if they have the requirements and fallback to Polling if they don't or if the Pubnub connection fails.

civis/pubnub.py
```diff
@@ -0,0 +1,80 @@
...
+
+        def status(self, pubnub, status):
+            pass
+
+        def presence(self, pubnub, presence):
+            pass
+
+
+class SubscribableResult(CivisAsyncResultBase):
+    def __init__(self, poller, poller_args, client):
```
stephen-hoover: Other places in the code that need a client object have the argument `api_key=None` and create a new client object from that. It's probably better to follow the same convention here. I think that at least the `client` input should be optional. It's cheap to create a new one.

menglewis: Cool, I didn't realize that the client initialization memoizes the swagger spec so I thought it was hitting the API every time it initializes an instance of the client. With that in mind, it definitely makes sense to match the API from other parts of the codebase and use `api_key=None`.

menglewis:
@elliothevel This is ready for a 90% review. Here are a list of things I changed since your last review:

* Separated out the Pubnub integration into a different module. I ended up extracting common functionality between the `PollableResult` and `SubscribableResult` into a base class. 
* Moved the `run_job` utility function to the `civis.utils` namespace. I see this function as a helper that provides a common set of operations to the user in the same way that `civio.io` works, but it didn't really fit within the `io` concept so I moved it to a separate module. I'm not really a fan of a vague name like `civis.utils` so if you have any better ideas for that namespace let me know.
* Added the pubnub Python package as an `extras_require`in `setup.py`. I checked a few other Python open source projects and they all defined `extras_require` in `setup.py` instead of creating a separate requirements file so I followed that pattern. Once it is released people will be able to install the client with pubnub requirements with `pip install civis[pubnub]`.

stephen-hoover:
@menglewis , is it worth matching the input arguments between the two result types? That would mean having a dummy `polling_interval` argument for the `SubscribableResult` and a dummy `api_key` argument for the `PollableResult`. That would make it easy to swap them out for each other; often code will specify a polling interval, and that would cause an error when swapping out the class type.

civis/pubnub.py
```diff
@@ -0,0 +1,100 @@
...
+                    and message['run']['id'] == self.poller_args[1] \
+                    and message['run']['state'] in DONE
+        except KeyError:
+            return False
+        return match
+
+    def _set_api_result(self, result=None):
+        with self._condition:
+            if result is None:
+                result = self.poller(*self.poller_args)
```
elliothevel: `PollableResult` wraps the `self.poller` call in a `try/except` since it could raise API errors. I think we should also catch exceptions here.

elliothevel: `PollableResult` wraps the `self.poller` call in a `try/except` since it could raise API errors. We should also catch exceptions here and call `set_exception`.

civis/pubnub.py
```diff
@@ -0,0 +1,100 @@
...
+        pnconfig.subscribe_key = channel_config['subscribe_key']
+        pnconfig.cipher_key = channel_config['cipher_key']
+        pnconfig.auth_key = channel_config['auth_key']
+        pnconfig.ssl = True
+        pnconfig.reconnect_policy = True
+        return pnconfig, channels
+
+    def _check_message(self, message):
+        try:
+            match = message['object']['id'] == self.poller_args[0] \
```
elliothevel: I think wrapping these lines in parentheses is more pythonic than the backslash continuations.

civis/base.py
```diff
@@ -84,3 +85,81 @@ def _call_api(self, method, path=None, params=None, data=None, **kwargs):
...
+STATE_TRANS = {}
+for name in FINISHED + FAILED:
+    STATE_TRANS[name] = futures._base.FINISHED
+for name in NOT_FINISHED:
+    STATE_TRANS[name] = futures._base.RUNNING
+for name in CANCELLED:
+    STATE_TRANS[name] = futures._base.CANCELLED_AND_NOTIFIED
+
+
+class CivisAsyncResultBase(futures.Future):
```
elliothevel: Would it be clearer to move some of the comments directly below the docstring for the `PollableResult` class here? Also worth mentioning that subclasses should override `_check_result` or `set_result`/`set_exception`. Are those the only methods that need to be implemented?

civis/utils/_jobs.py
```diff
@@ -0,0 +1,28 @@
...
+    job_id : str or int
+        The ID of the job.
+    api_key : str, optional
+        Your Civis API key. If not given, the :envvar:`CIVIS_API_KEY`
+        environment variable will be used.
+
+    Returns
+    -------
+    results : :class:`~civis.polling.PollableResult`
+        A `PollableResult` object.
```
elliothevel: or a `SubscribableResult` object if pubnub is available?

setup.py
```diff
@@ -35,6 +35,9 @@ def main():
         packages=find_packages(),
         long_description=read('README.md'),
         install_requires=required,
+        extras_require={
```
elliothevel: 👍 (Not necessary for this PR, but we should eventually list `pandas` here as well.)

civis/base.py
menglewis: Yeah, that makes sense. `_check_result` doesn't have to be overwritten by subclasses, by default it will just check the `_result` attribute. It is overwritten by `PollableResult` because it checks by polling where the check isn't overwritten on `SubscribableResult` so it just checks if something has been set on `._result`. 

`set_result` or `set_exception` should be called and if there are any intermediate states that need to be specified than either setting `_result` or overwriting `_check_result` can be used to accomplish that. I'll draft up a docstring for the Base class describing how it should be used by subclasses.

menglewis:
@elliothevel Thanks for the review. I made the changes you suggested and it's ready for another pass. I also made some updates based on feedback from @stephen-hoover so `PollableResult` and `SubscribableResult` both have the same signature on `__init__` (and any other future subclasses of `CivisAsyncResultBase`)

elliothevel:
LGTM!

PR #24, opened 2017-01-11
Use Endpoint._base_url when fetching swagger

jcobian:
The url for fetching the swagger was harcoded in `civis/resources/_resources.py`. Using `Endpoint._base_url` in case the url ever changes we only need to update it in one place.


skomanduri:
LGTM

PR #26, opened 2017-01-31
[CIVP-10187] Dynamically decide between polling and notifications

stephen-hoover:
Now that the `pubnub.SubscribableResult` is available, create a function `result.make_platform_future` which instantiates either a new `SubscribableResult` (if `pubnub` is available and the user has access to the notifications endpoint) or a `PollableResult` object. We should prefer to use the `SubscribableResult`, since notifications are much faster (we don't have to wait for polling) and greatly decrease API traffic.

### DIFF
--- a/civis/base.py
+++ b/civis/base.py
@@ -10,6 +10,7 @@ FAILED = ['failed']
 NOT_FINISHED = ['queued', 'running']
 CANCELLED = ['cancelled']
 DONE = FINISHED + FAILED + CANCELLED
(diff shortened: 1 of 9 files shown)


menglewis:
You can ignore those Travis errors. It looks like an issue caused by `generate_classes` lru cache persisting across tests when some Pubnub tests require a swagger spec with channels. I'll create a separate PR that fixes that.

stephen-hoover:
#35 is a much better approach to this.

PR #27, opened 2017-02-01
Fixed reconnection policy for pubnub connections

menglewis:
I think when I initially built this, I was looking at documentation on an older Pubnub SDK when adding in the reconnection policy. This PR should fix the reconnection policy based on SDK v4.


skomanduri:
LGTM

PR #32, opened 2017-02-07
[CIVP-10066] Catch unexpected keyword arguments passed to methods

keithing:
Currently, API resource methods take optional parameters in the form of `**kwargs` in the method signature.  This is convenient for parsing the API specification, but can produce unexpected behavior when an unexpected keyword argument is passed into the method.  Instead of alerting the user that the argument is unexpected, it is simply ignored (causing some confusion).  This change adds an explicit check on all method calls for unexpected arguments so that they may be caught earlier.

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -256,8 +256,8 @@ def raise_for_unexpected_kwargs(method_name, arguments, sig_args, sig_kwargs,
         expected |= {'iterator'}
     unexpected = set(arguments.keys()) - expected
     if unexpected:
-        msg_fmt = "{}() got an unexpected keyword argument '{}'"
-        raise TypeError(msg_fmt.format(method_name, next(iter(unexpected))))
+        msg_fmt = "{}() got an unexpected keyword argument(s) {}"
+        raise TypeError(msg_fmt.format(method_name, str(unexpected)))
 
 
 def bracketed(x):
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -223,27 +223,41 @@ def create_method(params, verb, method_name, path, doc):
         A function which will make an API call
     """
     elements = split_method_params(params)
-    args, kwargs, body_params, query_params, path_params = elements
-    sig = create_signature(args, kwargs)
+    sig_args, sig_kwargs, body_params, query_params, path_params = elements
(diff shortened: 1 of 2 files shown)


civis/resources/_resources.py
```diff
@@ -223,29 +223,43 @@ def create_method(params, verb, method_name, path, doc):
...
+def raise_for_unexpected_kwargs(method_name, arguments, sig_args, sig_kwargs,
+                                is_iterable):
+    """Raise TypeError if arguments are not in sig_args or sig_kwargs."""
+    expected = set(sig_args) | set(sig_kwargs)
+    if is_iterable:
+        expected |= {'iterator'}
+    unexpected = set(arguments.keys()) - expected
+    if unexpected:
+        msg_fmt = "{}() got an unexpected keyword argument '{}'"
+        raise TypeError(msg_fmt.format(method_name, next(iter(unexpected))))
```
stephen-hoover: Should this allow for multiple unexpected arguments? You could change "argument" to "argument(s)" in `msg_fmt` and use
```
msg_fmt.format(method_name, str(unexpected))
```

civis/tests/test_resources.py
```diff
@@ -54,10 +54,8 @@ def test_create_method_iterator_kwarg():
 
 
 def test_create_method_no_iterator_kwarg():
-    # We don't do any validation on keyword arguments if **kwargs is in the
-    # signature. They are just ignored if they aren't in the expected body,
-    # query, or path params. Therefore, method should return a Type Error,
-    # but method2 contains **kwargs so should overwrite iterator = False
+
+    # Test for TypeError when iterator is present and no **kwargs
```
stephen-hoover: I find the statement about kwargs difficult to understand. Perhaps this could say,

> Test that dynamically-created function errors when an unexpected "iterator" parameter is passed in

civis/tests/test_resources.py
```diff
@@ -68,13 +66,16 @@ def test_create_method_no_iterator_kwarg():
 
     assert 'keyword argument' in str(excinfo.value)
 
+    # Test for TypeError when iterator is present and method takes **kwargs,
+    # but iterator is not valid
```
stephen-hoover: Similarly, maybe this could be
 
> Dynamic functions handle optional argument through a different code path; verify that this also rejects unexpected arguments.

PR #34, opened 2017-02-08
Removing intersphinx references

keithing:
References to 3rd party documentation is occasionally causing build errors.  The upside from having the linked documentation is probably not worth the downside of having nondeterministic errors, so I'm removing the links in this PR to `pandas` and `requests` libraries.

### DIFF
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -264,12 +264,12 @@ class APIClient(MetaMixin):
     return_type : str, optional
         The following types are implemented:
 
-        - ``'raw'`` Returns the raw :class:`requests:requests.Response` object.
+        - ``'raw'`` Returns the raw ``requests.Response`` object.
         - ``'snake'`` Returns a :class:`civis.response.Response` object for the
           json-encoded content of a response. This maps the top-level json
           keys to snake_case.
-        - ``'pandas'`` Returns a :class:`pandas:pandas.DataFrame` for
-          list-like responses and a :class:`pandas:pandas.Series` for single a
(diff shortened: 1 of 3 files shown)


skomanduri:
@keithing This seems like a slick feature to me, though.  Did you consider downloading the inventories to use as a backup?
```
intersphinx_mapping = {'python': ('https://docs.python.org/3.4',
                                  (None, 'python-inv.txt'))}
```

elliothevel:
The errors are a result of using the `-W` flag, which turns all warnings into errors, in the `sphinx-build` command. The docs can build successfully even if the inventories are missing, the links just won't be added.

I think downloading the inventories and adding them as backups only if doing a test build is a good solution.

Another option would be to monkey-patch `sphinx.ext.intersphinx.fetch_inventory` to ignore warnings raised when trying to download. Something like this:
```python
from sphinx.ext import intersphinx

def ignore_warnings(func):
    def f(*args, **kwargs):
        app = args[0]
        error_setting = app.warningiserror
        app.warningiserror = False
        result = func(*args, **kwargs)
        app.warningiserror = error_setting
        return result
    return f
intersphinx.fetch_inventory = ignore_warnings(intersphinx.fetch_inventory)
```

skomanduri:
> The errors are a result of using the -W flag, which turns all warnings into errors, in the sphinx-build command. The docs can build successfully even if the inventories are missing, the links just won't be added.

Oh!  In that case I would advocate for just dropping that -W flag.  That way we still get the links in ReadTheDocs.

keithing:
closing in favor of https://github.com/civisanalytics/civis-python/pull/37

PR #35, opened 2017-02-10
[CIVP-10187] Utilize Polling or Pubnub more transparently to users

menglewis:
* Updated `SubscribableResult.__init__` to check for the existence of `channels` and the presence of the Pubnub sdk before connecting to Pubnub
* Created `civis.futures.CivisFuture` which combines the functionality of `PollableResult` and `SubscribableResult`
* Updated `civis.utils.run_job` to use `CivisFuture`

TODO:
* Write tests for new functionality
* Add docstring to `CivisFuture` and update documentation
* Update other places in the client that currently use `PollableResult` to `CivisFuture`

Notes on CivisFuture
* It uses multiple inheritance to combine the functionality of `PollableResult` and `SubscribableResult`.
* On init, it will check if it meets the conditions to connect to Pubnub and if so, it will use the `_LONG_POLLING_INTERVAL` which is set to 10 minutes for polling instead of the passed in or default polling interval


menglewis:
@stephen-hoover Could you give a 30% review on the architectural decisions here? I decided to use multiple inheritance because it was less code and it keeps the different pieces of functionality more modular. If you think those decisions are okay, I'll continue with this approach and update the tests and documentation.

stephen-hoover:
I think that multiple inheritance is the right thing to do here, but you'll have to be careful where methods overlap. The `__init__` as written, for example, only calls the `PollableResult.__init__`, but I think you meant it to call both. I have three suggestions for how to restructure:

First possibility: What if the two existing classes were separated into a mixin and the instantiable class, and the `CivisFuture` was defined as `class CivisFuture(CivisAsyncResultBase, PollableMixin, SubscribableMixin)`? Then there'd be, for example
```python
class PollableResult(CivisAsyncResultBase, PollableMixin):
    def __init__(...):
        super().__init__(...)
        # Other PollableResult-specific init
```

Second possibility: Keep the `CivisFuture` as-is (maybe add `CivisAsyncResultBase` as the first superclass) and write a combined `__init__` instead of calling the specialized subclass inits. You'd have to do that anyway for the first option. I prefer the mixins to this, since I think it makes it clearer what's going on.

Third possibility: Get rid of `SubscribableResult` and replace it with `CivisFuture(PollableResult)`, which has all of the old `SubscribableResult`'s methods copied in. Downside is that this alters the library's public API. However, I think that you really need the occasional polling to make the `SubscribableResult` work properly anyway, so I'm not sure we want people using the v1.2 `SubscribableResult`.

I lean toward the third one, but the first could also work. What do you think?

menglewis:
On the first point regarding the current implementation, with how the MRO is setup on the class, the `super()` call in `PollableResult` will end up calling `__init__` on `SubscribableResult`. Then `super()` on `SubscribableResult` will call `CivisAsyncResultBase`, then `futures.Future`. 

That being said I think the third option that you listed is more clear and maintainable. I agree based on earlier conversation that `SubscribableResult` needs occasionally polling to handle some of the edge cases such as the run being finished before it connects to Pubnub or if it disconnects for more than 10 minutes. I can go ahead with building `CivisFuture` in that way. 

Since `SubscribableResult` is part of the public api in v1.2, we can add some handling to properly deprecate it. We can modify `SubscribableResult` to be a function that shows a deprecation warning then instantiates an instance of `CivisFuture` with the arguments passed to it.

stephen-hoover:
That sounds like a good solution!

So that I don't forget it later -- I think it would be helpful to have an option to poll for run status on object creation. It should default to True, so that it catches instances where the run has completed before the object subscribed to the notification channel, but that would allow us to skip that API call if we know for certain that the run hasn't finished. That could matter if a user is creating a large number of these at once.

menglewis:
@stephen-hoover Back to you. I implemented `CivisFuture` so it inherits from `PollableResult` and everything from `SubscribableResult` is implemented on `CivisFuture`. I also modified `SubscribableResult` so that it sends a deprecation warning and instantiates an instance of `CivisFuture`. I also added a parameter called `poll_on_creation`. When True it will function the same way that `PollableResult` always has. When False, it will set `_last_polled` to the current time at instantiation which will cause the poller to only poll once the time specified in `polling_interval` has passed.

civis/tests/test_futures.py
```diff
@@ -155,9 +164,32 @@ def test_set_api_result_failed(self, mock_subscribe, mock_api):
...
+    def test_overwrite_polling_interval_with_channels(self, *mocks):
+        future = CivisFuture(lambda x: x, (1, 20), polling_interval=5)
+        assert future.polling_interval == _LONG_POLLING_INTERVAL
+
+    @pytest.mark.skipif(not has_pubnub, reason="pubnub not installed")
+    @patch(swagger_import_str, return_value=civis_api_spec_base)
+    @patch.object(CivisFuture, '_subscribe')
+    def test_polling_interval(self, *mocks):
+        # This test uses a different swagger spec than the other tests so it
+        # should clear the cached values before and after
```
stephen-hoover: This test is the case when the channels endpoint isn't available and we fall back on polling, correct? Could you state that in this comment? Also, please add an `assert not hasattr(future, '_pubnub')`.

civis/tests/test_futures.py
```diff
@@ -155,9 +164,32 @@ def test_set_api_result_failed(self, mock_subscribe, mock_api):
...
         assert result._state == 'FINISHED'
         with pytest.raises(CivisJobFailure):
             result.result()
+
+    @pytest.mark.skipif(not has_pubnub, reason="pubnub not installed")
+    @patch(swagger_import_str, return_value=civis_api_spec_channels)
+    @patch.object(CivisFuture, '_subscribe')
+    def test_overwrite_polling_interval_with_channels(self, *mocks):
+        future = CivisFuture(lambda x: x, (1, 20), polling_interval=5)
+        assert future.polling_interval == _LONG_POLLING_INTERVAL
```
stephen-hoover: Maybe add `assert hasattr(future, '_pubnub')`? That seems to be the one visible change between using notifications and not. Maybe it's worth having a public boolean attribute `subscribed` which tells users if it's subscribed to the notifications channel.

civis/tests/test_polling.py
```diff
@@ -97,6 +98,17 @@ def _check_result(self):
...
 
+    def test_poll_on_creation(self):
+        poller = mock.Mock(side_effect=Response({"state": "running"}))
+        pollable = PollableResult(poller,
+                                  (),
+                                  polling_interval=0.1,
+                                  poll_on_creation=False)
+        repr(pollable)
+        assert poller.call_count == 0
+        time.sleep(0.2)
```
stephen-hoover: Will this test still work if the polling interval is 0.01 and the sleep is 0.02? Adding 0.2 seconds here and there in the tests will add up.

civis/utils/_jobs.py
```diff
@@ -16,15 +15,11 @@ def run_job(job_id, api_key=None):
...
     client = APIClient(api_key=api_key, resources='all')
     run = client.jobs.post_runs(job_id)
-    if 'pubnub' in client.feature_flags and has_pubnub:
-        return SubscribableResult(client.jobs.get_runs,
-                                  (job_id, run['id']),
-                                  api_key=api_key)
-    return PollableResult(client.jobs.get_runs, (job_id, run['id']))
+    return CivisFuture(client.jobs.get_runs,
+                       (job_id, run['id']),
+                       api_key=api_key)
```
stephen-hoover: This can use `poll_on_creation = False`, correct? Since it's creating a job immediately before initializing the object, you know it hasn't finished.

menglewis:
Good call on making sure that Pubnub connections are unsubscribed from when Polling sets the result. I refactored some of the logic for setting the result since `PollableResult` and `CivisFuture` had almost the exact same logic. I added the `_set_api_result` method to `PollableResult` which takes a result from the poller and determines the necessary action based on the result state. I also added a `cleanup` method which gets called in `_set_api_result` which doesn't do anything on `PollableResult`, but it allows `CivisFuture` to define the method to unsubscribe from Pubnub.

I also added a `subscribed` property to `CivisFuture` that will return True if the object has a `_pubnub` attribute and returns at least 1 item from Pubnub's `get_subscribed_channels` method.

I believe that I addressed everything in your review so it is ready for another round of review.

civis/futures.py
```diff
@@ -0,0 +1,148 @@
...
+
+try:
+    from pubnub.pubnub import PubNub
+    from pubnub.pnconfiguration import PNConfiguration, PNReconnectionPolicy
+    from pubnub.callbacks import SubscribeCallback
+    has_pubnub = True
+except ImportError:
+    has_pubnub = False
+
+_LONG_POLLING_INTERVAL = 9.5 * 60
```
stephen-hoover: Could you add a comment about where this number comes from? It's that `pubnub` can recover from network outages of up to 10 minutes, but we need to ensure that we catch completions which happen in longer outages, correct?

civis/polling.py
```diff
@@ -123,16 +137,25 @@ def _check_result(self):
...
+
+    def _set_api_result(self, result):
+        with self._condition:
+            if result.state in FAILED:
+                try:
+                    err_msg = str(result['error'])
+                except:
+                    err_msg = str(result)
+                self.set_exception(CivisJobFailure(err_msg,
+                                                   result))
```
stephen-hoover: This fits on one line now.

PR #36, opened 2017-02-21
ENH Use requests-toolbelt for streaming uploads of Files

stephen-hoover:
If users have `requests-toolbelt` installed, then we can stream files into AWS without needing to read them into memory all at once. This allows users to upload files larger than they could have otherwise (e.g. tens of GB).

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -6,9 +6,9 @@ from civis import APIClient
 from civis.base import EmptyResultError
 try:
     from requests_toolbelt.multipart.encoder import MultipartEncoder
-    has_toolbelt = True
+    HAS_TOOLBELT = True
 except ImportError:
-    has_toolbelt = False
+    HAS_TOOLBELT = False
 
 
 def file_to_civis(buf, name, api_key=None, **kwargs):
@@ -66,7 +66,7 @@ def file_to_civis(buf, name, api_key=None, **kwargs):
     form_key['file'] = buf
 
     url = file_response.upload_url
-    if has_toolbelt:
+    if HAS_TOOLBELT:
         # This streams from the open file buffer without holding the
         # contents in memory.
         en = MultipartEncoder(fields=form_key)
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -69,9 +69,14 @@ def file_to_civis(buf, name, api_key=None, **kwargs):
     if has_toolbelt:
         # This streams from the open file buffer without holding the
         # contents in memory.
-        enc = MultipartEncoder(fields=form_key)
-        response = requests.post(url, data=enc,
-                                 headers={'Content-Type': enc.content_type})
+        en = MultipartEncoder(fields=form_key)
+        if en.len / 2 ** 20 < 100:
+            # Semi-arbitrary cutoff for "small" files.
+            # Send these with requests directly because that uses less CPU
+            response = requests.post(url, files=form_key)
+        else:
+            response = requests.post(url, data=en,
+                                     headers={'Content-Type': en.content_type})
     else:
         response = requests.post(url, files=form_key)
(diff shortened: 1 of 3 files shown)


stephen-hoover:
@keithing , if you know someone will more experience in these file uploads, feel free to shift the review request to them.

@seancquinn for viz.

keithing:
This is tricky.  From what I can tell, `MultiPartEncoder` seems to be substantially slower than using the default behavior of `requests`--based on the [docs](https://toolbelt.readthedocs.io/en/latest/uploading-data.html?highlight=multipartencoder) and [this issue](https://github.com/sigmavirus24/requests-toolbelt/issues/75).  I wonder if we want to separate this into a new function so that default behavior is to use the fast upload, but we still have the option to upload without reading into memory?

skomanduri:
Can you use http://docs.python-requests.org/en/master/user/advanced/#streaming-uploads directly from the requests library?

stephen-hoover:
Interesting. I tested with an 80 MB file just now, and I see similar overall performance between the two codepaths. I peak around 20 MB/second transfer speeds either way. The `MultiPartEncoder` does appear to use about 8x more CPU time (about 1s versus 130 ms), but the full upload is about 6-7 seconds either way.

I tried a 3.1 GB file as well. That fails completely with the old code path. Using the `MultiPartEncoder`, I had an average transfer speed of ~15 MB/s over a 3.5 minute upload.

Using a relatively new laptop and a good network connection, I seem to still be network limited. If someone wanted to run these functions in a thread, or if they had a significantly slower machine or significantly faster network connection, there could be issues.

I don't like the idea a separate function for this. They're doing the same thing, and it seems confusing for users to try to juggle the performance issues themselves, especially if it's not a big deal. I added a suggestion for switching between the two methods based on file size. What do you think of that?

stephen-hoover:
@skomanduri , I saw that part of the `requests` documentation, but I was unable to get it to work. The docs imply that the file pointer must be the entire body, but we have several other keys we need to include. If you can get it to work, then that would be a better solution -- even if the performance is the same, it would remove the need for an (optional) external dependency.

stephen-hoover:
Changing `requests.post(url, files=form_key)` to `requests.post(url, data=form_key)` causes `requests` to read the entire file into memory.

keithing:
> I tried a 3.1 GB file as well. That fails completely with the old code path. Using the MultiPartEncoder, I had an average transfer speed of ~15 MB/s over a 3.5 minute upload.

That's good enough for me!  I agree we should avoid multiple functions if the `MultiPartEncoder` is usually better or on par with the current method.

civis/io/_files.py
```diff
@@ -4,6 +4,11 @@
 
 from civis import APIClient
 from civis.base import EmptyResultError
+try:
+    from requests_toolbelt.multipart.encoder import MultipartEncoder
+    has_toolbelt = True
+except ImportError:
+    has_toolbelt = False
```
keithing: Capitalize the const to `HAS_TOOLBELT` here?

keithing:
From Request's [Quickstart](http://docs.python-requests.org/en/master/user/quickstart/?highlight=toolbelt)
> In the event you are posting a very large file as a multipart/form-data request, you may want to stream the request. By default, requests does not support this, but there is a separate package which does - requests-toolbelt. You should read the toolbelt's documentation for more details about how to use it.

From the quickstart it looks like we'll have to use `requests-toolbelt`

civis/io/_files.py
stephen-hoover: Done.

keithing:
LGTM.  @skomanduri any other comments?

skomanduri:
@keithing No other comments.  Thanks for digging into the `requests` docs.

👍

PR #38, opened 2017-02-22
BUG make sure click fails properly

### LINKED ISSUE #25: Buggy error in click parameter type
This line is wrong

https://github.com/civisanalytics/civis-python/blob/master/civis/cli/__main__.py#L54

It needs to read

```python
self.fail("Could not load YAML from path: %s" % value, param, ctx)
```

See this example from the click docs given in the comments in the code.

```python
import click

class BasedIntParamType(click.ParamType):
    name = 'integer'

    def convert(self, value, param, ctx):
        try:
            if value[:2].lower() == '0x':
                return int(value[2:], 16)
            elif value[:1] == '0':
                return int(value, 8)
            return int(value, 10)
        except ValueError:
            self.fail('%s is not a valid integer' % value, param, ctx)

BASED_INT = BasedIntParamType()
```

beckermr:
Closes #25

### DIFF
--- a/civis/cli/__main__.py
+++ b/civis/cli/__main__.py
@@ -51,7 +51,7 @@ class YAMLParamType(click.ParamType):
                 result = yaml.load(f)
                 return result
         except Exception:
-            self.fail("Could not load YAML from path: %s", value)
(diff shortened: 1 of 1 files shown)


keithing:
Thank! LGTM.

PR #39, opened 2017-02-22
[CIVP-10187] Replace PollableResult with CivisFuture

stephen-hoover:
Now that the `futures.CivisFuture` class exists, replace all uses of `PollableResult` with `CivisFuture`. This improves performance by increasing speed (we find out that runs complete immediately, instead of after polling) and by reducing the number of API calls needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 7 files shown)


stephen-hoover:
@keithing , Travis fails because there's a warning when building the docs. Do you know how to resolve that?

keithing:
That should be solved after https://github.com/civisanalytics/civis-python/pull/37

keithing:
LGTM

PR #40, opened 2017-03-02
BUG Make sure that the CivisFuture always cleans up after exceptions

stephen-hoover:
There were a couple of circumstances where exceptions raised in polling could have left us still subscribed to the `pubnub` notifications channel. Refactor the exception setting so that it only happens in one place; that way we can be sure that `cleanup` always gets called.

After this change, there's now only one place where we call `set_exception`, just as there's only one place where we call `set_result`.


stephen-hoover:
@keithing for viz

PR #43, opened 2017-03-10
[CIVP-10381] Allow base url to be configured from environment variable

elliothevel:
Allow the base url of the API to be configured through the `CIVIS_API_ENDPOINT` environment variable.


keithing:
LGTM

PR #44, opened 2017-03-13
ENH Allow use of existing client objects without re-creating

stephen-hoover:
High-level functions and classes in this library all take `api_key` inputs and instantiate new `APIClient` objects, rather than allowing users to pass in an existing `APIClient` object. There are cases when it can be very helpful to be able to pass in existing client objects. The top two:
- Sometimes we have an `APIClient` but not an `api_key`. I've written `api_key=self.client._session.auth[0]` a couple of times in my code now, and that's not good practice.
- For testing, it's very helpful to be able to create one `APIClient` mock and pass that around. If all functions instantiate new objects, the `APIClient` needs to be patched separately in each module.

This PR modifies the `APIClient` so that if one provides an existing `APIClient` to the `api_key` parameter, that object will be used unchanged, rather than instantiating a new object.

### DIFF
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -258,9 +258,11 @@ class APIClient(MetaMixin):
 
     Parameters
     ----------
-    api_key : str, optional
+    api_key : str or APIClient, optional
         Your API key obtained from the Civis Platform. If not given, the
         client will use the :envvar:`CIVIS_API_KEY` environment variable.
+        If the input is an existing :class:`civis.civis.APIClient`, it will
+        be returned and a new client object will not be created.
     return_type : str, optional
         The following types are implemented:
 
@@ -285,6 +287,9 @@ class APIClient(MetaMixin):
     """
     def __init__(self, api_key=None, return_type='snake',
                  retry_total=6, api_version="1.0", resources="base"):
+        if api_key is self:
+            # The object is already initialized.
+            return
         if return_type not in ['snake', 'raw', 'pandas']:
             raise ValueError("Return type must be one of 'snake', 'raw', "
                              "'pandas'")
@@ -310,6 +315,14 @@ class APIClient(MetaMixin):
                                    resources=resources)
         for class_name, cls in classes.items():
             setattr(self, class_name, cls(session, return_type))
+
+    def __new__(cls, api_key=None, *args, **kwargs):
+        if isinstance(api_key, cls):
+            # If the "api_key" is already an APIClient, use the
+            # existing client and don't create a new one.
+            return api_key
+        else:
+            return super().__new__(cls)
 
     @property
     def feature_flags(self):
--- a/civis/tests/test_client.py
+++ b/civis/tests/test_client.py
@@ -39,3 +39,13 @@ class ClientTests(CivisVCRTestCase):
             client.feature_flags
             client.feature_flags
             self.assertEqual(client.users.list_me.call_count, 1)
+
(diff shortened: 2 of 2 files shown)


skomanduri:
> High-level functions and classes in this library all take api_key inputs and instantiate new APIClient objects, rather than allowing users to pass in an existing APIClient object.

It seems like a better solution here would be to modify the existing functions/classes to take `APIClient`s directly rather than an API key, and enforce that convention on high-level functions going forward.  Are there a lot of them?

stephen-hoover:
There's O(10) places in this library. It would also be a significantly API-breaking change; all users of this library would need to update their own code. I'm indifferent whether we replace `api_key` parameters with `client` parameters, as long as there's some way to use existing client objects. 

The change I propose here is the simplest, and would have no disruption to users. On the other hand, replacing the `api_key` parameter would be cleaner, if we expect that users would seldom input the key directly. I don't know what the normal use looks like.

Perhaps @keithing has insights on the original design decision to take `api_key` as a parameter rather than `client`?

keithing:
>Sometimes we have an APIClient but not an api_key. I've written api_key=self.client._session.auth[0] a couple of times in my code now, and that's not good practice.

Do you mind clarifying this?  I don't see why passing `api_key` into the higher level functions is bad practice.  We memoize the instantiation to `APIClient` and thanks you @stephen-hoover, that is much faster now. 

>For testing, it's very helpful to be able to create one APIClient mock and pass that around. If all functions instantiate new objects, the APIClient needs to be patched separately in each module.

This strikes me as a nice to have and isn't a strong enough reason on its own to merit the change.

stephen-hoover:
I don't mean that taking an `api_key` input is bad practice. I meant that finding an API key inside a private attribute is bad practice. An existing `APIClient` object stores its key at `_session.auth[0]`, which is an implementation detail. Maybe I could create a new client every time I use it and only store the key, but if the `APIClient` isn't meant to be kept around, it should be a function and not an object.

Performance-wise, there's a small hit in that some places will use `resources='all'` and some will use `resources='base'`, leading to parsing the swagger twice when we only really needed to parse it once. That's not a huge deal, though.

Testing is actually my biggest motivation for wanting this change. I write a lot of tests of code which uses the client, and I've spent a lot of time tracking down which module I need to patch. I have cases where I need to patch the APIClient in three different modules (and make sure that each mock is returning the same object). Changes which make code easier to test save developer time and result in better-tested code.

keithing:
Forgot to make this point earlier:

> Perhaps @keithing has insights on the original design decision to take api_key as a parameter rather than client?

This was a bit of a future-proofing effort.  If we ever use a different `api_version` in `APIClient`, then the higher level functions may only work with a specific `api_version`.  If users can pass a specific `APIClient`, then we'd have to support multiple `api_versions` or raise an error with old versions.  That might be an unlikely scenario.

keithing:
> The change I propose here is the simplest, and would have no disruption to users. On the other hand, replacing the api_key parameter would be cleaner, if we expect that users would seldom input the key directly. I don't know what the normal use looks like.

Ok, I'm convinced the latter is the right way to go.  We can add the parameter `client` and deprecate `api_key` so no code breaks.  We'll have to be careful to when users put in a `client` object that would cause the high level function to break, but ultimately this change would make the high level functions more flexible.

stephen-hoover:
Okay, closing this.

PR #45, opened 2017-03-14
BUG Ensure that polling thread shuts down when no longer needed

stephen-hoover:
When a Python session shuts down, all threads are joined. That means that we need to wait for the polling thread to complete before the interpreter can shut down. It hadn't been noticable when polling was the primary way of getting results, but with the use of the notifications endpoint, it matters. I observed a very short job which used notifications wait 9.5 minutes to complete, because it had to wait for the `sleep` inside our polling thread to finish.

There's no way to shut down a thread in a ``ThreadPoolExecutor`` early, so replace it with our own thread. The ``cancel`` method lets us stop it early, and overloading ``join`` ensures that we never need to wait for a polling cycle to complete when exiting the interpreter. This also solves a problem I'd observed in testing, that the test session wouldn't exit until the poller shut down.

This PR also stops wrapping `_check_result` in a try/except inside the thread. It's only necessary if we think there could be coding bugs inside `_check_result`. If there are, we should fix them rather than swallowing them.

This PR requires a modification of `test_io` so that it takes a reasonable amount of time. When we record tests with ``vcr``, we need to have normal polling intervals, so as not to break the Platform API. But when running tests later, we don't want to wait through the entire sequence of polling at the normal intervals. We had been solving this by patching some internals of `PollableResult`. That solution breaks when we swap in the new threading poller. Replace with conditional poll intervals. We need to remember to set `polling_interval` for all new tests, but in return we don't need to patch private methods of `PollableResult` (and it works with the new threading).


stephen-hoover:
@keithing , tests fixed. The failure was on the new test I added -- it counts numbers of times the polling function gets called, and I think switching threads is not exact at the millisecond level. I extended the polling interval in that test. I don't want it to be too long and slow the tests down. If you think that new test isn't helpful, we could get rid of it.

civis/polling.py
```diff
@@ -8,6 +8,37 @@
 _DEFAULT_POLLING_INTERVAL = 15
 
 
+class ResultPollingThread(threading.Thread):
```
stephen-hoover: I left this public, but maybe it ended up specific enough that it should be private?

keithing: Yeah, might as well make this private.

civis/tests/test_polling.py
```diff
@@ -64,51 +57,33 @@ def test_error_setting(self):
...
                                   poll_on_creation=False)
-        repr(pollable)
+        pollable.done()  # Check status once to start the polling thread
         assert poller.call_count == 0
-        time.sleep(0.02)
+        time.sleep(0.015)
         assert poller.call_count == 1
 
 
+def test_repeated_polling():
```
keithing: Could this vary on different machines?

stephen-hoover: It shouldn't if the interval is long enough that we reliably check the thread every polling period. These are in units of seconds.

civis/polling.py
stephen-hoover: Done.

keithing:
LGTM

stephen-hoover:
@keithing , I add to resolve a couple of minor merge conflicts by hand.

keithing:
LGTM

PR #46, opened 2017-03-14
[CIVP-10592] Deprecate the api_key in favor of APIClient inputs

stephen-hoover:
High-level functions and classes in this library all take `api_key` inputs and instantiate new `APIClient` objects, rather than allowing users to pass in an existing `APIClient` object. There are cases when it can be very helpful to be able to pass in existing client objects. The top two:
- Sometimes we have an `APIClient` but not an `api_key`. I've written `api_key=self.client._session.auth[0]` a couple of times in my code now, and that's not good practice.
- For testing, it's very helpful to be able to create one `APIClient` mock and pass that around. If all functions instantiate new objects, the `APIClient` needs to be patched separately in each module.


civis/io/_databases.py
```diff
@@ -41,7 +46,12 @@ def query_civis(sql, database, api_key=None, credential_id=None,
     >>> run = query_civis(sql="DELETE schema.table", database='database')
     >>> run.result()  # Wait for query to complete
     """
-    client = APIClient(api_key=api_key)
+    if api_key is not None:
+        warnings.warn('The "api_key" parameter is deprecated and will be '
+                      'removed in v2. Please use the `client` parameter '
+                      'instead.', FutureWarning)
+    if client is None:
+        client = APIClient(api_key=api_key, resources='all')
```
keithing: We shouldn't use `resources='all'` in a default position.  In CivisFuture we have to for pubnub, but here we ought to use `resources='base'`.  The long term fix is to get the required endpoints for CivisFuture into the base resources.

civis/base.py
```diff
@@ -134,22 +135,29 @@ class CivisAsyncResultBase(futures.Future):
...
                  poll_on_creation=True):
         super().__init__()
         self.poller = poller
         self.poller_args = poller_args
         self.polling_interval = polling_interval
-        self.api_key = api_key
+        if api_key is not None:
+            warnings.warn('The "api_key" parameter is deprecated and will be '
+                          'removed in v2. Please use the `client` parameter '
+                          'instead.', FutureWarning)
```
keithing: this warning is in a lot of places, might be nice to create a single function and call that instead.

keithing:
This is a lot cleaner. Thanks for getting to is so quickly!

civis/base.py
stephen-hoover: @keithing , before I finish changing everything, is a905aeacd1f7e98cbfb1ad5fefdd1d31595475ff what you had in mind? That's what I think when I hear "single function to do deprecations".

keithing: This is actually more general than I had in mind.  I was actually thinking of just a small function like `warn_if_api_key(api_key)`.  That said, I think a905aea is pretty nice too.  Maybe needs a more specific name (`deprecate_param` or `warn_param_removal`, etc) and small amount of documentation, but otherwise I like it.

stephen-hoover: Okay, I've used this decorator for the deprecations.

civis/io/_databases.py
stephen-hoover: I removed `resources='all'` where I could, but most places still require it because they put the client into `CivisFuture`. What's the downside to using `resources='all'`?

civis/base.py
keithing: this is nice.  I think we can use this for `archive` as well (in another PR).

civis/io/_databases.py
keithing: `resources='all'` has the potential to add methods to `APIClient` that are provisional.  These methods may either (a) be subject to sudden breaking changes or (b) not be available to all api-key holders.  It's not a major issue, but we'll want to get everything we need into base resources so we can pull out `resources='all'` from this codebase eventually.

keithing:
LGTM!

stephen-hoover:
@keithing , apologies, I realized I forgot to add the decorator to the changelog. Reapprove? I will squash everything into one commit unless you want a different arrangement.

keithing:
New commit also looks good.  Feel free to merge.

PR #47, opened 2017-03-16
ENH Improve wait-for-job completion in jobs wrappers

stephen-hoover:
Add a `wait` parameter to the `utils.run_jobs`; users often want to immediately wait for completion after starting a run, so make that easy. Similarly, add a `wait_for_run` wrapper which handles creation of the `CivisFuture` and waiting for completion.

### DIFF
--- a/civis/utils/_jobs.py
+++ b/civis/utils/_jobs.py
@@ -4,18 +4,22 @@ from civis.utils._deprecation import deprecate_param
 
 
 @deprecate_param('v2.0.0', 'api_key')
-def run_job(job_id, api_key=None, client=None):
+def run_job(job_id, wait=False, api_key=None, client=None):
     """Run a job.
+
+    A "job" could be a Container Script, a Query, a Python 3 script, etc.
 
     Parameters
     ----------
     job_id : str or int
         The ID of the job.
+    wait : bool
+        If True, block until the run completes.
     api_key : DEPRECATED str, optional
         Your Civis API key. If not given, the :envvar:`CIVIS_API_KEY`
         environment variable will be used.
-    client : :class:`civis.APIClient`, optional
-        If not provided, an :class:`civis.APIClient` object will be
+    client : :class:`~civis.APIClient`, optional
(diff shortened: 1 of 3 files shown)


stephen-hoover:
@keithing , thoughts? I think at least the `wait` parameter is helpful, and I think that `wait_for_run` is too.

civis/utils/_jobs.py
```diff
@@ -26,7 +30,41 @@ def run_job(job_id, api_key=None, client=None):
...
-                       (job_id, run['id']),
-                       client=client,
-                       poll_on_creation=False)
+    fut = CivisFuture(client.jobs.get_runs,
+                      (job_id, run['id']),
+                      client=client,
+                      poll_on_creation=False)
+    if wait:
+        fut.result()
+    return fut
```
keithing: The user will still have to call `.result()` to get the contents, right? Maybe a better solution is to add a method to `CivisFuture` like `.block()` or `.wait()` and recommend that users call this function like
```python
run_job(id, ...).block()
```
to force waiting for completion.

stephen-hoover: The `result` method is blocking. Maybe this is unnecessary and `run_job(job_id).result()` is good enough. ... In fact, that's about the same number of extra characters as `run_job(job_id, wait=True)`. Okay, I can drop the `wait` argument.

stephen-hoover:
Now that I'm thinking about this more, the whole PR seems like it doesn't add a lot of value. I'm happy closing it.

civis/utils/_jobs.py
keithing: Hmm.  On second thought, I'm warming up to this approach.  Returning a future in a "done" state seems pretty reasonable, and adding sugar to CivisFuture like `block()` will be confusing because it is just `.result()` that returns `None`.  I do generally want to avoid adding the parameter `wait` to the higher level functions in preference of suggests chaining `.result()` instead.

stephen-hoover: Are you suggesting that a `wait` parameter here is a good idea, or not?

PR #50, opened 2017-03-20
[CIVP-10635] Adding retries during the call to /endpoints in get_swagger_spec

keithing:
This retries the call to `/endpoints` while instantiating the client for certain errors.   Another possible route might be to pass the `_session` object from `APIClient`, but this would prevent us from being able to memoize `generate_classes`.


skomanduri:
LGTM

PR #51, opened 2017-03-21
[CIVP-10208] Retry all HTTP verbs with a Retry-After header and an approved error

stephen-hoover:
If we have a rate limit error (429) and a Retry-After header, it should always be safe to retry the request, regardless of the verb. The `Retry.is_retry` method starts by checking if the verb is on the whitelist (defaults to ['HEAD', 'GET', 'PUT', 'DELETE', 'OPTIONS', 'TRACE']) and immediately returns False if it is not. This PR changes our code so that we will retry on the `RETRY_AFTER_STATUS_CODES` ([413, 429, 503]) regardless of verb.

See https://github.com/shazow/urllib3/blob/50705abd83be4ad7997eeefe9e732564af1db118/urllib3/util/retry.py#L285L299 .

`urllib3` respects the "Retry-After" header starting at v1.19. We get `urllib3` from `requests`; `requests` updated `urllib3` to a compatible version at v2.12.0. Therefore this PR also needs to increase the version of `requests` which we require.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 Retries to http request in ``get_swagger_spec`` to make calls to ``APIClient`` robust to network failure
 
+### Changed
+- Modified retry behavior so that 413, 429, or 503 errors accompanied by a "Retry-After" header will be retried regardless of the HTTP verb used.
+
 ## 1.4.0 - 2017-03-17
 ### API Changes
 - Deprecate ``api_key`` input to higher-level functions and classes in favor of an ``APIClient`` input. The ``api_key`` will be removed in v2.0.0. (#46)
--- a/README.md
+++ b/README.md
@@ -60,6 +60,19 @@ database = client.databases.list()
 See the [full documentation](https://civis-python.readthedocs.io) for a more
 complete user guide.
 
+## Retries
+
+The API client will automatically retry for certain API error responses.
+
+If the error is one of [413, 429, 503] and the API client is told how long it needs
+to wait before it's safe to retry (this is always the case with 429s, which are
+rate limit errors), then the client will wait the specified amount of time
+before retrying the request.
+
+If the error is one of [429, 502, 503, 504] and the request is not a `patch*` or `post*`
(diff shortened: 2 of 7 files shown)


stephen-hoover:
@keithing , I've verified that I get 429 errors from patching without this PR. After this PR, the client waits and retries on patch requests.

keithing:
Out of curiosity, why not add 'POST' and 'PATCH' to whitelist instead?  Is that approach equivalent?

stephen-hoover:
It's not quite identical, because the `RETRY_AFTER_STATUS_CODES` set is different from the retry list we provide. My understanding is that, at least in principle, users could get some kinds of error codes even if their command had been partially or fully executed. So you don't want to retry POST on all error types, because you might be running the command twice. 

I think that if we get one of the errors in the `RETRY_AFTER_STATUS_CODES` along with a "Retry-After" header, we should be safe to retry all the verbs. I'm not sure why `urllib3` doesn't do that already, but I think it's the right thing to do for us.

stephen-hoover:
Removed the log emit. While digging through code, I realized that `urllib3` already gives log emits for retries.

keithing:
Test Error is unrelated, you might get lucky if you restart Travis and pass.  We'll have to redesign that test.  Changes all LGTM, just want to get one last opinion on the safety of these retries:

@elliothevel, do you agree it's safe to retry POST and PATCH requests after 413, 429 or 503 HTTP error codes?

stephen-hoover:
Ugh, apologies for the bad test. I added a commit to try to fix it. If you prefer, I can make that a separate PR.

keithing:
>If you prefer, I can make that a separate PR.
Let's keep your fix in!  It'll be nice to get that into the codebase.

elliothevel:
> do you agree it's safe to retry POST and PATCH requests after 413, 429 or 503 HTTP error codes?

I'm not certain that 413 and 503 responses will always contain the "Retry-After" header, but I agree that it's safe to retry these requests if the header is there.

stephen-hoover:
I don't expect 413 or 503s from the Civis Platform to include a "Retry-After" at this time. Maybe they could in the future.

keithing:
LGTM

PR #54, opened 2017-03-28
Add file ID to multifile manifest

jmemich:
There is currently no logging available to store this but it is incredibly useful for anyone interested in replicating an analysis on the result of this unload.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -402,11 +402,13 @@ def civis_to_multifile_csv(sql, database, job_name=None, api_key=None,
         raise EmptyResultError("Unload query {} returned no manifest."
                                .format(script_id))
 
+    manifest_file_id = outputs[0]['file_id']
     buf = io.BytesIO()
-    civis_to_file(outputs[0]['file_id'], buf)
+    civis_to_file(manifest_file_id, buf)
     txt = io.TextIOWrapper(buf, encoding='utf-8')
     txt.seek(0)
     unload_manifest = json.load(txt)
+    unload_manifest['file_id'] = manifest_file_id
 
     return unload_manifest
 
--- a/civis/tests/test_io.py
+++ b/civis/tests/test_io.py
@@ -155,7 +155,7 @@ class ImportTests(CivisVCRTestCase):
         sql = "SELECT * FROM scratch.api_client_test_fixture"
         result = civis.io.civis_to_multifile_csv(
             sql, database='redshift-general', polling_interval=POLL_INTERVAL)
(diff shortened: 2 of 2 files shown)


civis/io/_tables.py
```diff
@@ -402,11 +402,13 @@ def civis_to_multifile_csv(sql, database, job_name=None, api_key=None,
...
                                .format(script_id))
 
+    manifest_file_id = outputs[0]['file_id']
     buf = io.BytesIO()
-    civis_to_file(outputs[0]['file_id'], buf)
+    civis_to_file(manifest_file_id, buf)
     txt = io.TextIOWrapper(buf, encoding='utf-8')
     txt.seek(0)
     unload_manifest = json.load(txt)
+    unload_manifest['file_id'] = manifest_file_id
```
stephen-hoover: How about "manifest_file_id"? Given how many file IDs are floating around in a manifest JSON, it's not obvious what this one would refer to.

jmemich:
I'm torn: on the one hand I think that `civis_to_csv` should, indeed, leave the user with a corresponding CSV at the file path but, on the other hand, we'll have to update the function anyways because we'll probably want to pass CSV settings to take advantage of changes to platform's export code. The change to CSV export function to allow CSV settings could leave the user with the possibility of triggering a multifile unload which would result in a manifest -- not a single CSV.

Originally, I separated these functions because I wanted users who know they were after a multifile unload to get exactly that without wanting to overload the `civis_to_csv` function. Now, I'm no longer sure that's the right approach, although I'm not sure allowing for the possibility that `civis_to_csv` could fail to write a CSV to disk is the best idea here. Thoughts?

jmemich:
After talking with @elliothevel I think it would make sense to keep this function and `civis_to_csv` separate, with the key difference being that the latter expects a CSV and writes it to disk whereas this function returns a manifest. Both functions could ignore any user given `force_multifile` parameter.

The question remains about adding a file ID to the manifest. @stephen-hoover if you'd like to keep the result in line with what a user would expect from a regular client API call, we could leave this manifest as is and this could very easily be re-uploaded to the files endpoint where needed. Otherwise, it makes sense to add something like `manifest_file_id` here.

stephen-hoover:
I agree that it makes sense to keep `civis_to_csv` and `civis_to_multifile_csv` separate. One of my concerns with adding additional keys to the output is that it then becomes confusing to users which parts come from Platform and which from the client. That could also cause confusion when using, e.g., the (forthcoming) R API client.

jmemich:
That seems reasonable. I'll close this and make a separate PR to add `csv_settings` where applicable in the `io` module.

PR #55, opened 2017-03-29
Add CSV settings options to civis.io module

jmemich:
No functional changes. This will be required to take advantage of changes to platform exports.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -233,8 +233,7 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
 def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
                  client=None, credential_id=None, include_header=True,
                  compression='none', delimiter=',', unquoted=False,
-                 prefix=None, archive=False, hidden=True,
-                 polling_interval=None):
+                 archive=False, hidden=True, polling_interval=None):
     """Export data from Civis to a local CSV file.
 
     Parameters
@@ -258,8 +257,8 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
         The ID of the database credential.  If ``None``, the default
         credential will be used.
     include_header: bool, optional
-        If ``True`` include a key in the returned dictionary containing a list
-        of column names. Default: ``True``.
+        If ``True``, the first line of the CSV will be headers.
+        Default: ``True``.
     compression: str, optional
         Type of compression to use, if any. One of ``'none'``, ``'zip'``, or
         ``'gzip'``. Default ``'none'``.
@@ -267,10 +266,7 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
         Which delimiter to use, if any. One of ``','``, ``'\t'``, or
         ``'|'``. Default: ``','``.
     unquoted: bool, optional
-        Where or not to quote fields. Default: ``False``.
-    prefix: str, optional
-        A user specified filename prefix for the output file to have. Default:
-        ``None``.
+        Whether or not to quote fields. Default: ``False``.
     polling_interval : int or float, optional
         Number of seconds to wait between checks for query completion.
     archive : bool, optional (deprecated)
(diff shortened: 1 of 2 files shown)


civis/io/_tables.py
```diff
@@ -255,6 +257,20 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
...
+        If ``True`` include a key in the returned dictionary containing a list
+        of column names. Default: ``True``.
+    compression: str, optional
+        Type of compression to use, if any. One of ``'none'``, ``'zip'``, or
+        ``'gzip'``. Default ``'none'``.
+    delimiter, str: optional
+        Which delimiter to use, if any. One of ``','``, ``'\t'``, or
+        ``'|'``. Default: ``','``.
+    unquoted: bool, optional
+        Where or not to quote fields. Default: ``False``.
```
keithing: Whether or not

civis/io/_tables.py
```diff
@@ -255,6 +257,20 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
     credential_id : str or int, optional
         The ID of the database credential.  If ``None``, the default
         credential will be used.
+    include_header: bool, optional
+        If ``True`` include a key in the returned dictionary containing a list
+        of column names. Default: ``True``.
```
keithing: Maybe something like "If ``True``, the first line of the csv will be headers"

civis/io/_tables.py
keithing: Might be clearer to state "When ``False``, the default, fields will be quoted"

civis/io/_tables.py
```diff
@@ -255,6 +257,20 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
...
+        Type of compression to use, if any. One of ``'none'``, ``'zip'``, or
+        ``'gzip'``. Default ``'none'``.
+    delimiter, str: optional
+        Which delimiter to use, if any. One of ``','``, ``'\t'``, or
+        ``'|'``. Default: ``','``.
+    unquoted: bool, optional
+        Where or not to quote fields. Default: ``False``.
+    prefix: str, optional
+        A user specified filename prefix for the output file to have. Default:
+        ``None``.
```
keithing: Is this only used internally in the Platform? Might be a bit confusing, since the user specifies a `filename` directly (and this doesn't affect that).

keithing:
Can you add a note to `CHANGELOG.md` as well?

civis/io/_tables.py
keithing: This doc change should be made in `civis_to_multifile_csv` as well.

civis/io/_tables.py
jmemich: Yeah, this is for when the files end up on platform's S3 bucket. Good catch, I'll remove.

stephen-hoover: I believe this does affect the name of the output Civis File objects, FWIW.

stephen-hoover:
Would it be worth letting users specify a `csv_settings` dictionary directly, instead of adding a parameter for each setting we want to expose?

civis/io/_tables.py
```diff
@@ -283,9 +299,19 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
                       "Use `hidden` instead.", FutureWarning)
     if client is None:
         client = APIClient(api_key=api_key, resources='all')
-    script_id, run_id = _sql_script(client, sql, database,
-                                    job_name, credential_id,
-                                    hidden=hidden)
+
+    delimiter = DELIMITERS.get(delimiter)
+    assert delimiter, "delimiter must be one of {}".format(DELIMITERS.keys())
```
stephen-hoover: This should be a `ValueError`.

jmemich:
@stephen-hoover I find this both more opaque and I would rather not give users the opportunity to mis-specify a key when we can do it for them, correctly. That OK with you?

keithing:
LGTM

PR #57, opened 2017-04-17
[CIVP-10738] Fixed pubnub reconnect policy

menglewis:
This should be the correct API from Pubnub as per https://www.pubnub.com/docs/python/api-reference


stephen-hoover:
@menglewis , could you make a new PR to add a note to the changelog about this?

PR #58, opened 2017-04-17
[CIVP-10864] ENH: Error on file too large.

jseabold:
We can't upload files larger than 5GB to s3. Also, AWS provides more informative errors than what requests will raise by default.

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -76,7 +76,11 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
         # This streams from the open file buffer without holding the
         # contents in memory.
         en = MultipartEncoder(fields=form_key)
-        if en.len / 2 ** 20 < 100:
+        # The refusal error from AWS states 5368730624 is the max size allowed
+        if en.len >= 5 * 2 ** 30:  # 5 GB
+            msg = "Cannot upload files greater than 5GB. Got {:d}."
+            raise ValueError(msg.format(en.len))
+        elif en.len <= 100 * 2 ** 20:  # 100 MB
             # Semi-arbitrary cutoff for "small" files.
             # Send these with requests directly because that uses less CPU
             response = requests.post(url, files=form_key)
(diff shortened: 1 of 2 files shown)


civis/io/_files.py
```diff
@@ -76,7 +97,11 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
         # This streams from the open file buffer without holding the
         # contents in memory.
         en = MultipartEncoder(fields=form_key)
-        if en.len / 2 ** 20 < 100:
+        # The refusal error from AWS states 5368730624 is the max size allowed
+        if en.len >= 5 * 2 ** 30:  # 5 GB
+            msg = "Cannot uploaded files greater than 5GB. Got {:d}."
```
stephen-hoover: "uploaded" --> "upload"

Does the AWS error happen immediately, or only after the user has uploaded 5 GB? If it's immediate, it seems better to let the attempt go through and show the user AWS's error message.

jseabold:
On Mon, Apr 17, 2017 at 10:54 AM, Stephen Hoover <notifications@github.com>
wrote:

> *@stephen-hoover* requested changes on this pull request.
>
> This also needs an entry in the changelog.
> ------------------------------
>
> In civis/io/_files.py
> <https://github.com/civisanalytics/civis-python/pull/58#discussion_r111760849>
> :
>
> > @@ -76,7 +97,11 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
>          # This streams from the open file buffer without holding the
>          # contents in memory.
>          en = MultipartEncoder(fields=form_key)
> -        if en.len / 2 ** 20 < 100:
> +        # The refusal error from AWS states 5368730624 is the max size allowed
> +        if en.len >= 5 * 2 ** 30:  # 5 GB
> +            msg = "Cannot uploaded files greater than 5GB. Got {:d}."
>
> "uploaded" --> "upload"
>
Fixed.


> Does the AWS error happen immediately, or only after the user has uploaded
> 5 GB? If it's immediate, it seems better to let the attempt go through and
> show the user AWS's error message.
>
It does not happen immediately.

PR #59, opened 2017-04-18
[CIVP-10867] Use cache to create APIClient

keithing:
This PR adds the parameter `local_api_spec` to `APIClient` to allow users to create a client object from a local cache of the Civis API specification.  This parameter can be either an OrderedDict or a string which points to a file of the specification (parsing expects an order to the specification, hence why a regular dict cannot be used).

I'm also exposing `get_swagger_spec` in `civis.resources` as the intended workflow is to call `get_swagger_spec` to download the specification for caching.

### DIFF
--- a/civis/resources/__init__.py
+++ b/civis/resources/__init__.py
@@ -1,3 +1,3 @@
-from ._resources import generate_classes
+from ._resources import generate_classes, get_api_spec
 
-__all__ = ["generate_classes"]
+__all__ = ["generate_classes", "get_api_spec"]
--- a/civis/resources/__init__.py
+++ b/civis/resources/__init__.py
@@ -4,4 +4,4 @@ from ._resources import (generate_classes,
 
 __all__ = ["generate_classes",
            "get_api_spec",
-           "generate_classes_maybed_cached"]
+           "generate_classes_maybe_cached"]
--- a/civis/resources/__init__.py
+++ b/civis/resources/__init__.py
@@ -1,3 +1,7 @@
-from ._resources import generate_classes, get_api_spec
+from ._resources import (generate_classes,
+                         get_api_spec,
+                         generate_classes_maybe_cached)
 
-__all__ = ["generate_classes", "get_api_spec"]
+__all__ = ["generate_classes",
+           "get_api_spec",
+           "generate_classes_maybed_cached"]
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -438,23 +438,23 @@ def parse_api_spec(api_spec, api_version, resources):
 
 
 @lru_cache(maxsize=4)
-def get_api_spec(api_key, user_agent, api_version):
+def get_api_spec(api_key, api_version="1.0"):
     """Download the Civis API specification.
 
     Parameters
     ----------
     api_key : str
         Your API key obtained from the Civis Platform.
-    user_agent : str
-        The user agent used in the request to get api endpoint specification.
     api_version : string, optional
         The version of endpoints to call. May instantiate multiple client
         objects with different versions.  Currently only "1.0" is supported.
     """
+    civis_version = civis.__version__
     session = requests.Session()
     session.auth = (api_key, '')
-    if user_agent:
-        session.headers.update({"User-Agent": user_agent.strip()})
+    session_agent = session.headers.get('User-Agent', '')
+    user_agent = "civis-python/{} {}".format(civis_version, session_agent)
+    session.headers.update({"User-Agent": user_agent.strip()})
     max_retries = AggressiveRetry(MAX_RETRIES, backoff_factor=.75,
                                   status_forcelist=civis.civis.RETRY_CODES)
     adapter = HTTPAdapter(max_retries=max_retries)
@@ -473,7 +473,7 @@ def get_api_spec(api_key, user_agent, api_version):
 
 
 @lru_cache(maxsize=4)
-def generate_classes(api_key, user_agent, api_version="1.0", resources="base"):
+def generate_classes(api_key, api_version="1.0", resources="base"):
     """ Dynamically create classes to interface with the Civis API.
 
     The Civis API documents behavior using an OpenAPI/Swagger specification.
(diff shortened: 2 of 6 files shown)


civis/resources/_resources.py
```diff
@@ -404,9 +405,25 @@ def parse_path(path, operations, api_version, resources):
...
-    """ Parse a swagger specifiction into a dictionary of classes
+    """ Dynamically create classes to interface with the Civis API.
+
+    Parse an OpenAPI (Swagger) specification into a dictionary of classes
     where each class represents an endpoint resource and contains
     methods to make http requests on that resource.
+
+    Parameters
+    ----------
+    api_spec : OrderedDict
```
stephen-hoover: This should be "swagger" instead of "api_spec", no? Although I think that `api_spec` is a better parameter name, so it would be preferable to change the code to match the docs.

civis/resources/_resources.py
```diff
@@ -421,10 +438,24 @@ def parse_swagger(swagger, api_version, resources):
...
 @lru_cache(maxsize=4)
-def get_swagger_spec(api_key, user_agent, api_version):
+def get_swagger_spec(api_key, user_agent=None, api_version="1.0"):
+    """Download the Civis API specification.
+
+    Parameters
+    ----------
+    api_key : str
+        Your API key obtained from the Civis Platform. If not given, the
+        client will use the :envvar:`CIVIS_API_KEY` environment variable.
```
stephen-hoover: At this point in the code, using the environment variable is no longer an option.

civis/resources/_resources.py
```diff
@@ -421,10 +438,24 @@ def parse_swagger(swagger, api_version, resources):
...
+    user_agent : str
+        The user agent used in the request to get api endpoint specification.
```
stephen-hoover: What's a "user agent"?

civis/resources/_resources.py
```diff
@@ -476,3 +507,22 @@ def generate_classes(api_key, user_agent, api_version="1.0", resources="base"):
...
+        if isinstance(cache, OrderedDict):
+            raw_spec = cache
+        elif isinstance(cache, str):
+            with open(cache, "r") as f:
+                raw_spec = json.load(f, object_pairs_hook=OrderedDict)
+        else:
+            msg = "cache must be an OrderedDict or str, given {}"
+            raise ValueError(msg.format(type(cache)))
+        spec = JsonRef.replace_refs(raw_spec)
+        classes = parse_swagger(spec, api_version, resources)
```
stephen-hoover: How long does this `parse_swagger` step take? I thought I remembered it as taking a few hundred ms. If so, this function should probably be memoized.

civis/tests/test_resources.py
```diff
@@ -245,3 +245,38 @@ def test_create_method_unexpected_kwargs():
     with pytest.raises(TypeError) as excinfo:
         method(mock_endpoint, foo=0, bar=0, baz=0)
     assert str(excinfo.value) == expected_msg
+
+
+@mock.patch('builtins.open', new_callable=mock.mock_open,
+            read_data='{"test": true}')
+@mock.patch('civis.resources._resources.generate_classes')
+@mock.patch('civis.resources._resources.parse_swagger')
```
stephen-hoover: For these two lines, you should also pass `autospec=True`. https://docs.python.org/3/library/unittest.mock.html#autospeccing

civis/tests/test_resources.py
```diff
@@ -245,3 +245,38 @@ def test_create_method_unexpected_kwargs():
...
+                                             api_version, resources)
+    mock_gen.assert_called_once_with(api_key, user_agent, api_version,
+                                     resources)
+
+    # Handles OrderedDict
+    spec = OrderedDict({"test": True})
+    _resources.generate_classes_maybe_cached(spec, api_key, user_agent,
+                                             api_version, resources)
+    mock_parse.assert_called_once_with(spec, api_version, resources)
+
```
stephen-hoover: Is it worth also asserting that `mock_gen` hasn't been called?

civis/resources/__init__.py
```diff
@@ -1,3 +1,3 @@
-from ._resources import generate_classes
+from ._resources import generate_classes, get_swagger_spec
 
-__all__ = ["generate_classes"]
+__all__ = ["generate_classes", "get_swagger_spec"]
```
stephen-hoover: You should also add "generate_classes_maybe_cached" so that you don't need to import it from a private module.

keithing: Yeah, I think you're right. I was debating the best way to do this since I think `get_swagger_spec` would be useful outside of `civis` while `generate_classes_maybe_cache` really isn't.

civis/resources/_resources.py
keithing: Good catch.

civis/resources/_resources.py
keithing: What do you think about this description?

> The User-Agent string passed in the HTTP request to get the Civis API specification.  If None, uses the default from the requests library.

stephen-hoover: That's definitely better. Do you have an example of one of the strings? Why would anyone want to change this from the default?

civis/resources/_resources.py
keithing: I started down the road of changing the language around swagger but I was worried there were too many changes to include it in the same PR as adding the cache functionality.  It occurs to me now that the best way to do this would be to first change that language in a separate PR and then rebase this on top.

stephen-hoover: I'm also fine switching the docstring to say "swagger". If you think it's worth another PR to improve the language, I'm happy to review it.

civis/resources/_resources.py
keithing: > Why would anyone want to change this from the default?

We use this function in the library and pass an altered User-Agent with `"civis-python/[VERSION]"`.  If you were a bit perplexed by the parameter, users will be too.  I'll make a small refactor to just default to the `"civis-python/[VERSION]"` User-Agent and remove the parameter.

civis/resources/_resources.py
keithing: Made those changes and rebased.

civis/resources/_resources.py
keithing: `lru_cache` won't be able to memoize this since `collections.OrderedDict` isn't hashable.  Unless you feel strongly, I think it's not worth rolling our own memoization function.

PR #60, opened 2017-04-18
REF: Refactoring out use of 'swagger'

keithing:
The Swagger Specification has been renamed to OpenAPI Specification.  This PR removes references of "swagger" to make the code more clear.

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -403,12 +403,12 @@ def parse_path(path, operations, api_version, resources):
     return class_name, methods
 
 
-def parse_swagger(swagger, api_version, resources):
-    """ Parse a swagger specifiction into a dictionary of classes
+def parse_api_spec(api_spec, api_version, resources):
+    """ Parse the Civis API specifiction into a dictionary of classes
     where each class represents an endpoint resource and contains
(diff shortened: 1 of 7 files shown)


stephen-hoover:
Looks like there's at least one reference to `parse_swagger` remaining.

civis/resources/_resources.py
```diff
@@ -403,12 +403,12 @@ def parse_path(path, operations, api_version, resources):
     return class_name, methods
 
 
-def parse_swagger(swagger, api_version, resources):
-    """ Parse a swagger specifiction into a dictionary of classes
+def parse_api_spec(api_spec, api_version, resources):
+    """ Parse the Civis API specifiction into a dictionary of classes
```
stephen-hoover: The word "specification" is misspelled. (Unrelated, I know. I think it's still worth correcting here.)

keithing:
>I see that the "civis_api_spec.json" file contains the key/value pair "swagger": "2.0". That's also in "civis/tests/petstore.json". Are those still correct?

Yes, those are the caches of the API spec used in testing. They haven't changed.  The `/endpoints` endpoint still returns a json with `swagger`.

PR #62, opened 2017-04-20
DOC Add note about index to dataframe_to_civis

### LINKED ISSUE #61: parameterize index in civis.io.dataframe_to_civis
@keithing I use `civis.io.dataframe_to_civis` to write to redshift, and had meaningfull indexes but these weren't written and it doesn't appear there's the ability at all to write them. We may want to make this a parameter (which would get passed to to_csv).

stephen-hoover:
Clarify that the index is not stored, and point users to `reset_index`.

Closes #61 .


keithing:
Thanks, LGTM!

stephen-hoover:
@nickahamed , would this extra documentation have helped you?

PR #63, opened 2017-04-25
[CIVP-10738] Poll on a lower interval when PubNub fails

menglewis:
* Added `_reset_polling_thread` to `PollableResult` which stops the current polling thread and creates a new one with the specified polling interval (defaults to the default interval of 15)
* Added a status handler in the PubNub listener class which will check if the status matches a set of statuses that pertain to unexpected disconnects or network errors and call the `_reset_polling_thread` function if it is disconnected.


menglewis:
@stephen-hoover Mind reviewing this when you have a chance?

PR #67, opened 2017-05-02
BUG remove argument that prevents sphinx from building docs

xdavio:
Sphinx's `make html` was failing for me, and I think it's because `user_agent` is no longer an argument for `generate_classes`.


keithing:
Good catch.  Thanks.  LGTM.

PR #68, opened 2017-05-02
[CIVP-10806] Add more utility functions for working with Civis files

stephen-hoover:
Add the functions `civis.io._files.file_id_from_run_output`, `civis.io._files.file_to_dataframe`, and `civis.io._files.file_to_json`. These functions will be used for the forthcoming modeling interface code, but are also of general interest.

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -1,16 +1,29 @@
 from collections import OrderedDict
+import io
+import json
+import logging
+import os
+import re
 
 import requests
 from requests import HTTPError
 
-from civis import APIClient
-from civis.base import EmptyResultError
+from civis import APIClient, find_one
+from civis.base import CivisAPIError, EmptyResultError
 from civis.utils._deprecation import deprecate_param
 try:
     from requests_toolbelt.multipart.encoder import MultipartEncoder
     HAS_TOOLBELT = True
 except ImportError:
     HAS_TOOLBELT = False
+try:
+    import pandas as pd  # NOQA
+except ImportError:
+    pass
+
+log = logging.getLogger(__name__)
+__all__ = ['file_to_civis', 'civis_to_file', 'file_id_from_run_output',
+           'file_to_dataframe', 'file_to_json']
 
 
 def _get_aws_error_message(response):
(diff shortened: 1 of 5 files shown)


stephen-hoover:
@skomanduri , @keithing , would one of you be able to review this?

civis/tests/test_io.py
```diff
@@ -200,3 +208,114 @@ def test_download_file(self, *mocks):
             with open(tmp.name, "r") as f:
                 data = f.read()
         assert data == expected
+
+
+def test_file_id_from_run_output_exact():
+    m_client = Mock()
+    m_client.scripts.list_containers_runs_outputs.return_value = \
```
keithing: Why not continue the line with 
```python
... = [
Response(...)
]
```
?

civis/io/_files.py
```diff
@@ -167,3 +180,154 @@ def _get_url_from_file_id(file_id, client):
...
+    ------
+    ImportError
+        If ``pandas`` is not available
+
+    See Also
+    --------
+    :func:`~pandas.read_csv`
+    """
+    # Make sure we have pandas; this will raise an ImportError if not.
+    import pandas  # NOQA
```
keithing: For constancy with other places in the code we should set a global `HAS_PANDAS` in the `except ImportError` above and raise an error here like

```python
if not HAS_PANDAS:
    raise ImportError('file_to_dataframe requires pandas to be installed.')
```

civis/io/_files.py
```diff
@@ -167,3 +180,154 @@ def _get_url_from_file_id(file_id, client):
...
+    file_url = file_info.file_url
+    file_name = file_info.name
+    if compression == 'infer':
+        for ext in ['.bz2', '.zip', '.xz']:
+            if os.path.splitext(file_name)[-1] == ext:
+                compression = ext[1:]
+                break
+        else:
+            if os.path.splitext(file_name)[-1] == '.gz':
+                compression = 'gzip'
```
keithing: Do we want compression to be `None` if nothing matches? Something like

```python
compression_exts = {'.gz': 'gzip', '.xz': 'xz', '.bz2': 'bz2', '.zip': 'zip'}
ext = os.path.splitext(file_name)[-1]
compression = compression_exts.get(ext)
```

stephen-hoover: I'd rather leave it as "infer" if we can't figure it out. In practice, I think we'll always end up with no compression in that case, but I believe pandas does check http headers to see if any compression format is indicated there.

Your code suggestion is still much clearer than what I wrote, so I changed to what you suggested (with an extra `if`).

civis/io/_files.py
stephen-hoover: Done.

civis/tests/test_io.py
stephen-hoover: Yes, that's better. Changed.

keithing:
Awesome.  I imagine these functions will be useful.  LGTM!

PR #69, opened 2017-05-03
[CIVP-10806] Add ModelFuture

stephen-hoover:
This is the next step in adding the new modeling interface. This PR adds the new `civis.ml` namespace, and populates it with `ModelFuture`, a subclass of `CivisFuture`. The `ModelFuture` object knows how to find and download results from a CivisML model. It also transports errors from Civis and re-raises them in the local environment to make debugging easier.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -291,8 +291,7 @@ class ModelFuture(CivisFuture):
             # but showing the resulting KeyError can be confusing and
             # mask the real error.
             warnings.warn("Received malformed metadata from Civis Platform. "
-                          "Something went wrong with execution. "
-                          "Please report this error.")
+                          "Something went wrong with job execution.")
 
     def cancel(self):
         """Submit a request to cancel the container/script/run.
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -1,25 +1,32 @@
-"""Running CivisML jobs and retrieving the results
+"""Run CivisML jobs and retrieve the results
 """
-from concurrent import futures
-from functools import wraps
 import io
 import os
 import tempfile
 import threading
-from typing import Callable, Dict, Union
 import warnings
+from concurrent import futures
+from functools import wraps
+from typing import Callable, Dict, Union
 
 try:
     import pandas as pd
     HAS_PANDAS = True
 except ImportError:
     HAS_PANDAS = False
+try:
+    import joblib
+    HAS_JOBLIB = True
+except ImportError:
+    HAS_JOBLIB = False
 
 from civis import APIClient, find_one
 from civis.base import CivisAPIError, CivisJobFailure
-from civis.io import civis_to_file, file_to_civis
+import civis.io as cio
 from civis.futures import CivisFuture
 from civis.polling import _ResultPollingThread
+
+__all__ = ['ModelFuture', 'ModelError']
 
 
 class ModelError(RuntimeError):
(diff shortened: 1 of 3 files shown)


stephen-hoover:
@keithing , I've split this PR into two commits. The first copy-pastes already-reviewed code from the modeling project repository, and the second modifies it to live in the API client and to work with the functions I added in #68 . If you review the commits separately, you can look at the first one at a higher level and assume that the details work.

stephen-hoover:
Haha. Our test environment is Python 3.4.2. I'll remove the type hints (those require Python >= 3.5.0).

civis/ml/_model.py
```diff
@@ -0,0 +1,443 @@
...
+            # then there's no improvements to make on
+            # any existing exception.
+            pass
+        except KeyError:
+            # KeyErrors always represent a bug in the modeling code,
+            # but showing the resulting KeyError can be confusing and
+            # mask the real error.
+            warnings.warn("Received malformed metadata from Civis Platform. "
+                          "Something went wrong with execution. "
+                          "Please report this error.")
```
keithing: What's going on here?  Can we directly tell if this issue has occurred? I would prefer not to ask users to be reporting errors if we can avoid it.

civis/ml/_model.py
```diff
@@ -0,0 +1,443 @@
...
+        buf.seek(0)
+        file_id = cio.file_to_civis(buf, name=civis_fname, client=client)
+    else:
+        with open(X) as _fin:
+            file_id = cio.file_to_civis(_fin, name=civis_fname, client=client)
+
+    return file_id
+
+
+def _decode_train_run(train_job_id, train_run_id, client):
```
keithing: I don't see where these private functions are being called: `_decode_train_run`, `_stash_local_data` and `_check_is_fit`.  Are they only used in testing?

civis/ml/_model.py
```diff
@@ -0,0 +1,443 @@
...
+    """Download a Civis file using a reference on a previous run
+
+    Parameters
+    ----------
+    fname: str
+        Name of the Civis file you wish to retrieve
+    job_id, run_id: int
+        Job ID and run ID of the run holding a reference to this file
+    local_dir: str
+        Download the file `fname` from Civis Platform to this directory
```
keithing: `client` param is missing.  Might not even need param docs on the private function here.

civis/ml/_model.py
stephen-hoover: That should never actually occur. If we don't handle it, then the user would get some garbled exception that won't make much sense. Also, it wouldn't necessarily need to halt execution of whatever this is running in. I wanted to let the user know that something was wrong without forcing a stop. Should I remove the "Please report this error." line and leave the rest? Do something else?

civis/ml/_model.py
stephen-hoover: Apologies. Those will be used in the next PR. I could remove them from this PR and let you review them separately if you prefer.

civis/ml/_model.py
stephen-hoover: I removed the parameters section of this docstring, since it's a private function.

stephen-hoover:
@keithing , your questions made me realize that I'd forgotten the "Parameters" section for `ModelFuture`. I added that.

civis/ml/_model.py
keithing: Oh, no that's fine.  I thought that might be the case, but wanted to check.

civis/ml/_model.py
keithing: Okay, perfect. Let's remove that line, but we don't need to do anything else.

stephen-hoover: Removed!

PR #70, opened 2017-05-04
[CIVP-10806] Add ModelPipeline class

stephen-hoover:
Add the `ModelPipeline` class, the primary abstraction for creating and running modeling jobs in the Civis Platform.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -2,6 +2,8 @@
 """
 from collections import namedtuple
 import io
+import json
+import logging
 import os
 import tempfile
 import threading
@@ -19,6 +21,11 @@ try:
     HAS_JOBLIB = True
 except ImportError:
     HAS_JOBLIB = False
+try:
+    from sklearn.base import BaseEstimator
+    HAS_SKLEARN = True
+except ImportError:
+    HAS_SKLEARN = False
 
 from civis import APIClient, find_one
 from civis.base import CivisAPIError, CivisJobFailure
@@ -26,7 +33,8 @@ import civis.io as cio
 from civis.futures import CivisFuture
 from civis.polling import _ResultPollingThread
 
-__all__ = ['ModelFuture', 'ModelError']
+__all__ = ['ModelFuture', 'ModelError', 'ModelPipeline']
+log = logging.getLogger(__name__)
 
 # sentinel value for default primary key value
 SENTINEL = namedtuple('Sentinel', [])()
@@ -573,21 +581,12 @@ class ModelPipeline:
     train_template_id = 8387
     predict_template_id = 8388
 
-    def __init__(self,
-                 model: Union[str, 'sklearn.base.BaseEstimator'],
-                 dependent_variable: Union[str, Sequence[str]],
-                 *,
-                 primary_key: str=None,
-                 parameters: Dict=None,
-                 cross_validation_parameters: Dict=None,
-                 model_name: str=None,
-                 calibration: str=None,
-                 excluded_columns: Sequence[str]=None,
-                 client: APIClient=None,
-                 cpu_requested: int=None,
-                 memory_requested: int=None,
-                 disk_requested: float=None,
-                 verbose: bool=False):
+    def __init__(self, model, dependent_variable, *,
+                 primary_key=None, parameters=None,
+                 cross_validation_parameters=None, model_name=None,
+                 calibration=None, excluded_columns=None, client=None,
+                 cpu_requested=None, memory_requested=None,
+                 disk_requested=None, verbose=False):
         self.model = model
         self._input_model = model  # In case we need to modify the input
         if isinstance(dependent_variable, str):
@@ -622,17 +621,11 @@ class ModelPipeline:
         self._client = APIClient(resources='all')
 
     @classmethod
-    def from_existing(cls,
-                      train_job_id: int,
-                      train_run_id: Union[str, int]='latest',
-                      client: APIClient=None) -> 'ModelPipeline':
-        """Create a ``ModelPipeline`` object from existing model IDs
-        """
+    def from_existing(cls, train_job_id, train_run_id='latest', client=None):
+        """Create a ``ModelPipeline`` object from existing model IDs"""
         if client is None:
             client = APIClient(resources='all')
-        train_run_id = decode_train_run(train_job_id,
-                                        train_run_id,
-                                        client)
+        train_run_id = _decode_train_run(train_job_id, train_run_id, client)
         try:
             fut = ModelFuture(train_job_id, train_run_id, client)
             container = client.scripts.get_containers(train_job_id)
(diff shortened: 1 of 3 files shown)


stephen-hoover:
@keithing , as before, the first commit copies reviewed code from the development repo. This commit should make modeling fully functional for anyone with access to the official templates. Future PRs will add Sphinx documentation and switch from using templates to using an API endpoint.

keithing:
>Future PRs will add Sphinx documentation and switch from using templates to using an API endpoint.

What is the time frame on these? Should we make a release before they come in?

stephen-hoover:
Documentation is ready to go. I'll get back to you on the API endpoint.

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+            train_args['WHERESQL'] = sql_where
+        if sql_limit:
+            train_args['LIMITSQL'] = sql_limit
+        if self.excluded_columns:
+            train_args['EXCLUDE_COLS'] = ' '.join(self.excluded_columns)
+        if fit_params:
+            train_args['FIT_PARAMS'] = json.dumps(fit_params)
+
+        if HAS_SKLEARN and isinstance(self.model, BaseEstimator):
+            import joblib
```
keithing: Why put the `import` here rather than in the `try` above where sklearn is imported?

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+
+    @property
+    @_check_is_fit
+    def estimator(self):
+        return self.train_result_.estimator
+
+    @_check_is_fit
+    def predict(self, X=None, table_name=None, database_name=None,
+                manifest=None, file_id=None, sql_where=None, sql_limit=None,
+                primary_key=SENTINEL, output_table=None, output_db=None,
```
keithing: Is the use of `SENTINEL` here to allow users explicitly pass `None`?  And if so, what happens if `primary_key` is `None`?

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+              run.id,
+              client=self._client,
+              polling_interval=polling_interval,
+              poll_on_creation=False,
+              **train_kwargs)
+
+        return fut, container, run
+
+    @property
+    @_check_is_fit
```
keithing: I am correct to say that `_check_is_fit` will *not* raise an error if the pipeline's model has started training, but not completed yet?

If so, I was slightly confused by the name `_check_is_fit` since it seemed to me that fitting had already been completed.  In that case a name like `_check_fit_initiated` might be more clear.

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+        polling_interval : float, optional
+            Check for job completion every this number of seconds.
+            Do not set if using the notifications endpoint.
+
+        Returns
+        -------
+        :class:`~ModelFuture`
+        """
+        if self.train_result_.running():
+            raise RuntimeError('Wait for the training to finish.')
```
keithing: Do you have a use case in mind for raising an error here?  I think it would be okay to simply block until the model finished training.

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+        state = self.__dict__.copy()
+        del state['_client']
+        return state
+
+    def __setstate__(self, state: dict):
+        self.__dict__ = state
+        self._client = APIClient(resources='all')
+
+    @classmethod
+    def from_existing(cls, train_job_id, train_run_id='latest', client=None):
```
keithing: Should this be a separate function outside of `ModelPipeline`?  Do the `ModelPipeline` objects created from this function inherit any state from the object `from_existing` was called from?

I'm thinking that if a user wants to download a previously stored model, they might do so with:

```
pipeline = ModelPipeline('sparse-logistic').from_existing(1001)
```

If the previously trained model was a gradient boosting tree model, the user might be surprised that it's not sparse logistic.

civis/ml/_model.py
```diff
@@ -450,3 +462,518 @@ def training_metadata(self):
...
+
+        Model outputs will always contain out-of-sample scores
+        (accessible through ``ModelFuture.table`` on this function's
+        output), and you may chose to store these out-of-sample scores
+        in a Civis Table with the ``oos_scores``, ``oos_scores_db``,
+        and ``if_exists`` parameters.
+
+        Parameters
+        ----------
+        X : pd.DataFrame or str, optional
```
mheilman: i think if this function is going to be polymorphic, it should have a separate param for every possibility.  With `X`, it's combining two things.  Perhaps remove `X` and have `df` and `csv_path` instead?

civis/ml/_model.py
stephen-hoover: I'll remove and replace with a `HAS_JOBLIB` check.

civis/ml/_model.py
stephen-hoover: That's correct. Passing `None` means "the prediction table doesn't have a primary key". I'll add that to the docstring.

civis/ml/_model.py
stephen-hoover: Correct, it only checks that training has started. Methods that care if it's finished will block themselves. I'll change the name.

civis/ml/_model.py
stephen-hoover: I think you're right. I'll remove these two lines.

civis/ml/_model.py
stephen-hoover: `from_existing` is being used as an alternative constructor here. It seems appropriate to have it be a class method on `ModelPipeline`.  Because it's a class method, it doesn't get any existing instances of the object as input. Having it be a class method like this makes it inheritable; it would create an instance of the subclass.

I added a full docstring with an example here. Does that make it clearer?

civis/ml/_model.py
stephen-hoover: There's already a lot of parameters; I hesitate to make more. In my mind, the thing that defined `X` was "local data". It's easy to tell a file name and a `DataFrame` apart. It's harder to tell a table name apart from a file name. Integer file IDs could probably be distinguished from file names, but I'm less keen on telling two scalars apart.

Would it make more sense to also combine `table_name` and `file_id` into a `data` and `data_source` field? (Maybe with an "infer" default to `data_source`?)

civis/ml/_model.py
keithing: Okay, I think you're right.  If we were ever to do something like an `EnsemblePipeline`, it would be nice to have the same API with `EnsemblePipeline.from_existing`.

civis/ml/_model.py
keithing: Maybe we should split this method into `train` and `train_remote` (where `train` could just upload the data and then call `train_remote`) ?

stephen-hoover: Everything is running remotely. If you provide a CSV or a `DataFrame`, the `ModelPipeline` uploads it to a Civis File (using `_stash_local_data`) and points CivisML at that.

stephen-hoover:
I added a new commit to fix a small issue with the database name being `None`.

civis/ml/_model.py
keithing: I think we'll probably need to compromise on this function to either splitting `X` into `df` and `csv_path` or splitting this function (maybe into `train` and `train_from_local` if we want to default train to be remote).  There is a lot going on in this function so we want to make it as clear to users as possible.  I like the abstraction of `data` and `data_source`, but I think if we are going to create an abstraction of remote vs local, we should use two functions instead of two polymorphic parameters.

stephen-hoover: I would find the `train` and `train_from_local` split to be confusing. It would suggest to me that the two methods are doing something different, when I don't think they really are. I can divide `X` into `df` and `csv_path` if you think that would be clearer.

stephen-hoover:
@keithing , @mheilman , I split `X` into `df` and `csv_path` in both `train` and `predict`.

stephen-hoover:
@mheilman ?

mheilman:
I just had that one comment, which was addressed.  feel free to merge.  i don't need to review the whole thing

PR #72, opened 2017-05-05
[CIVP-10806] DOC Documentation for civis.ml namespace

stephen-hoover:
This adds Sphinx documentation for the new `civis.ml` namespace. This PR includes changes to the doc strings in `civis.ml` to make the formatting nicer on the autodoc'ed Sphinx page.

### DIFF
--- a/docs/source/ml.rst
+++ b/docs/source/ml.rst
@@ -68,7 +68,7 @@ recognizes. You can use code from
 - `scikit-learn <http://scikit-learn.org>`_ v0.18.1
 - `glmnet <https://github.com/civisanalytics/python-glmnet>`_ v2.0.0
 - `xgboost <http://xgboost.readthedocs.io>`_ v0.6a2
-- `muffnn <https://github.com/civisanalytics/muffnn>`_ v1.0.0
+- `muffnn <https://github.com/civisanalytics/muffnn>`_ v1.1.1
 
 When you're assembling your own model, remember that you'll have to make certain that
 either you add a missing value imputation step or that your data doesn't have any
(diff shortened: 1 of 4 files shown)


stephen-hoover:
@keithing , the first commit has been reviewed by @beckermr , but I did enough small formatting changes in the second commit that I think it wouldn't hurt to look through the whole thing.

docs/source/ml.rst
```diff
@@ -0,0 +1,182 @@
...
+
+       
+Pre-Defined Models
+------------------
+
+You can use the following pre-defined models with CivisML.
+All models start by imputing missing values with the mean of non-null
+values in a column. The "sparse_*" models include a LASSO regression step
+(using the `glmnet <https://github.com/civisanalytics/python-glmnet>`_ package)
+to do feature selection before passing data to the final model.
```
keithing: This might be worth a bit of an explanation.  Maybe we should call these `two_stage_*` instead of `sparse_*`?

keithing: I suppose these are parameters in the modeling code, so we should be consistent with those name in the API.  (i.e. let's stick with `sparse_*`).  Not sure if there is a better way to explain the two stage step, unless we asterisk the algorithm in the table below to point out that `LogisticRegression` really isn't the full story?

stephen-hoover: The final stage estimator isn't the full story for any of them, since we're also doing missing value imputation (and categorical expansion, if you count that).  What if I changed "Algorithm" in the table header to "Final Stage Estimator"? Would that be an improvement, or make things worse?

keithing: > The final stage estimator isn't the full story for any of them, since we're also doing missing value imputation

That's a good point.  I think the documentation is clear, we can't avoid that there is a little complexity going on behind the scenes, but it's make everything easier for users.

PR #73, opened 2017-05-07
[CIVP-10391] Add Python 2.7 compatibility

jseabold:


civis/base.py
```diff
@@ -1,7 +1,10 @@
+from __future__ import absolute_import
```
stephen-hoover: Is it worth including `division` with all of the `__future__` imports? I don't know if we're dividing anything anywhere, but that can introduce easy-to-overlook bugs.

civis/cli/__main__.py
```diff
@@ -26,6 +27,11 @@
 from civis.cli._cli_commands import \
     civis_ascii_art, files_download_cmd, files_upload_cmd
 
+try:
+    FileNotFoundError
+except NameError:
+    FileNotFoundError = IOError
```
stephen-hoover: This is probably worth a comment that `FileNotFoundError` isn't present in Python 2. What about defining our own `FileNotFoundError` class? `

civis/cli/__main__.py
```diff
@@ -130,11 +136,12 @@ def param_case_map(param_names):
     return result
 
 
-def invoke(method, path, op, *args, json_output=False, **kwargs):
+def invoke(method, path, op, *args, **kwargs):
```
stephen-hoover: Since `json_output` isn't explicit in the parameters any longer, you should mention it in a top-level comment or doc string.

civis/cli/__main__.py
```diff
@@ -130,11 +136,12 @@ def param_case_map(param_names):
...
 
     # Remove None b/c click passes everything in as None if it's not set.
     kwargs = {k: v for k, v in kwargs.items()
               if v is not None}
+    json_output = kwargs.setdefault('json_output', False)
```
stephen-hoover: The code in this function iterates over `kwargs`, and that never used to contain `json_output`. This line should be something like `json_output = kwargs.pop('json_output', False)` to keep it working the same as it did before.

civis/ml/_model.py
```diff
@@ -798,15 +812,17 @@ def train(self, df=None, csv_path=None, table_name=None,
...
+            fout = os.path.join(tempdir, 'estimator.pkl')
+            joblib.dump(self.model, fout, compress=3)
+            with open(fout, 'rb') as _fout:
+                n = self.model_name if self.model_name else "CivisML"
+                estimator_file_id = cio.file_to_civis(
+                    _fout, 'Estimator for ' + n, client=self._client)
+            self._input_model = self.model  # Keep the estimator
             self.model = str(estimator_file_id)
+            os.remove(fout)
+            os.removedirs(tempdir)
```
stephen-hoover: What about replacing these two lines with `shutil.rmtree(tempdir)`? That should work even if there's more than one file in the directory.

civis/ml/tests/test_model.py
```diff
@@ -123,10 +127,11 @@ def foo(arg):
 
 @mock.patch.object(_model.cio, 'file_to_civis', return_value=-11)
 def test_stash_local_data_from_file(mock_file):
-    with tempfile.TemporaryDirectory() as tempdir:
-        fname = os.path.join(tempdir, 'filename')
+    with tempfile.NamedTemporaryFile() as tempfname:
```
stephen-hoover: It looks like a `NamedTemporaryFile` here won't work on Windows: 

"Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later)." (https://docs.python.org/2/library/tempfile.html?highlight=tempfile#tempfile.NamedTemporaryFile)

Perhaps that doesn't matter in the tests?

civis/ml/_model.py
stephen-hoover: The cleanup should be in a `finally` block so that we clean up temporary files even if there's an exception.

civis/ml/_model.py
```diff
@@ -140,9 +150,12 @@ def _load_estimator(job_id, run_id, filename='estimator.pkl', client=None):
...
         raise ImportError('Install joblib to download models '
                           'from Civis Platform.')
-    with tempfile.TemporaryDirectory() as tempdir:
-        path = _retrieve_file(filename, job_id, run_id, tempdir, client=client)
-        return joblib.load(path)
+    tempdir = tempfile.mkdtemp()
+    path = _retrieve_file(filename, job_id, run_id, tempdir, client=client)
+    obj = joblib.load(path)
+    os.remove(path)
+    os.removedirs(tempdir)
```
stephen-hoover: The cleanup should be in a `finally` block. Maybe `shutil.rmtree` would be better here?

.travis.yml
```diff
@@ -11,4 +13,4 @@ env:
 before_script: flake8 civis
 script:
     - py.test --cov civis
-    - sphinx-build -b html -n docs/source/ docs/build/
+    - if [[ $TRAVIS_PYTHON_VERSION == 3* ]] ; then sphinx-build -b html -n docs/source/ docs/build/ ; fi
```
keithing: This is okay for now.  I believe there is an outstanding bug that docs only build in certain versions of Python 3.  We'll sort that out later since we're hosting docs.

civis/ml/_model.py
jseabold: Done.

civis/ml/_model.py
jseabold: Done.

civis/ml/tests/test_model.py
jseabold: Ah, right. Yeah, I'm going to punt on this. I'll see about adding appveyor support to make sure we're covered for CI on windows soon-ish.

.travis.yml
jseabold: Yeah, I'm not super worried about compatibility here since we host them publicly. Can address later if we want.

civis/cli/__main__.py
jseabold: Yeah, it didn't look like it was used anywhere except the tests and that none of this was intended to be really be user-facing.

jseabold: Done, though I don't know how useful it is without any other accompanying documentation.

civis/cli/__main__.py
jseabold: Done.

civis/cli/__main__.py
jseabold: I just moved it to compat so it's clear.

civis/base.py
jseabold: :man_shrugging: I would hope tests cover this, but I don't know where if anywhere that we're doing float division in here?

jseabold:
I think I addressed all of the comments.

stephen-hoover:
@jseabold , I poked through the documentation and didn't see anything about installation requirements. Maybe it's worth mentioning Python version requirements (Python >= 3.4.2 or Python 2.7, AFAICT) in the "Installation" section of `index.rst`?

civis/compat.py
```diff
@@ -0,0 +1,14 @@
...
+
+import six
+
+if six.PY3:
+    from unittest import mock
+    from functools import lru_cache
+    from inspect import signature
+    FileNotFoundError = FileNotFoundError
+else:
+    import mock
```
stephen-hoover: `mock` is only a requirement for development, but this will error if it's not installed. How about
```python
try:
    import mock
except ImportError:
    pass
```
or else add a `civis/compat_tests.py` module.

stephen-hoover:
There's an `os.makedirs` call in `civis/ml/_model.py` which uses the "exist_ok" parameter. That's not present in Python 2.7. You'll need to remove the parameter and put it behind an `if not os.path.exists`.

civis/compat.py
jseabold: Done.

jseabold:
> There's an `os.makedirs` call in `civis/ml/_model.py` which uses the "exist_ok" parameter. That's not present in Python 2.7. You'll need to remove the parameter and put it behind an `if not os.path.exists`.

Fixed.

jseabold:
Added Python versions to docs

civis/ml/_model.py
```diff
@@ -134,7 +134,8 @@ def _retrieve_file(fname, job_id, run_id, local_dir, client=None):
     """Download a Civis file using a reference on a previous run"""
     file_id = cio.file_id_from_run_output(fname, job_id, run_id, client=client)
     fpath = os.path.join(local_dir, fname)
-    os.makedirs(os.path.dirname(fpath), exist_ok=True)
+    if not os.path.exists(os.path.dirname(local_dir)):
```
stephen-hoover: The `os.path.dirname(local_dir)` here will check a different path than the `os.makedirs` below makes.
```
In [32]: tdir = tempfile.TemporaryDirectory()

In [33]: print(tdir.name, '\n', os.path.dirname(tdir.name))
/var/folders/jn/z8wk4r050bn_zsxmyrhb4gsw0000gn/T/tmpirbgopmi
 /var/folders/jn/z8wk4r050bn_zsxmyrhb4gsw0000gn/T
```

How about
```python
output_dir = os.path.dirname(fpath)
if not os.path.exists(output_dir):
    os.makedirs(output_dir)
```

jseabold: Thinko. Got it.

jseabold: Does fname contain a path? If not, I don't need the dirname because I already have the path?

stephen-hoover: In the one place where this function is called, `fname` doesn't contain a path. Seems okay to use `local_dir` directly.

jseabold: Went with your no assumptions version.

jseabold:
Comments should be addressed.

PR #76, opened 2017-05-09
[CIVP-11088] Use correct job/run ID when fetching training metadata

stephen-hoover:
We had been using the associated job and run IDs, when we should have been using the job and run IDs of the connected training job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
     - "3.4"
 install:
+    - pip install --upgrade pip setuptools
     - pip install -r requirements.txt
     - pip install -r dev-requirements.txt
     - pip install -e .[pubnub]
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -460,8 +460,10 @@ class ModelFuture(CivisFuture):
     @_block_and_handle_missing
     def training_metadata(self):
         if self._train_metadata is None:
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -1,5 +1,5 @@
 pyyaml>=3.0,<=3.99
 click>=6.0,<=6.99
 jsonref>=0.1.0,<=0.1.99
-requests>=2.12.0,==2.*
+requests>=2.12.0,<2.14
```
keithing: Do you know what is incompatible in our code with 2.14?

stephen-hoover: It doesn't look like our code has any incompatibility. It seems to be in the Travis environment. You can look at the errors on the first two commits -- something went wrong on the `pip install` step. I'm still trying to figure it out.

stephen-hoover:
The problem was an out-of-date version of `pip`. I'd assumed that travis change in #73 was 2.7-related, but it wasn't.

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: python
 python:
     - "3.4"
 install:
+    - pip install --upgrade pip setuptools
```
keithing: Good detective work!

stephen-hoover:
@keithing , I rebased to squash that middle commit, but I think that we should keep these last two separate. No conflicts on the rebase. Re-approve?

PR #78, opened 2017-05-10
[CIVP-10806] BUG Allow `from_existing` to handle v0 templates

stephen-hoover:
Older model templates used "WORKFLOW" where the new templates have "MODEL". When loading an old model, check first for "MODEL" and next for "WORKFLOW". It will have one of these if it's a CivisML job.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -665,7 +665,9 @@ class ModelPipeline:
             raise
 
         args = container.arguments
-        model = args['MODEL']
+
+        # Older templates used "WORKFLOW" instead of "MODEL"
+        model = args.get('MODEL', args.get('WORKFLOW'))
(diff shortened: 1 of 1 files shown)


civis/ml/_model.py
```diff
@@ -665,7 +665,9 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
             raise
 
         args = container.arguments
-        model = args['MODEL']
+
+        # Older templates used "WORKFLOW" instead of "MODEL"
+        model = args.get('MODEL', args.get('WORKFLOW'))
```
keithing: This could return `None` if `args` doesn't have `MODEL` or `WORKFLOW`.  What's the implication of that?

jmemich: We have only given this template these two arguments over our codebases development history.

PR #79, opened 2017-05-10
[CIVP-11103] ENH ModelFuture recognize and raies MemoryErrors

stephen-hoover:
Modify the `ModelFuture` to notice when the jobs it's connected to failed because they ran out of memory. If that happened, re-raise an appropriate error so that the user can handle it. We can't do anything about this on the Platform side -- Python dies abruptly without being able to write any outputs.

Before:

screen shot 2017-05-10 at 4 24 59 pm

After:

screen shot 2017-05-10 at 4 24 08 pm

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -151,6 +151,26 @@ def _load_estimator(job_id, run_id, filename='estimator.pkl', client=None):
         return joblib.load(path)
 
 
+def _exception_from_logs(job_id, run_id, client, nlog=15):
+    """Create an exception if the log has a recognizable error
+
+    Search "error" emits in the last ``n_log`` lines.
+    This function presently recognizes the following errors:
+
+    - MemoryError
+    """
+    logs = client.scripts.list_containers_runs_logs(job_id, run_id, limit=nlog)
+    msgs = [l['message'] for l in logs if l['level'] == 'error']
+
+    # Check for memory errors
+    mem_err = [m for m in msgs if m.startswith('Process ran out of its')]
+    if mem_err:
+        exc = MemoryError(mem_err[0])
+    else:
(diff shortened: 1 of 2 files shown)


civis/ml/_model.py
```diff
@@ -151,6 +151,26 @@ def _load_estimator(job_id, run_id, filename='estimator.pkl', client=None):
...
+    Search "error" emits in the last ``n_log`` lines.
+    This function presently recognizes the following errors:
+
+    - MemoryError
+    """
+    logs = client.scripts.list_containers_runs_logs(job_id, run_id, limit=nlog)
+    msgs = [l['message'] for l in logs if l['level'] == 'error']
+
+    # Check for memory errors
+    mem_err = [m for m in msgs if m.startswith('Process ran out of its')]
```
keithing: I don't think we should be scraping logs for errors.  Can we create a more general error (`CivisModelTrainError`?) and show the last n lines of the log as the message instead?

stephen-hoover:
@keithing , the reason I thought it was worth scraping the logs here is that running out of memory is something which might happen with some regularity -- if datasets are too large, then the training job would fail. It's impossible to catch that on the Platform side, but the log error messages seem to be clear and well-defined. It's better for the users to get a clear report of what happened and why.

Furthermore, if we raise a `MemoryError` here, you can imagine users catching that and retrying the job with an increased memory budget. This is something which could be automatically fixed if you know what's going on.

It might be worthwhile to include log lines in a generic exception -- perhaps by using them as the message for a `CivisJobFailure` -- but I'd still like to special-case the out-of-memory errors.

stephen-hoover:
I added functionality to return log lines in an error message if we don't recognize the error.

PR #86, opened 2017-05-15
[CIVP-11393] BUG Only set exceptions on ModelFuture if there was an error

stephen-hoover:
In #79, the entire `if` gate in `ModelPipeline._set_model_exception` was removed, when I should only have removed the part after the `and`. If we don't have the `if an exception happened in the job` check, we'll set an exception on all `ModelFuture` objects, even if everything succeeded.


keithing:
LGTM!

PR #88, opened 2017-05-15
TST Fix test which relied on dictionary order

stephen-hoover:
Dictionaries are not guaranteed to have any order, and in particular, the order may change between Python versions. Test both possible orderings.

### DIFF
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -569,7 +569,7 @@ def test_modelpipeline_classmethod_constructor_nonint_id():
 
     out = json.dumps({'job': mp.train_result_.job_id,
                       'run': mp.train_result_.run_id})
-    assert out == '{"job": 1, "run": 2}'
+    assert out == '{"job": 1, "run": 2}' or out == '{"run": 2, "job": 1}'
 
 
 @mock.patch.object(_model, 'ModelFuture', autospec=True)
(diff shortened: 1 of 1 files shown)


civis/ml/tests/test_model.py
```diff
@@ -569,7 +569,7 @@ def test_modelpipeline_classmethod_constructor_nonint_id():
 
     out = json.dumps({'job': mp.train_result_.job_id,
                       'run': mp.train_result_.run_id})
-    assert out == '{"job": 1, "run": 2}'
+    assert out == '{"job": 1, "run": 2}' or out == '{"run": 2, "job": 1}'
```
keithing: Should we be testing `mp.train_result_.job_id == 1` and `mp.train_result_.run_id == 2`?

stephen-hoover: I added a comment to this test to explain what's going on. The original error was from `json.dumps` -- we have to serialize the job and run ID to pass them as arguments to the custom script. But if they're an `np.int64` (which you get if you, for example, read them from a CSV with `pandas.read_csv`), then `json.dumps` gets confused and chokes. This test is verifying that we can create the JSON blob.

keithing: Ah, okay.  +1 for avoid the dreaded serialization errors.

PR #89, opened 2017-05-16
[CIVP-11405] BLD: Fix functools32 installation

jseabold:
Hopefully fixes the following error on Python 3, but I haven't been able to replicate not installing from PyPI.

```
Using cached functools32-3.2.3-2.zip
Complete output from command python setup.py egg_info:
This backport is for Python 2.7 only.
```

Issue is described in the following

https://github.com/inveniosoftware/invenio-oaiserver/issues/112
https://github.com/Julian/jsonschema/issues/228
https://hynek.me/articles/conditional-python-dependencies/
https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies

And the differing usage of install_requires vs requirements.txt here

https://packaging.python.org/requirements/

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 pyyaml>=3.0,<=3.99
 click>=6.0,<=6.99
 funcsigs==1.0.2 ; python_version == '2.7'
-future>-0.16,<=0.99 ; python_version == '2.7'
+future>=0.16,<=0.99 ; python_version == '2.7'
 futures==3.1.1 ; python_version == '2.7'
 jsonref>=0.1.0,<=0.1.99
 requests>=2.12.0,==2.*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+- Fix issue with Python 2/3 dependency management (#89).
+
 ## 1.5.1 - 2017-05-15
 ### Fixed
 - Fixed a bug which caused an exception to be set on all ``ModelFuture`` objects, regardless of job status (#86).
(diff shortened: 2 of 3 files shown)


keithing:
@jseabold we need to make a patch release for this bugfix.  Do you mind merging against `release-1.5` instead of `master`?

keithing:
Do you mind also adding a line to the changelog?

keithing:
@jseabold, actually hold on for a second before changing the merge branch.  We're settling on the naming convention for release branches.

keithing:
Okay, go ahead and switch the merge into branch to `release-1.5`.  (We originally were using `release-v1.5`, but are going to move ahead without the leading v for versions).

peterskipper:
Has this been merged? I'm running into this problem trying to reconstruct an environment with conda

requirements.txt
```diff
@@ -1,7 +1,7 @@
 pyyaml>=3.0,<=3.99
 click>=6.0,<=6.99
 funcsigs==1.0.2 ; python_version == '2.7'
-future>-0.16,<=0.99 ; python_version == '2.7'
+future>=0.16,<=0.99 ; python_version == '2.7'
```
keithing: If we're no longer basing our requirements off `requirements.txt`, it might be a good idea to remove it (to avoid confusion).

stephen-hoover: Or maybe parse the requirements file to fill the `install_requires` and `extras_require` fields.

jseabold: I specifically did not want to parse here. They have different uses as linked in the PR. If someone wants to pin requirements.txt to versions for its intended use. that's fine, but probably a different PR.

keithing: > They have different uses as linked in the PR. If someone wants to pin requirements.txt to versions for its intended use. that's fine, but probably a different PR

Yeah, let's not get into pinning versions here.

PR #90, opened 2017-05-16
[CIVP-11423] BUG get validation metadata from training job for prediction future

jmemich:
This was the only `@property` that wasn't tying the file ID lookups to the training job.

EDIT: now pointing at `release-1.5`.

### DIFF
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -465,8 +465,9 @@ def test_estimator(mock_le):
         "The Estimator is only downloaded once and cached."
 
 
-@mock.patch.object(_model.cio, "file_id_from_run_output")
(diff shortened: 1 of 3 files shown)


civis/ml/tests/test_model.py
```diff
@@ -465,26 +465,60 @@ def test_estimator(mock_le):
         "The Estimator is only downloaded once and cached."
 
 
-@mock.patch.object(_model.cio, "file_id_from_run_output",
-                   mock.Mock(return_value=11, spec_set=True))
+@mock.patch.object(_model.cio, "file_id_from_run_output")
```
stephen-hoover: Could you add `autospec=True` here, and in the other patch decorators you've changed?

jmemich:
Was I meant to update the data on the `CHANGELOG`?

PR #91, opened 2017-05-16
[CIVP-11427] BUG Improve thread safety in PollableResult

stephen-hoover:
We have at least once instance of a bug which looks to be caused by non-threadsafe code. The most likely culprit is in the `PollableResult` stack. Add some extra locks to avoid possibly double-destroying anything. `cleanup` looks like it was always called from inside a function holding the lock, so it's probably fine. It doesn't hurt to add an extra lock. `_reset_polling_thread` wasn't behind a lock, and it was being used as a callback from `pubnub` -- perhaps this could have been a problem?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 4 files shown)


stephen-hoover:
@menglewis , do you have time to review this in the next day or so?
@keithing for viz.

stephen-hoover:
Closed in favor of #92 .

PR #83, opened 2017-05-12
[CIVP-11087] Convert README.md to README.rst

jacksonllee:
`setup.py` points `long_description` to the readme file, but the [PyPI site for the Civis client](https://pypi.python.org/pypi/civis) doesn't render Markdown. This PR converts the readme from Markdown to reStructuredText.

PR #95, opened 2017-05-18
[CIVP-11087] Fix readme filename extension in MANIFEST.in

jacksonllee:
I missed this when converting the readme from Markdown to reStructuredText at #83. Probably no bugfix release is needed.


keithing:
Thanks for catching this.  LGTM!

PR #99, opened 2017-06-05
[PARO-307] ModelPipeline email notifications

jmemich:
Example email: image

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -560,6 +560,18 @@ class ModelPipeline:
         Memory requested from Civis Platform for training jobs, in MiB
     disk_requested : float, optional
         Disk space requested on Civis Platform for training jobs, in GB
+    notification_urls : list
+        URLs to receive a POST request at job completion
+    success_email_subject : string
+        Custom subject line for success e-mail.
+    success_email_body : string
+        Custom body text for success e-mail, written in Markdown.
+    success_email_addresses : list
+        Addresses to notify by e-mail when the job completes successfully.
+    failure_email_addresses : list
+        Addresses to notify by e-mail when the job fails.
+    stall_warning_minutes : integer
+        Stall warning emails will be sent after this amount of minutes.
     verbose : bool, optional
         If True, supply debug outputs in Platform logs and make
         prediction child jobs visible.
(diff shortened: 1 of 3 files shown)


civis/ml/_model.py
```diff
@@ -631,7 +643,10 @@ def __init__(self, model, dependent_variable,
                  cross_validation_parameters=None, model_name=None,
                  calibration=None, excluded_columns=None, client=None,
                  cpu_requested=None, memory_requested=None,
-                 disk_requested=None, verbose=False):
+                 disk_requested=None, notification_urls=None,
+                 success_email_subject=None, success_email_body=None,
+                 success_email_addresses=None, failure_email_addresses=None,
+                 stall_warning_minutes=None, verbose=False):
```
keithing: What do you think about having a single `notifications` parameter that takes a dict with all the other params?  Then documentation can say something like "notifications to pass to APIClient.scripts.post_custom" ?

jmemich: I thought about this but we don't do it for required resources and additionally I worry the user will just have to look up the docs without this. Happy to have to change if you'd rather but this seems like it would be both easier to use and is currently in line with convention.

keithing: I slightly prefer just passing a notification dict so we're not maintain docs in two different places.  We'll also get around needing to default to "success_on" and "failure_on", which might be unexpected to the user.  If we were ever to expose `schedule`, too, I think we couldn't expose all the parameters.

stephen-hoover: That almost sounds like an argument to refactor the resource requirement arguments into a single parameter. Pluses of having these be separate parameters is that it's easier to find them with tab completion. Minuses is that this is six extra (rarely used?) parameters in an `init` which is already probably too full of parameters.

jmemich: OK, I'm happy making the change to `notifications` here. Should I also make the change to `resources` as well or open a new PR and do it there?

stephen-hoover: `resources` would be a new PR. That's an API-breaking change, though, and I'm not sure it's worth going through a deprecation cycle just for that.

jmemich: Yeah, I was similarly concerned. I'll file a ticket for it and only address the `notifications` change.

civis/ml/_model.py
```diff
@@ -560,6 +561,9 @@ class ModelPipeline:
         Memory requested from Civis Platform for training jobs, in MiB
     disk_requested : float, optional
         Disk space requested on Civis Platform for training jobs, in GB
+    notifications : dict
+        See :func:`~civis.resources._resources.Scripts.post_custom` for
```
jmemich: @keithing I cribbed this from somewhere else -- does this look right to you? (I'm not as familiar with sphinx documentation as I should be)

keithing: I just built the docs locally and it works for me.

PR #101, opened 2017-06-08
Ensuring null values are preserved in data transformation.

mlnichols:
Null values should be retained in the data transformation (although irrelevant here, in other datasets this can be important and will allow people to more readily follow the example).

The parenthesis around the second condition are unnecessary but IMO provide clarity.

### DIFF
--- a/examples/CivisML_parallel_training.ipynb
+++ b/examples/CivisML_parallel_training.ipynb
@@ -128,7 +128,7 @@
    "metadata": {},
    "source": [
     "## Munging Data\n",
-    "- We'll add in the column names, and tell pandas that null values are encoded as a "?".\n",
+    "- We'll add in the column names, and tell pandas that null values are encoded as a '?'.\n",
     "- We can peek at shape. This is a small table, which I chose to allow for quick training.\n",
     "- We'll recode the response 4 (cancer) and 2 (benign) as 1 and 0 respectively (keeping care to ensure null values are retained). **I don't have to do this**, but to me it feels more intuitive to deal with ones and zeros"
    ]
--- a/examples/CivisML_parallel_training.ipynb
+++ b/examples/CivisML_parallel_training.ipynb
@@ -128,7 +128,7 @@
    "metadata": {},
    "source": [
     "## Munging Data\n",
-    "- We'll add in the column names, and tell pandas that null values are encoded as a "np.nan".\n",
+    "- We'll add in the column names, and tell pandas that null values are encoded as a "?".\n",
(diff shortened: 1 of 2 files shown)


examples/CivisML_parallel_training.ipynb
```diff
@@ -153,7 +153,7 @@
     "\n",
     "print(\"Shape: \", df.shape)\n",
     "\n",
-    "df['is_cancer'] = [1 if val == 4 else 0 for val in df['is_cancer'] ]\n",
+    "df['is_cancer'] = [1 if val == 4 else (0 if val == 2 else '?') for val in df['is_cancer'] ]\n",
```
stephen-hoover: Thanks for the PR! Perhaps `np.nan` instead of `'?'`? That's a more usual way to represent nulls in modeling.

mlnichols: I originally left it off to keep it closer to what the author used. Though I agree np.nan is better. I added it (along with changelog modification). For clarity I added an import to numpy rather than use it via pandas since pd.np.nan can be confusing.

examples/CivisML_parallel_training.ipynb
```diff
@@ -149,11 +150,11 @@
     "COL_NAMES = ['sample_id', 'clump_thickness', 'uniformity_cell_size', 'uniformity_cell_shape',\n",
     "            'marginal_adhesion', 'epithelial_cell_size', 'bare_nuclei', 'bland_chromatin',\n",
     "            'normal_nucleoli', 'mitoses', 'is_cancer']\n",
-    "df = pd.read_csv(\"breast_cancer_data.csv\", names=COL_NAMES, na_values='?')\n",
+    "df = pd.read_csv(\"breast_cancer_data.csv\", names=COL_NAMES, na_values=np.nan)\n",
```
stephen-hoover: This line should stay the same. The `na_values` argument tells `pandas` which strings represent missing data. All of them get encoded as `np.nan`. We still want any "?" strings present in the CSV to be treated as missing data.

https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html#pandas-read-csv

mlnichols: Yep. Sorry, forgot it's for the CSV encoding rather than how you want pandas to encode it. Fixed now.

stephen-hoover:
Thank you for the contribution, @mlnichols !

peterskipper:
Late to this, but thanks @mlnichols for making it better!

PR #102, opened 2017-06-22
[CIVP-11787] ENH ContainerFuture class to track container scripts

stephen-hoover:
Add a `ContainerFuture` object which knows how to query container scripts, how to cancel container scripts, and how to retry container scripts. The `civis.ml.ModelFuture` can be a subclass of the `ContainerFuture`; models are a specialized container script.

This functionality (especially retries) adds more tooling for running concurrent jobs on the Civis Platform, and is a setup for executor classes coming in the next PR.

### DIFF
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import
 
 from builtins import super
+import logging
+import time
 
 from civis import APIClient
 from civis.base import DONE
-from civis.polling import PollableResult
+from civis.polling import PollableResult, _ResultPollingThread
 
 try:
     from pubnub.pubnub import PubNub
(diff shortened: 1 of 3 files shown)


keithing:
This is cool.  I like the abstraction `CivisFuture -> ContainerFuture -> ModelFuture`.  There is a lot more code that is going to move here, is the idea to keep `ContainerFuture` in `futures.py` and then add a module (like `parallel`) for the rest of that code?

stephen-hoover:
I was planning to put the executor code in `civis.futures`; that seems the most appropriate namespace for it. I wasn't sure what the best place would be for the joblib-backend code. Suggestions welcome.

keithing:
> I was planning to put the executor code in `civis.futures`

That is a good plan.  I'm worried that it might get a little unwieldily to put everything in single file; we could always create a `futures` module with files beneath it and import what we need with `futures.__init__`.  But let's hold off on that for now.

PR #103, opened 2017-06-23
[PARO-356] ENH CivisML 1.1

jmemich:
@keithing I think this gets merged into master -- right?

You can find a successful job [here](https://platform.civisanalytics.com/#/scripts/custom/6206831). Feedback welcome.


docs/source/ml.rst
```diff
@@ -77,6 +77,22 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
 you can add one by wrapping it in a :class:`~sklearn.calibration.CalibratedClassifierCV`.
 
 
+Custom Dependencies
+-------------------
+
+Installing packages to from PyPI is straightforward. You can specify a `dependencies`
```
keithing: `to from` typo

civis/ml/_model.py
```diff
@@ -564,6 +576,15 @@ class ModelPipeline:
     notifications : dict
         See :func:`~civis.resources._resources.Scripts.post_custom` for
         further documentation about email and URL notification.
+    dependencies : array, optional
+        List of packages to install from PyPI or Github. If a private Github
+        pacakge is specificed, please include a ``github_token_name``
```
keithing: `pacakge` typo

docs/source/ml.rst
```diff
@@ -77,6 +77,22 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+argument to `~civis.ml.ModelPipeline` which will install the dependencies in your runtime
+environment. Installing a package from github only requires passing the HTTPS URL in
+the form of, for example, `git+https://github.com/scikit-learn/scikit-learn`.
+
+Additionally, you can store your Github API Token in platform as a credential to use for
+installing private Github repositores. Simply go to Github at the
+`https://github.com/settings/tokens` URL, copy your token into the password field of a
+credential, and pass the credential name to the `github_token_name` argument in
+`~civis.ml.ModelPipeline`. Note, installing private dependencies (i.e., submodules) as
+part of your package is not yet supported.
```
keithing: What other kinds of private dependencies could this mean besides submodules?

stephen-hoover:
Note that this PR needs to include an update of the template IDs so that model training happens with v1.1 templates.

docs/source/ml.rst
jmemich: From Github, nothing that I know of. I can change this if you'd rather -- as it stands it's unclear.

civis/ml/_model.py
```diff
@@ -527,6 +539,15 @@ class ModelPipeline:
...
         further documentation about email and URL notification.
+    dependencies : array, optional
+        List of packages to install from PyPI or Github. If a private Github
+        package is specificed, please include a ``github_token_name``
+        argument as well (see below).
+    github_token_name : str, optional
+        Name of Github API token stored in platform as the password field in a
+        custom platform credential. Used only when installing private Github
+        repositories. Note, these API tokens can be generated from this URL:
+        https://github.com/settings/tokens.
```
stephen-hoover: This could potentially be a token from another service as well, couldn't it? E.g. https://blog.bitbucket.org/2016/06/06/app-passwords-bitbucket-cloud/ ?

civis/ml/_model.py
```diff
@@ -527,6 +539,15 @@ class ModelPipeline:
     notifications : dict
         See :func:`~civis.resources._resources.Scripts.post_custom` for
         further documentation about email and URL notification.
+    dependencies : array, optional
+        List of packages to install from PyPI or Github. If a private Github
```
stephen-hoover: This should be more general than just GitHub, e.g. https://stackoverflow.com/questions/38335377/install-python-package-from-private-bitbucket-repo .

docs/source/ml.rst
```diff
@@ -77,6 +77,21 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
 you can add one by wrapping it in a :class:`~sklearn.calibration.CalibratedClassifierCV`.
 
 
+Custom Dependencies
+-------------------
```
stephen-hoover: Perhaps this should link to the pip documentation somewhere, e.g. https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support . This section of the documentation should probably mention that what we're running is `pip install [your input here]`.

docs/source/ml.rst
```diff
@@ -77,6 +77,21 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+
+Installing packages from PyPI is straightforward. You can specify a `dependencies`
+argument to `~civis.ml.ModelPipeline` which will install the dependencies in your runtime
```
stephen-hoover: I'd give an explicit example to show what this looks like. And you should strongly encourage users to peg versions of dependencies. If they don't, then there's a serious risk of scoring jobs failing. It's probably worth mentioning both here and in the doc string.

civis/ml/_model.py
jmemich: I'm going to rename this to `git_token_name` so we differentiate git hosting services from github specifically.

docs/source/ml.rst
```diff
@@ -77,6 +77,34 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+  from pyearth import Earth
+  deps = ['git+https://github.com/scikit-learn-contrib/py-earth.git']
+  est = Earth()
+  model = ModelPipeline(est, dependent_variable='age')
+  train = model.train(table_name='donors.from_march', database_name='client')
+
+Additionally, you can store a remote git hosts' API token in platform as a
+credential to use for installing private git repositores. For example, you can go to
+Github at the `https://github.com/settings/tokens` URL, copy your token into the
+password field of a credential, and pass the credential name to the `git_token_name`
```
keithing: A simple example of how to do this through the API (like below) would be nice.

```python
import civis
password = [token_copied from https://github.com/settings/tokens]
username = [my_github_username]
git_token_name = [name for credential (i.e. "github_credentials")]

client = civis.APIClient()
credential = client.credentials.post(password=password, username=username, name=git_token_name, type="Github")

pipeline = civis.ml.ModelPipeline(..., git_token_name = git_token_name)
```

civis/ml/_model.py
```diff
@@ -701,6 +724,10 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
             name = name[:-len(' Train')]
         notifications = {camel_to_snake(key): val for key, val
                          in container.notifications.items()}
+        dependencies = args.get('DEPENDENCIES', None)
+        if dependencies:
+            dependencies = dependencies.split()
+        git_token_name = args.get('GITHUB')
```
stephen-hoover: The parameter name should probably change, no? Maybe "GIT_CRED"?

civis/ml/_model.py
```diff
@@ -192,6 +192,18 @@ def _exception_from_logs(exc, job_id, run_id, client, nlog=15):
     return exc
 
 
+def _get_credential(credential, client=None):
+    """Get the credential ID from either an integer or a string."""
```
stephen-hoover: This will fail if the credential name is a string which can be cast to an integer. In addition, the documentation says that the input should be a name, rather than an ID. Could you remove this function and assume that the user inputs a credential name?

docs/source/ml.rst
```diff
@@ -77,6 +77,34 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+environment. VCS support is also enabled (see [docs](https://pip.pypa.io/en/stable/reference/pip_install/#vcs-support). Installing a remote git repository from, say, Github
+only requires passing the HTTPS URL in the form of, for example,
+`git+https://github.com/scikit-learn/scikit-learn`.
+
+CivisML will run `pip install [your package here]`. We strongly encourage you to pin
+package versions for consistency. Example code looks like:::
+
+  from civis.ml import ModelPipeline
+  from pyearth import Earth
+  deps = ['git+https://github.com/scikit-learn-contrib/py-earth.git']
```
stephen-hoover: Hah. And this repo has no releases yet. We could still pull from a commit. How about `deps = ['git+https://github.com/scikit-learn-contrib/py-earth.git@da856e1']`? We should demonstrate best practices in the documentation.

docs/source/ml.rst
```diff
@@ -77,6 +77,34 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+  est = Earth()
+  model = ModelPipeline(est, dependent_variable='age')
+  train = model.train(table_name='donors.from_march', database_name='client')
+
+Additionally, you can store a remote git hosts' API token in platform as a
```
stephen-hoover: "hosts'" -> "host's" and "platform" -> "the Civis Platform"

docs/source/ml.rst
```diff
@@ -77,6 +77,34 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+credential to use for installing private git repositores. For example, you can go to
+Github at the `https://github.com/settings/tokens` URL, copy your token into the
+password field of a credential, and pass the credential name to the `git_token_name`
+argument in `~civis.ml.ModelPipeline`. This also works with other hosting services.
+Note, installing private dependencies with submodules is not yet supported.
```
stephen-hoover: Strike "yet"?

docs/source/ml.rst
```diff
@@ -77,6 +77,34 @@ method. If the class you're using doesn't have a ``predict_proba`` method,
...
+`git+https://github.com/scikit-learn/scikit-learn`.
+
+CivisML will run `pip install [your package here]`. We strongly encourage you to pin
+package versions for consistency. Example code looks like:::
+
+  from civis.ml import ModelPipeline
+  from pyearth import Earth
+  deps = ['git+https://github.com/scikit-learn-contrib/py-earth.git']
+  est = Earth()
+  model = ModelPipeline(est, dependent_variable='age')
```
stephen-hoover: This needs a `dependencies=deps` parameter.

civis/ml/_model.py
jmemich: The API will give you back the ID for the credential rather than the name:
```python
In [8]: container = client.scripts.get_containers(6229384)

In [9]: container.arguments
Out[9]:
{'CALIBRATION': None,
 'CIVIS_FILE_ID': 5092501,
 'CVPARAMS': '{}',
 'DB': {},
 'DEBUG': True,
 'EXCLUDE_COLS': 'tsmart_county_fips tsmart_census_block_fips',
 'GITHUB': 5991,
 'IF_EXISTS': 'fail',
 'MODEL': '5092505',
 'OOSDB': {},
 'PARAMS': '{}',
 'PRIMARY_KEY': 'voterbase_id',
 'TABLE_NAME': None,
 'TARGET_COLUMN': 'registered_dem'}
```
I had included this originally so that I could handle the result of this call in `from_existing`. I'll remove this and have `from_existing` convert the ID returned to a string before instantiating the class.

civis/ml/_model.py
jmemich: I'll change this (and some of the documentation) upstream.

civis/ml/_model.py
jmemich: Oh -- now I remember why I did this. When you post a credential parameter via the API you have to use the ID, not the name.

I'll assume that by the time I'm making the `_create_custom_run` call, I'll have the name of the credential irrespective of how this was instantiated and convert it to an integer.

jmemich:
This needs to wait on a few upstream changes & some testing before we cut new templates.

civis/ml/_model.py
```diff
@@ -727,7 +715,7 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
         dependencies = args.get('DEPENDENCIES', None)
         if dependencies:
             dependencies = dependencies.split()
-        git_token_name = args.get('GITHUB')
+        git_token_name = client.credentials.get(args.get('GIT_CRED')).name
```
stephen-hoover: This errors if the target script either doesn't have a "GIT_CRED" or doesn't have the argument filled.

civis/ml/_model.py
```diff
@@ -910,8 +898,9 @@ def _create_custom_run(self, template_id, job_name=None, table_name=None,
                 # modify via arguments if users give a non-default value.
                 script_arguments[key] = value
         if self.git_token_name:
-            script_arguments['GITHUB'] = _get_credential(
-                self.git_token_name)
+            cred = find_one(self._client.credentials.list(),
+                            name=self.git_token_name)
```
stephen-hoover: I'd add the `type='Custom'` parameter to the `.list()` call. That will reduce the number of credentials returned, and reduce the chance of name collisions. I think it's also worth verifying that the name is unique and giving an error if not. There could easily be more than one custom credential with the same name.

stephen-hoover:
Thanks! I'm happy with the current state of this PR. I'll approve once it includes v1.1 template IDs and we've released and vetted the v1.1 Docker image.

stephen-hoover:
@jmemich , the v1.1 templates are 9112 for training and 9113 for prediction. Please also add an entry to the changelog.

jmemich:
Acceptance tests pass. We're good to merge.

civis/ml/_model.py
```diff
@@ -37,7 +37,8 @@
 
 # Map training template to prediction template so that we
 # always use a compatible version for predictions.
-_PRED_TEMPLATES = {8387: 8388,  # v1.0
+_PRED_TEMPLATES = {9112: 9113,  # v1.1
+                   8387: 8388,  # v1.0
```
stephen-hoover: Please set this to predict with template 9113; we should use the most up-to-date template possible for the predictions.

civis/_version.py
```diff
@@ -1 +1 @@
-__version__ = "1.5.2"
+__version__ = "1.6.0"
```
stephen-hoover: I don't think this is the right PR for this change.

stephen-hoover:
@jmemich , tests are failing here.

PR #104, opened 2017-06-26
[CIVP-11787] ENH futures.Executor classes for Civis

stephen-hoover:
Add two Civis-specific Executor classes (subclasses of the standard library `concurrent.futures.Executor`). The `_ContainerShellExecutor` uses Civis Platform container jobs to parallelize over either shell commands or container script arguments. The `CustomScriptExecutor` parallelizes over arguments to a custom script.

### DIFF
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -9,6 +9,8 @@ import logging
 import time
 import threading
 
+import six
+
 from civis import APIClient
 from civis.base import DONE
 from civis.polling import PollableResult, _ResultPollingThread
@@ -307,7 +309,8 @@ def create_docker_command(*args, **kwargs):
                      for k, v in sorted(kwargs.items())])
 
 
-class CivisExecutor(Executor, metaclass=ABCMeta):
+@six.add_metaclass(ABCMeta)
+class _CivisExecutor(Executor):
     def __init__(self,
                  script_name=None,
                  hidden=True,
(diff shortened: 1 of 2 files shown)


stephen-hoover:
@keithing , this is a lot of lines, but I think only the second commit needs a line-by-line scrutiny. Also, I plan to add documentation on the new features to the markdown pages in a future PR.

civis/futures.py
```diff
@@ -285,3 +292,386 @@ def cancel(self):
                 self._invoke_callbacks()
                 return self.cancelled()
             return False
+
+
+def create_docker_command(*args, **kwargs):
```
keithing: Do we want this to be private as well?

stephen-hoover: You're right, it should be. Fixed.

civis/futures.py
```diff
@@ -285,3 +292,386 @@ def cancel(self):
...
+            hidden=self.hidden,
+            params=self.params,
+            arguments=combined_args,
+            git_credential_id=self.git_credential_id,
+        )
+
+        return job
+
+
+class CustomPoolExecutor(_CivisExecutor):
```
keithing: `CustomScriptPoolExecutor`?

civis/futures.py
```diff
@@ -285,3 +292,386 @@ def cancel(self):
...
+                 max_n_retries=0,
+                 client=None,
+                 polling_interval=None,
+                 inc_script_names=False):
+        self.from_template_id = from_template_id
+        self.arguments = arguments
+
+        if script_name is None:
+            date_str = datetime.datetime.today().strftime("%Y-%m-%d")
+            script_name = "CustomPoolExecutorScript {}".format(date_str)
```
keithing: If we change the name, we need to update that here too.

keithing:
Everything looks good, after we decide what the correct name is for `CustomPoolExecutor` this looks ready to go.

civis/futures.py
stephen-hoover: How about `CustomScriptExecutor`? It's shorter, and there's not really a "pool".

keithing: Great, I like that.

stephen-hoover:
I renamed `CustomPoolExecutor` to `CustomScriptExecutor`. I also changed `_ContainerPoolShellExecutor` to `_ContainerShellExecutor`; it's shorter and probably more accurate.

civis/futures.py
```diff
@@ -285,3 +292,386 @@ def cancel(self):
...
+        self.params = params
+        self.arguments = arguments
+
+        if required_resources is None:
+            required_resources = {'cpu': 1024, 'memory': 1024}
+        self.required_resources = required_resources
+
+        if script_name is None:
+            date_str = datetime.datetime.today().strftime("%Y-%m-%d")
+            script_name = ("ContainerPoolShellExecutorScript "
```
keithing: This needs to be renamed too.

stephen-hoover: Oops. Fixed.

PR #106, opened 2017-06-29
[CIVP-11787] Add joblib backend

stephen-hoover:
Add a new `civis.parallel` module which contains tools to register a custom backend for the `joblib` library. This will let users leverage the Civis Platform infrastructure for parallel computing, and makes it easy to distribute parallel scikit-learn computations over the Civis infrastructure.

Most of this code was developed in a separate repo and has been in Civis internal use. This PR copies over that code, adapts it to work in the civis-python repo, and makes small modifications for Python 2 compatibility.

### DIFF
--- a/civis/parallel.py
+++ b/civis/parallel.py
@@ -1,0 +1,732 @@
+"""
+A module that facilitates the use of Civis's container scripts for
+parallelizing Python workflows.
+"""
+from concurrent.futures import wait
+from datetime import datetime, timedelta
+from io import BytesIO
+import logging
+import os
+from tempfile import TemporaryDirectory
+import time
+
+import joblib
+from joblib._parallel_backends import ParallelBackendBase
+from joblib.my_exceptions import TransportableException
+import civis
+from civis.base import CivisAPIError
+import requests
+
+from civisjobs.containers import ContainerPoolExecutor, CustomPoolExecutor
+
+_LOGGER = logging.getLogger(__name__)
+_THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+_DEFAULT_SETUP_CMD = ":"  # An sh command that does nothing.
+_DEFAULT_REPO_SETUP_CMD = "cd /app; python setup.py install; cd /"
+_ALL_JOBS = 50  # Give the user this many jobs if they request "all of them"
+
+
+def infer_backend_factory(required_resources=None,
+                          params=None,
+                          arguments=None,
+                          client=None,
+                          polling_interval=None,
+                          setup_cmd=None,
+                          max_submit_retries=0,
+                          max_job_retries=0,
+                          hidden=True):
+    """Infer the container environment and return a backend factory.
+
+    This function helps you run additional jobs from code which executes
+    inside a Civis container job. The function reads settings for
+    relevant parameters (e.g. the Docker image) of the container
+    it's running inside of.
+
+    .. note:: This function will read the state of the parent
+    container job at the time this function executes. If the
+    user has modified the container job since the run started
+    (e.g. by changing the GitHub branch in the container's GUI),
+    this function may infer incorrect settings for the child jobs.
+
+    Parameters
+    ----------
+    required_resources : dict or None, optional
+        The resources needed by the container. See the
+        `container scripts API documentation
+        <https://platform.civisanalytics.com/api#resources-scripts>`
+        for details. Resource requirements not specified will
+        default to the requirements of the current job.
+    params : list or None, optional
+        A definition of the parameters this script accepts in the
+        arguments field. See the `container scripts API documentation
+        <https://platform.civisanalytics.com/api#resources-scripts>`
+        for details.
+            Parameters of the child jobs will default to the parameters
+        of the current job. Any parameters provided here will override
+        parameters of the same name from the current job.
+    arguments : dict or None, optional
+        Dictionary of name/value pairs to use to run this script.
+        Only settable if this script has defined params. See the `container
+        scripts API documentation
+        <https://platform.civisanalytics.com/api#resources-scripts>`
+        for details.
+            Arguments will default to the arguments of the current job.
+        Anything provided here will override portions of the current job's
+        arguments.
+    client : `civis.APIClient` instance or None, optional
+        An API Client object to use.
+    polling_interval : int, optional
+        The polling interval, in seconds, for checking container script status.
+        If you have many jobs, you may want to set this higher (e.g., 300) to
+        avoid `rate-limiting <https://platform.civisanalytics.com/api#basics>`.
+        You should only set this if you aren't using ``pubnub`` notifications.
+    setup_cmd : str, optional
+        A shell command or sequence of commands for setting up the environment.
+        These will precede the commands used to run functions in joblib.
+        This is primarily for installing dependencies that are not available
+        in the dockerhub repo (e.g., "cd /app && python setup.py install"
+        or "pip install gensim").
+            With no GitHub repo input, the setup command will
+        default to a command that does nothing. If a `repo_http_uri`
+        is provided, the default setup command will attempt to run
+        "python setup.py install". If this command fails, execution
+        will still continue.
+    max_submit_retries : int, optional
+        The maximum number of retries for submitting each job. This is to help
+        avoid a large set of jobs failing because of a single 5xx error. A
+        value higher than zero should only be used for jobs that are idempotent
+        (i.e., jobs whose result and side effects are the same regardless of
+        whether they are run once or many times).
+    max_job_retries : int, optional
+        Retry failed jobs this number of times before giving up.
+        Even more than with `max_submit_retries`, this should only
+        be used for jobs which are idempotent, as the job may have
+        caused side effects (if any) before failing.
+        These retries assist with jobs which may have failed because
+        of network or worker failures.
+    hidden: bool, optional
+        The hidden status of the object. Setting this to true
+        hides it from most API endpoints. The object can still
+        be queried directly by ID. Defaults to True.
+
+    Raises
+    ------
+    RuntimeError
+        If this function is not running inside a Civis container job.
+
+    See Also
+    --------
+    make_backend_factory
+    """
+    if client is None:
+        client = civis.APIClient(resources='all')
+
+    if not os.environ.get('CIVIS_JOB_ID'):
+        raise RuntimeError('This function must be run '
+                           'inside a container job.')
+    state = client.scripts.get_containers(os.environ['CIVIS_JOB_ID'])
+    if state.from_template_id:
+        # If this is a Custom Script from a template, we need the
+        # backing script. Make sure to save the arguments from
+        # the Custom Script: those are the only user-settable parts.
+        template = client.templates.get_scripts(state.from_template_id)
+        try:
+            custom_args = state.arguments
+            state = client.scripts.get_containers(template.script_id)
+            state.arguments = custom_args
+        except civis.base.CivisAPIError as err:
+            if err.status_code == 404:
+                raise RuntimeError('Unable to introspect environment from '
+                                   'your template\'s backing script. '
+                                   'You may not have permission to view '
+                                   'script ID {}.'.format(template.script_id))
+            else:
+                raise
+
+    # Default to this container's resource requests, but
+    # allow users to override it.
+    state.required_resources.update(required_resources or {})
+
+    # Update parameters with user input
+    params = params or []
+    for input_param in params:
+        for param in state.params:
+            if param['name'] == input_param['name']:
+                param.update(input_param)
+                break
+        else:
+            state.params.append(input_param)
+
+    # Update arguments with input
+    state.arguments.update(arguments or {})
+
+    return make_backend_factory(docker_image_name=state.docker_image_name,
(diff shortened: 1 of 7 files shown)


stephen-hoover:
@mheilman , the second commit is the one that needs more attention here.

civis/parallel.py
```diff
@@ -0,0 +1,719 @@
...
+            # the container, and then run it on the uploaded job.
+            # Only download the runner script if it doesn't already
+            # exist in the destination environment.
+            runner_remote_path = "joblib_remote_worker"
+            cmd = ("{setup_cmd} && "
+                   "if command -v {runner_remote_path} >/dev/null; "
+                   "then {runner_remote_path} {func_file_id}; "
+                   "else pip install joblib=={jl_version} && "
+                   "civis files download {run_file_id} {runner_remote_path} "
+                   "&& python {runner_remote_path} {func_file_id}; fi"
```
beckermr: We should probably add `exec` here like we talked about, in front of both `python` and in between `then {runner_remote_path}`.

civis/parallel.py
```diff
@@ -0,0 +1,719 @@
...
+        A definition of the parameters this script accepts in the
+        arguments field. See the `container scripts API documentation
+        <https://platform.civisanalytics.com/api#resources-scripts>`
+        for details.
+    arguments : dict or None, optional
+        Dictionary of name/value pairs to use to run this script.
+        Only settable if this script has defined params. See the `container
+        scripts API documentation
+        <https://platform.civisanalytics.com/api#resources-scripts>`
+        for details.
```
beckermr: We are missing the `cancel_timeout` parameter here.

beckermr: Ditto above.

dev-requirements.txt
```diff
@@ -5,3 +5,4 @@ mock==2.0.0 ; python_version == '2.7'
 pytest-cov>=2.4.0,==2.*
 vcrpy>=1.10.0,<=1.10.9
 vcrpy-unittest==0.1.6
+joblib~=0.11
```
beckermr: If the container has to have the civis client for the joblib backend stuff and the civis client depends on joblib, then I don't think we need to install joblib on the fly above, right?

setup.py
```diff
@@ -58,11 +58,13 @@ def main():
                 'futures==3.1.1',
                 'functools32>=3.2,<=3.99'
             ],
-            'pubnub': ['pubnub>=4.0.0,<=4.99']
+            'pubnub': ['pubnub>=4.0.0,<=4.99'],
+            'joblib': ['joblib>=0.11.0,<=0.11.99'],
```
beckermr: Or are we making joblib optional? Is there any good reason to make either of these optional? It feels like maybe we can deprecate this behaivor?

dev-requirements.txt
stephen-hoover: This is the dev requirements file. I don't want to require joblib to install the client, but we need it for testing.

civis/parallel.py
stephen-hoover: I have another commit ready for that, but I thought it would be better to do that in a different PR. The diff in this PR is large enough already.

setup.py
stephen-hoover: Deprecate which behavior? You mean make these both required? That sounds reasonable to me, since they're both pretty light. We've made an effort to keep dependencies to a minimum in the past, and I was sticking with that.

beckermr: Yeah it just gets so confusing. One list of deps for dev and running seems like a win to me.

civis/parallel.py
beckermr: Sounds good. We can add the cancel timeout parameter there too.

dev-requirements.txt
beckermr: Why not require joblib for the client install? The parallel module is useless without it. What's the point?

civis/run_joblib_func.py
```diff
@@ -0,0 +1,81 @@
...
+
+    client = civis.APIClient()
+    job_id = os.environ.get('CIVIS_JOB_ID')
+    run_id = os.environ.get('CIVIS_RUN_ID')
+    if not job_id or not run_id:
+        raise RuntimeError("This function must be run inside a "
+                           "Civis container job.")
+
+    func_buffer = BytesIO()
+    civis.io.civis_to_file(func_file_id, func_buffer)
```
beckermr: All of these API calls need retries

civis/run_joblib_func.py
```diff
@@ -0,0 +1,81 @@
...
+        if result is not None:
+            # If the function exits without erroring, we may not have a result.
+            result_buffer = BytesIO()
+            joblib.dump(result, result_buffer, compress=3)
+            result_buffer.seek(0)
+            output_name = "Results from Joblib job {} / run {}".format(job_id,
+                                                                       run_id)
+            output_file_id = civis.io.file_to_civis(result_buffer, output_name,
+                                                    expires_at=expires_at)
+            client.scripts.post_containers_runs_outputs(job_id, run_id,
```
beckermr: These two calls too.

civis/run_joblib_func.py
beckermr: Note that even for functions being executed here which are not idempotent, the downloading and uploading of result steps are idempotent.

civis/run_joblib_func.py
beckermr: Ditto on the idepotency from above.

civis/parallel.py
```diff
@@ -0,0 +1,719 @@
...
+                # Catch BaseException so we can also re-raise a
+                # KeyboardInterrupt where it can be properly handled.
+                log.debug('Exception during result download: %s', str(exc))
+                fut.remote_func_output = exc
+            else:
+                fut.result_fetched = True
+                if not fut.cancelled() and not fut.exception():
+                    # The next job will start when this callback is called.
+                    # Only run it if the job was a success.
+                    joblib_callback(fut.remote_func_output)
```
beckermr: We should support an `archive_on_success` option (maybe as opposed to `hidden`?). The idea is that all jobs would be visible in the API and the website (so `hidden=False`), but if they complete successfully, they get archived with a call to `put_containers_archive`. That call would go here.

This way people can see progress and look at running jobs, but the jobs clean themselves up unless they throw an error, in which case they are visible for inspection/debugging.

beckermr:
Also I just realized that the remote worker depends on the API client which would have the remote worker function in it. Thus if I am not mistaken, we don't need to upload the joblib worker function. I don't think backwards compatibility with containers that have older API clients is a good enough reason the keep this, but (shrug).

setup.py
mheilman: i don't have a strong opinion on this particular case because i think pubnub and joblib are pretty lightweight (and pure python?), but keeping soft dependencies doesn't seem like a bad idea.  it could be very annoying to some user if they have trouble installing our software because of a conflict having to do with functionality they don't use.

setup.py
```diff
@@ -58,11 +58,13 @@ def main():
...
         },
         entry_points={
             'console_scripts': [
-                'civis = civis.cli.__main__:main'
+                'civis = civis.cli.__main__:main',
+                'joblib_remote_worker = civis.run_joblib_func:main',
```
mheilman: `civis_joblib_remote_worker`? `civis_remote_worker`?  it might be confusing to a user to have this executable show up when autocompleting, etc.

civis/parallel.py
```diff
@@ -496,34 +503,34 @@ def _fetch_result(fut):
             as a new attribute ``remote_func_output``.
             """
             if fut.succeeded():
-                _LOGGER.debug(
+                log.debug(
                     "Ran job through Civis. script ID: %d, run ID: %d;"
```
mheilman: should the text also say "job ID" instead of "script ID"?

civis/parallel.py
mheilman: this sounds like a good idea for a future PR

civis/run_joblib_func.py
mheilman: is the default [retry behavior in the client](https://github.com/civisanalytics/civis-python/blob/master/civis/civis.py#L310-L311) not sufficient?

mheilman:
For the record, could you add a note to the PR description that most of this code was reviewed previously in PRs to an internal repo?

Also, it looks like this adds three functions to the API (`make_backend_factory`, `make_backend_template_factory`, `infer_backend_factory`), along with some miscellaneous things like exception classes, etc.  This seems good. Am I missing anything?

What's the plan for updating the documentation and changelog?

civis/run_joblib_func.py
beckermr: I was 500ed randomly recently. No idea why but I reran the job and it was fine.

setup.py
beckermr: Yeah agreed either way. I personally like a simple list of deps in one spot. Plus forcing people to have joblib simplifies the code in other places. Just something to think about...

civis/run_joblib_func.py
beckermr: We could retry 500s a few times here.

civis/parallel.py
beckermr: Works for me!

civis/run_joblib_func.py
stephen-hoover: We retry 500s from Civis, but not 500s from AWS. Those happen more frequently than I would have expected.

beckermr: Huh? The codes we retry are here

https://github.com/civisanalytics/civis-python/blob/master/civis/civis.py#L16

500 is not in them. We should retry the 500 from amazon in this case. 

So I think we need to add retries here.

stephen-hoover: Sorry, that wasn't clear. When `civis_to_file` generates a 500, it's the `requests.get` call against the pre-signed S3 URL from AWS. I don't see a way to have that call automatically retry. Worse, `civis_to_file` isn't idempotent. It modifies the state of an input buffer. If it failed halfway through, a naive retry would leave the buffer corrupted. I think the easiest thing to do is to write a new function which downloads an entire file and retries the full download on a 500. That's what https://github.com/civisanalytics/civis-python/pull/106/files#diff-a315c710eca2436eb1f8909b3633198cR412 is.

beckermr: Yep exactly. This is what I am suggesting.

civis/parallel.py
```diff
@@ -0,0 +1,719 @@
...
+                log.debug("uploaded serialized function to File: %d",
+                          func_file_id)
+
+            # Use the Civis CLI client to download the job runner script into
+            # the container, and then run it on the uploaded job.
+            # Only download the runner script if it doesn't already
+            # exist in the destination environment.
+            runner_remote_path = "joblib_remote_worker"
+            cmd = ("{setup_cmd} && "
+                   "if command -v {runner_remote_path} >/dev/null; "
```
beckermr: I think this branch is not needed any more. The joblib worker is in the client which they have to have anyways.

setup.py
mheilman: let's see what @keithing thinks when he's back tomorrow

keithing: I agree with @beckermr that we should try to use as few soft dependencies as possible to make installation more straightforward. `pandas` and `sklearn` seem to me to be good candidates for a soft dependency, while lighter packages (as @mheilman noted) like `requests-toolbelt`, `joblib`, and `pubnub` could be required.

My vote would be to leave it a soft dependency in this PR, but refactor it (along with the other listed packages) to make it required later.  We can then discuss the tradeoffs in that PR.

setup.py
stephen-hoover: Good point. Renamed to `civis_joblib_worker`.

civis/run_joblib_func.py
stephen-hoover: The `post_containers_runs_outputs` will automatically retry already. The `file_to_civis` call is in a similar situation as `civis_to_file`, and those could be addressed together in another PR.

stephen-hoover:
@mheilman , I think I've addressed all of your comments. I added a note to the changelog about this (thanks for reminding me). I plan to write documentation in another PR early next week.

@beckermr , I added another commit which skips uploading the worker function and instead pip installs civis if it's not present already. I also put on the `exec` calls. I'd planned to leave those for a follow-on PR, to keep the size of this diff down, but I suppose those changes are small. And it shouldn't break anything.

Writing new versions of the file transfer functions which can retry on AWS network errors seems like a bigger job, and not in scope for this PR.

civis/run_joblib_func.py
beckermr: Another PR is fine, but we should be making issues for these things.

beckermr:
So three new issues here:

1) add `cancel_timeout` parameter for containers
2) add retry on file i/o in remote worker func
3) change `hidden` to `archive_on_success` (or just add the later?)

beckermr:
LGTM from me (not that you all need it!)

beckermr:
One more task is to work on the deps

beckermr:
Where did we put these 4 issues?

stephen-hoover:
@beckermr , issues filed here (#108 , #109 )and on Jira.

beckermr:
Cool. We should make the archive on success issue too. I am happy to work on that one.

PR #107, opened 2017-07-13
BUG Include joblib.load in wrapper function error handling

stephen-hoover:
In the joblib backend remote runner script, the `joblib.load` call will fail if we try to load an object which refers to code that's not in our remote environment. That's an easy mistake to make, and we should make sure that users see an informative error from it. By moving the `load` call inside the try/finally block, we can record any exceptions from the `load` call and pass them back to the parent process.

### DIFF
--- a/civis/run_joblib_func.py
+++ b/civis/run_joblib_func.py
@@ -30,14 +30,14 @@ def worker_func(func_file_id):
         raise RuntimeError("This function must be run inside a "
                            "Civis container job.")
 
-    func_buffer = BytesIO()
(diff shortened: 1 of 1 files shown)


beckermr:
LGTM (if you just don't want to wait! 😄 )

stephen-hoover:
Tempting, but I think that code review is like experimentation in that you need to declare your end condition and stick with it. I'll ask you for review next time I have something like this. :-)

keithing:
LGTM!

PR #110, opened 2017-07-14
[CIVP-12032] DEP Make joblib and pubnub hard dependencies

stephen-hoover:
Both the `joblib` and `pubnub` external libraries provide valuable functionality for users of the Python API client, and they are light, fast to install packages. Make them hard dependencies. Remove code paths which allowed those libraries to be absent, and modify the documentation to reflect the fact that all users will have those two libraries installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 install:
     - pip install --upgrade pip setuptools
     - pip install -r dev-requirements.txt
-    - pip install -e .[pubnub]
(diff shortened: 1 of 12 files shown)


keithing:
This is a step in the right direction. Thanks.  LGTM!

stephen-hoover:
@keithing , I forgot the changelog entry. Just added. How does that look?

keithing:
LGTM

PR #113, opened 2017-07-18
[CIVP-12029] ENH Retry uploads and downloads in joblib backend

### LINKED ISSUE #109: Retry file I/O in joblib backend helper function
In the Civis joblib backend, the remote helper function downloads the function to run and uploads results. API calls to the Civis platform automatically retry, but the communications with S3 do not retry. We occasionally have errors when communicating with S3, so we should add retries to the file I/O.

stephen-hoover:
Large-scale jobs sometimes encounter 500 errors from web communications with S3. We can greatly reduce the chance of these failures interrupting a job by retrying ``file_to_civis`` and ``civis_to_file`` operations.

Add a wrapper for ``civis.io.file_to_civis`` which retries on network errors. Rename ``_robust_result_download`` to ``_robust_pickle_download``, since we don't care that the contents of the file are a result. Use these functions for upload and download on both sides of the joblib backend.

There's no new changelog entry because this PR modifies unreleased code.

Closes #109 .

### DIFF
--- a/civis/parallel.py
+++ b/civis/parallel.py
@@ -409,11 +409,33 @@ class JobSubmissionError(Exception):
     pass
 
 
-def _robust_pickle_download(output_file_id, client, n_retries=5, delay=0.0):
+def _robust_pickle_download(output_file_id, client=None,
+                            n_retries=5, delay=0.0):
     """Download and deserialize the result from output_file_id
 
     Retry network errors `n_retries` times with `delay` seconds between calls
+
+    Parameters
+    ----------
+    output_file_id : int
+        ID of the file to download
+    client : civis.APIClient, optional
+    n_retries : int, optional
+        Retry the upload this many times before raising an error.
+    delay : float, optional
+        If provided, wait this many seconds between retries.
+
+    Returns
+    -------
+    obj
+        Any Python object; the result of calling ``joblib.load`` on the
+        downloaded file
+
+    See Also
+    --------
+    joblib.load
     """
+    client = client or civis.APIClient(resources='all')
     retry_exc = (requests.HTTPError,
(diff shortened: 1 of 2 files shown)


civis/parallel.py
```diff
@@ -436,6 +436,63 @@ def _robust_result_download(output_file_id, client, n_retries=5, delay=0.0):
...
+    .. note:: This function starts by calling ``.seek(0)`` on the
+              buffer, and will do so before every retry.
+
+    Parameters
+    ----------
+    buf : File
+        File-like bytes object to send to a Civis File
+    name : str
+        Name of the new Civis File
+    client : civis.APIClient, optional
```
beckermr: Hmmmm. It says the client is optional but it is not given as a keyword argument.

civis/parallel.py
```diff
@@ -409,7 +409,7 @@ class JobSubmissionError(Exception):
     pass
 
 
-def _robust_result_download(output_file_id, client, n_retries=5, delay=0.0):
+def _robust_pickle_download(output_file_id, client, n_retries=5, delay=0.0):
```
beckermr: Did we want a bigger doc string here? The other function has one which is why I am asking.

stephen-hoover: It's private, which is why I didn't bother before. In the other function, I though it was helpful to say what the kwargs were for. I added a longer docstring for this function.

civis/parallel.py
stephen-hoover: Oops. Fixed.

stephen-hoover:
I hadn't made `client` optional before because these are private functions. No harm in making it optional; done.

PR #114, opened 2017-07-18
[CIVP-12028] ENH Allow arbitrary keyword arguments to joblib backend

### LINKED ISSUE #108: Add the `cancel_timeout` parameter to executors and joblib backend
Container scripts take a `cancel_timeout` parameter which provides a delay between a user's cancel request and the system killing the container. This is useful to let code running in the container perform cleanup tasks. We should add this option to the `futures._ContainerShellExecutor` and the joblib backend.

stephen-hoover:
Allow users to pass keyword arguments to ``_ContainerShellExecutor`` and ``make_backend_factory`` and ``infer_backend_factory``. These keyword arguments will be passed directly to ``APIClient.scripts.post_containers``. An example of a new argument which users might want to pass is "cancel_timeout".

There's no modification to the changelog since this PR changes only unreleased code.

Closes #108 .

TODO
- [x] Update `infer_joblib_backend` to infer arbitrary arguments.

### DIFF
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -455,9 +455,8 @@ class _ContainerShellExecutor(_CivisExecutor):
     Parameters
     ----------
     docker_image_name: str, optional
-        The name of the Docker image to be used by Civis.
-    docker_image_tag: str, optional
-        The name of the tag for the Docker image.
+        The name of the Docker image to be used by Civis. You may also
+        wish to specify a ``docker_image_tag`` in the keyword arguments.
     script_name: str, optional
         The name for containers in Civis.
         Defaults to "ContainerShellExecutorScript" followed by the date.
@@ -495,7 +494,6 @@ class _ContainerShellExecutor(_CivisExecutor):
     civis.APIClient.scripts.post_containers
     """
     def __init__(self, docker_image_name="civisanalytics/datascience-python",
-                 docker_image_tag="latest",
                  script_name=None,
                  required_resources=None,
                  hidden=True,
@@ -505,7 +503,6 @@ class _ContainerShellExecutor(_CivisExecutor):
                  inc_script_names=False,
                  **kwargs):
         self.docker_image_name = docker_image_name
-        self.docker_image_tag = docker_image_tag
         self.container_kwargs = kwargs
 
         if required_resources is None:
@@ -541,7 +538,6 @@ class _ContainerShellExecutor(_CivisExecutor):
             required_resources=self.required_resources,
             docker_command=cmd,
             docker_image_name=self.docker_image_name,
-            docker_image_tag=self.docker_image_tag,
             hidden=self.hidden,
             **self.container_kwargs
         )
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -4,6 +4,7 @@ from abc import ABCMeta, abstractmethod
 from builtins import super
 from concurrent.futures import Executor
 from concurrent import futures
+import copy
 import datetime
 import logging
 import time
@@ -457,27 +458,15 @@ class _ContainerShellExecutor(_CivisExecutor):
         The name of the Docker image to be used by Civis.
     docker_image_tag: str
         The name of the tag for the Docker image.
-    repo_http_uri: str
-        The URI for the GitHub repository to check out to /app.
-    repo_ref: str
-        The reference (branch, tag, or commit) for the GitHub repository.
-    git_credential_id: int
-        See :func:`~civis.scripts.post_containers` for details.
     script_name: str
         The name for containers in Civis.
     required_resources: dict
         A dictionary specifying what resources the job needs.
         See :func:`~APIClient.scripts.post_containers` for details.
-    time_zone: str, optional
-        The time zone of this script.
     hidden: bool, optional
         The hidden status of the object. Setting this to ``True`` hides it
         from most API endpoints. The object can still be queried
         directly by ID. Defaults to ``True``.
-    params: list of dict, optional
-        See :func:`~civis.APIClient.scripts.post_containers` for details.
-    arguments: dict, optional
-        See :func:`~civis.APIClient.scripts.post_containers` for details.
     max_n_retries: int, optional
         Retry failed jobs this many times before giving up.
         Retried jobs will be restarted with exactly the same parameters
@@ -495,6 +484,9 @@ class _ContainerShellExecutor(_CivisExecutor):
     inc_script_names: bool, optional
         If ``True``, a counter will be added to the ``script_name`` to create
         the script names for each submission.
+    **kwargs:
+        Additional keyword arguments will be passed
+        directly to ``scripts.post_containers``.
 
     See Also
     --------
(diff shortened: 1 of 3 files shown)


civis/futures.py
```diff
@@ -495,35 +484,27 @@ class _ContainerShellExecutor(_CivisExecutor):
     inc_script_names: bool, optional
         If ``True``, a counter will be added to the ``script_name`` to create
         the script names for each submission.
+    **kwargs:
+        Additional keyword arguments will be passed
+        directly to ``scripts.post_containers``.
```
beckermr: Should this be 

```python
See :func:`~civis.APIClient.scripts.post_containers` for details.
```
?

civis/parallel.py
```diff
@@ -109,6 +110,10 @@ def infer_backend_factory(required_resources=None,
         The hidden status of the object. Setting this to true
         hides it from most API endpoints. The object can still
         be queried directly by ID. Defaults to True.
+    **kwargs:
+        Additional keyword arguments will be passed
+        directly to ``scripts.post_containers``, potentially overriding
```
beckermr: Ditto for the ``scripts.post_containers``.

civis/parallel.py
```diff
@@ -161,34 +166,34 @@ def infer_backend_factory(required_resources=None,
...
+                                **kwargs)
 
 
 def make_backend_factory(docker_image_name="civisanalytics/datascience-python",
                          docker_image_tag="latest",
-                         repo_http_uri=None,
-                         repo_ref=None,
-                         required_resources=None,
-                         params=None,
-                         arguments=None,
```
beckermr: Hmmmm. Why do the `params` and `arguments` keywords still appear here, but not in the `make_backend_factory` function?

civis/parallel.py
```diff
@@ -321,27 +308,27 @@ def make_backend_factory(docker_image_name="civisanalytics/datascience-python",
...
     must have Pandas installed). In particular, the function that is called by
     ``joblib`` must be available in the specified environment.
+
+    See Also
+    --------
+    civis.APIClient.scripts.post_containers
     """
     if setup_cmd is None:
-        if repo_http_uri is not None:
+        if kwargs.get('repo_http_uri'):
```
beckermr: Should we add a `None` default here in case the key is not present in `kwargs`?

civis/parallel.py
```diff
@@ -161,34 +166,34 @@ def infer_backend_factory(required_resources=None,
...
-                                docker_image_tag=state.docker_image_tag,
-                                repo_http_uri=state.repo_http_uri,
-                                repo_ref=state.repo_ref,
-                                required_resources=state.required_resources,
+    # Set defaults on other keyword arguments with
+    # values from the current script
+    for key in ['docker_image_name', 'docker_image_tag', 'repo_http_uri',
+                'repo_ref', 'remote_host_credential_id', 'git_credential_id',
+                'cancel_timeout', 'time_zone']:
+        kwargs.setdefault(key, state[key])
```
beckermr: So the whole point of this change was to avoid any explicit mention of what options were possible. Maybe we iterate over the keys in state?

civis/futures.py
```diff
@@ -495,35 +484,27 @@ class _ContainerShellExecutor(_CivisExecutor):
...
     --------
     civis.APIClient.scripts.post_containers
     """
     def __init__(self, docker_image_name="civisanalytics/datascience-python",
                  docker_image_tag="latest",
-                 repo_http_uri=None,
-                 repo_ref=None,
-                 git_credential_id=None,
                  script_name=None,
                  required_resources=None,
```
beckermr: Why did we leave the required resources as a keyword argument? Ditto for anything set in the API like the docker image stuff, the script names and hidden.

stephen-hoover: The `required_resources` and `docker_image_name` are required arguments to `post_containers`, not optional keyword parameters, so it doesn't seem inconsistent to treat those differently. The bigger reason I left them was that I wanted to control the defaults and keep the defaults explicit. `hidden` defaults to `True` here, which is different from the Platform default. The default on `docker_image_tag` happens to match the Platform default, but that's important enough I'd rather have it explicit.

`script_name` maybe could be left to keyword arguments. Looking at it, I at least need to improve the docstring. The parameter name in `post_containers` is "name", and it's a bit ambiguous to pass a "name" parameter here. Does it mean the name of the executor? I could go either way on that one.

civis/parallel.py
stephen-hoover: That was the first thing I wanted to do. I see two problems -- first, not all keys in the state are valid input parameters. For example, there's a "last_run" and a "state". (And there's no obvious way to introspect valid input parameters. The API client uses **kwargs here.) Second, not all valid inputs are appropriate to transfer to child jobs. For example, I don't think we should reproduce the notification settings.

civis/parallel.py
stephen-hoover: Those are treated specially in the code. They don't clobber values from the existing state, but rather add to them. So you can insert (or replace) additional parameters without losing existing ones, or override a single argument while not changing others.

civis/parallel.py
stephen-hoover: The default default in `get` is `None`.

civis/futures.py
beckermr: I see your point, but am generally not in favor in any case. Special casing the docker image to latest doesn't do a lot since latest can change under you anyways.

How about we remove the docker image tag one and leave (required resources, docker image name, hidden)?

Then later we remove the `hidden=True` default after we have the archive on success option implemented? This seems like a decent solution.

civis/parallel.py
beckermr: Ack yeah for the state and last run. Special casing what not to pass on seems more maintainable? We rarely remove things from the API but do add stuff...

civis/parallel.py
beckermr: Mmmmk. I am not 100% sure this is the best thing to do here, but let's go with it for now. We can deprecate later.

civis/parallel.py
beckermr: Really?! Whoa. (boom) <- what just happened to my head

civis/tests/test_parallel.py
```diff
@@ -312,6 +311,10 @@ def _get_container(job_id):
                        'max_submit_retries': mock.ANY,
                        'max_job_retries': mock.ANY,
                        'hidden': True}
+    for key in ['docker_image_name', 'docker_image_tag', 'repo_http_uri',
+                'repo_ref', 'remote_host_credential_id', 'git_credential_id',
+                'cancel_timeout', 'time_zone']:
+        expected_kwargs[key] = mock_script[key]
```
beckermr: Let's centralize this list in one spot so we don't have to maintain it in the tests and the code?

civis/parallel.py
stephen-hoover: What if the thing that gets added is something we don't want to include? It seems safer to add these by hand. Worst case is that users need to manually specify a new argument in `infer` and not have it introspected.

civis/futures.py
beckermr: We should remove the script name one. This is clearly a case where name is better since it follows the API. The name of the executor is not really a thing anyways. If people are using this, they must have a fair bit of background already.

stephen-hoover: That works. I'll remove the image_tag and note in the `docker_image_name` docstring to be careful of your tag. The `docker_image_tag` defaults to "latest" in Platform anyway.

civis/parallel.py
beckermr: I mean one way or the other right? Happy to leave as is if we move the list to a global and document the hell out of it?

civis/futures.py
stephen-hoover: Changed to use the link.

civis/parallel.py
stephen-hoover: Moved to a global and added into the docstring.

civis/tests/test_parallel.py
stephen-hoover: Done.

stephen-hoover:
I changed "script_name" to "name" but left it in the parameters because we change the default. I think that it's useful to have a different default name so that we can tell the difference between user-created scripts and auto-created scripts.

civis/futures.py
```diff
@@ -453,31 +451,20 @@ class _ContainerShellExecutor(_CivisExecutor):
...
+    name: str, optional
         The name for containers in Civis.
-    required_resources: dict
+        Defaults to "ContainerShellExecutorScript" followed by the date.
+    required_resources: dict, optional
         A dictionary specifying what resources the job needs.
         See :func:`~APIClient.scripts.post_containers` for details.
-    time_zone: str, optional
-        The time zone of this script.
+        Defaults to 1 CPU and 1 GiB of RAM.
```
beckermr: Is this true? I think it defaults to 256 which is 1/4 of a CPU?

beckermr: Ack ignore this one.

PR #117, opened 2017-07-20
[CIVP-10885] ENH Retry connection errors in CLI

stephen-hoover:
When retrieving the API spec in the Civis client CLI, use the same code as we use in the client library. This has the advantage of including automatic retries for connection failures, and automatic waiting for rate limit refreshes. Also modify `invoke` so that we include the same retries as we would if we were using the client library.

Includes a refactor to move all of our connection setup logic to the same place.

The CLI will now use a User-Agent string which starts with "civis-cli" so that we can distinguish it from the Python client.

### DIFF
--- a/civis/_utils.py
+++ b/civis/_utils.py
@@ -32,13 +32,13 @@ def to_camelcase(s):
     return re.sub(r'(^|_)([a-zA-Z])', lambda m: m.group(2).upper(), s)
 
 
-def open_session(api_key, max_retries=5):
+def open_session(api_key, max_retries=5, user_agent="civis-python"):
     """Create a new Session which can connect with the Civis API"""
     civis_version = client_version
     session = requests.Session()
     session.auth = (api_key, '')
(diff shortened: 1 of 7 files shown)


stephen-hoover:
@mheilman , I'm asking you for review because the only functional changes here are to the CLI, and you're most familiar with that. If you don't have time, I can ask someone else.

civis/_utils.py
```diff
@@ -1,6 +1,16 @@
...
+
 import re
 import uuid
 
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util import Retry
+
+import civis
+from civis import __version__ as client_version
```
keithing: Any reason not to alias this like a global var `CLIENT_VERSION`?

stephen-hoover: The explicit version import is really unneeded. I put that in when debugging an import error and forgot to remove it. Fixed.

PR #122, opened 2017-08-15
[CIVP-12351] BUG make sure to pass proper defaults for PARAMS and CVPARAMS

beckermr:
This PR fixes a bug where instantiating a `ModelPipeline` instance from a CivisML model run with a templated script fails due to improper defaults for the `PARAMS` and `CVPARAMS` keys.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Catch unnecessary warning while importing xgboost in CivisML_parallel_training.ipynb (#121)
 
+### Fixed
+- Fixed bug where instantiating a ``ModelPipeline`` from a CivisML model built w/ the templated script breaks.
+
 ## 1.6.0 - 2017-07-27
 ### Changed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - Catch unnecessary warning while importing xgboost in CivisML_parallel_training.ipynb (#121)
 
+### Fixed
+- Fixed bug where instantiating a ``ModelPipeline`` from a CivisML model built w/ the templated script breaks (#122).
```
stephen-hoover: This could be a bit clearer. All CivisML models use custom scripts created from the public templates, and not all `from_existing` calls fail. Perhaps instead of "from a CivisML model built w/ the templated script", use "from an existing model with empty "PARAMS" and "CV_PARAMS" boxes"? It would be helpful to call out `ModelPipeline.from_existing` explicitly, since that's where the problem was.

beckermr:
@stephen-hoover back to you!

PR #127, opened 2017-08-23
DOC Add API client example notebook

stephen-hoover:
This notebook demonstrates the use of all the abstractions and patterns used in the Python API client. It serves as an introduction to the Python API client for technical users of Civis Platform.


stephen-hoover:
Oops. Updated the changelog.

PR #130, opened 2017-08-29
DOC Modify table list in API client demo notebook

stephen-hoover:
Reduce clutter in the list of tables; focus on the ones relevent for the demo rather than filling the screen with the full list.


stephen-hoover:
The original notebook addition from #127 is still unreleased, so I don't think this needs anything new in the changelog.

PR #131, opened 2017-08-29
[PARO-436] BUG raise more informative error when multi-indexed dataframe is passed in

elsander:
Currently, multi-indexed dataframes get read in, and the job fails down the line, throwing an uninformative error. This PR checks for multi-indexed dataframes directly, and throws an informative error to let the user know they are not supported.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -83,6 +83,9 @@ def _block_and_handle_missing(method):
 
 def _stash_local_dataframe(df, client=None):
     """Store data in a temporary Civis File and return the file ID"""
+    # Standard dataframe indexes do not have a "levels" attribute,
+    # but multiindexes do. Checking for this attribute means we don't
+    # need to import pandas to do error handling here.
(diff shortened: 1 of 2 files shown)


civis/ml/_model.py
```diff
@@ -83,6 +83,12 @@ def wrapper(self):
 
 def _stash_local_dataframe(df, client=None):
     """Store data in a temporary Civis File and return the file ID"""
+    # Standard dataframe indexes do not have a "levels" attribute,
+    # but multiindexes do. Checking for this attribute means we don't
+    # need to import pandas to do error handling here.
+    if getattr(getattr(df, "index", None), "levels", None) is not None:
+        raise TypeError("CivisML does not currently support "
```
stephen-hoover: This feels nitpicky, but could you remove the "currently"? We don't have any plans to support multi-indexes, and I don't want to mislead people. You might also add a suggestion in the error message that users could call `.reset_index` to get around this.

PR #133, opened 2017-09-06
BUG Fully implement iterator protocol in PaginatedResponse

stephen-hoover:
The `PaginatedResponse` provided iteration support, but didn't implement the `__next__` method. This meant that the object itself was not an iterator. A small tweak can turn it into an iterator itself. Since the `PaginatedResponse` is returned when users provide an `iterator=True` argument to list endpoints, it seems reasonable for them to expect the returned object is an iterator.

### DIFF
--- a/civis/response.py
+++ b/civis/response.py
@@ -200,3 +200,5 @@ class PaginatedResponse:
         if self._iter is None:
             self._iter = self._get_iter()
         return next(self._iter)
+
+    next = __next__  # Python 2 compatibility
--- a/civis/response.py
+++ b/civis/response.py
@@ -137,7 +137,7 @@ class Response(dict):
 
 
 class PaginatedResponse:
-    """A response object that supports iteration.
+    """A response object which is an iterator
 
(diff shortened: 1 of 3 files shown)


civis/response.py
```diff
@@ -190,3 +195,8 @@ def __iter__(self):
                 yield converted_data
 
             self._params['page_num'] += 1
+
+    def __next__(self):
+        if self._iter is None:
+            self._iter = self._get_iter()
+        return next(self._iter)
```
keithing: Does this need a `next` method to be Python 2 compliant?

https://stackoverflow.com/questions/29578469/how-to-make-an-object-both-a-python2-and-python3-iterator

stephen-hoover:
@keithing , thanks for pointing that out. Given the test failure on the Python 2 Travis build, the iterator response never worked in Python 2. Our test had been wrapping the `PaginatedResponse` in an `iter` call. Without that, the Python 2 iteration broke. This PR fixes it.

PR #134, opened 2017-09-06
BUG ModelPipeline.from_existing handle future training templates

stephen-hoover:
If users called `ModelPipeline.from_existing` on a model training job which used a version of CivisML newer than what the client is aware of, we got errors because the prediction template ID was set to `None`. Let's warn and default to the newest known prediction template. At the very least, this should generate much clearer errors when the prediction job does a version check.


stephen-hoover:
@jmemich , I needed to address a merge conflict in the changelog. Re-approve?

PR #139, opened 2017-09-11
[PARO-439] ENH add civisML 2.0 params to ModelPipeline

elsander:
Add parameters to ModelPipeline for skipping validation, specifying n_jobs during train, and passing a custom ETL estimator.


elsander:
Should we also add a line in `from_existing`  like this?
```
klass._etl_train = args.get('ETL', None)
```
One note about this is that right now, the `_etl_train` attribute is not set to `None` by default; instead, it is not set at all.

civis/ml/_model.py
```diff
@@ -819,6 +820,13 @@ def train(self, df=None, csv_path=None, table_name=None,
         polling_interval : float, optional
             Check for job completion every this number of seconds.
             Do not set if using the notifications endpoint.
+        validate : bool, optional
```
stephen-hoover: We'd like to be able to support user-provided test sets in the future, and I think it would be better to be able to use one parameter to switch between the default of cross-validating over the training data, not testing the model, and testing on user-provided data. Do you think this parameter will still be usable when the extra input options are available?

elsander: Maybe it would be better to have a `validate_data` parameter instead, which defaults to `'train'`, but could be `None` for no validation, or a file_id if the user provides a separate validation set?

civis/ml/_model.py
```diff
@@ -819,6 +820,13 @@ def train(self, df=None, csv_path=None, table_name=None,
...
+            If False, validation will be skipped.
+        n_jobs : int, optional
+            Number of jobs to use for training and validation.
```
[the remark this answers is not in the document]
elsander: Yeah, we should probably just default to 1 for clarity.

civis/ml/_model.py
stephen-hoover: Yes, that could work. Maybe `validation_data` or `test_data`? `validate_data` sounds more like it's asking about data quality checks on the training data.

stephen-hoover:
I think you should set `self._etl_train = None` in the constructor. Good catch on the `from_existing`, though. You'll need to grab "ETL" from the existing training job and store it in `self.etl`, won't you?

elsander:
I did not add `etl` (or `validate`, or `n_jobs`) as an attribute, because they are training-specific (and ETL should get tacked on to the head of the model estimator at the end of training anyway). Users might want to access it, though, so I can add it as a user-facing attribute as well.

stephen-hoover:
Oops, you're right, there is no `self.etl`, so `from_existing` shouldn't set it.

elsander:
Any other changes you would recommend, @stephen-hoover? Otherwise we can put this on hold until the CivisML v2 container is available.

civis/ml/_model.py
```diff
@@ -740,6 +741,7 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
                     git_token_name=git_token_name,
                     verbose=args.get('DEBUG', False))
         klass.train_result_ = fut
+        klass._etl_train = args.get('ETL', None)
```
stephen-hoover: Isn't `_etl_train` an `Estimator` when it gets set in `train`? Does it make sense to have it mean something different here? FWIW, it looks like we aren't setting `self._input_model` here.

elsander:
Ok, made the name change. `validation_data` can be `'train'` (default) to use training data, or `'skip'` to skip validation. See https://github.com/civisanalytics/civis-model/pull/382 for the associated change in `civis-model`.

civis/ml/_model.py
```diff
@@ -639,6 +653,23 @@ def __init__(self, model, dependent_variable,
         self._client = client
         self.train_result_ = None
 
+        if '_NEWEST_CIVISML_VERSION' not in globals():
+            global _NEWEST_CIVISML_VERSION
+            try:
+                client.templates.get_scripts(id=self.train_template_id)
+                client.templates.get_scripts(id=self.predict_template_id)
```
stephen-hoover: It seems like this block of new code (L656 through L671) would work well as a new method to `ModelPipeline`. There should also be a comment to explain why you're calling `templates.get_scripts` twice and ignoring the outputs.

You'll need to call this method ("_set_template_version"?) in `__setstate__` below. Users can unpickle models without going through the constructor.

civis/ml/_model.py
```diff
@@ -747,21 +778,30 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
...
                           % (train_job_id, __version__), RuntimeWarning)
             p_id = max(_PRED_TEMPLATES.values())
         klass.predict_template_id = p_id
 
+        # The user might construct a model from an older version
+        # but still have access to the newest version, so we need
+        # to check the version separately in this function
+        is_newest_version = template_id == max(
+            _PRED_TEMPLATES.keys())
+        if is_newest_version:
```
stephen-hoover: Is it possible to make this `if _NEWEST_CIVISML_VERSION`? If every switch looks the same, it'll be easier to search-and-replace when we need to update these blocks.

civis/ml/_model.py
```diff
@@ -819,6 +859,15 @@ def train(self, df=None, csv_path=None, table_name=None,
         polling_interval : float, optional
             Check for job completion every this number of seconds.
             Do not set if using the notifications endpoint.
+        validation_data : str, optional
+            Source for validation data. There are currently two options:
+            `'train'` (the default), which uses training data for validation;
```
stephen-hoover: "uses training data" --> "cross-validates over training data"? CivisML doesn't construct metrics with in-sample predictions.

civis/ml/_model.py
```diff
@@ -819,6 +859,15 @@ def train(self, df=None, csv_path=None, table_name=None,
...
+            and `'skip'`, which skips the validation step.
+        n_jobs : int, optional
+            Number of jobs to use for training and validation.
```
stephen-hoover: Could you suggest when users might want to change this from the default, and how they should decide where to set it?

civis/ml/_model.py
```diff
@@ -819,6 +859,15 @@ def train(self, df=None, csv_path=None, table_name=None,
...
+        etl : Estimator, optional
+            Custom ETL estimator which overrides the default ETL, and
+            is run before training and validation.
```
stephen-hoover: It seems like this should be part of the constructor. It's part of the model, and we set the model in the constructor.

civis/ml/_model.py
```diff
@@ -857,6 +908,8 @@ def train(self, df=None, csv_path=None, table_name=None,
             train_args['FIT_PARAMS'] = json.dumps(fit_params)
         if self.dependencies:
             train_args['DEPENDENCIES'] = ' '.join(self.dependencies)
+        if n_jobs and self.train_template_id == max(_PRED_TEMPLATES.keys()):
```
stephen-hoover: Why is this one checking `self.train_template_id == max(_PRED_TEMPLATES.keys())` instead of `_NEWEST_CIVISML_VERSION`?

civis/ml/_model.py
```diff
@@ -871,6 +924,19 @@ def train(self, df=None, csv_path=None, table_name=None,
                 self.model = str(estimator_file_id)
             finally:
                 shutil.rmtree(tempdir)
+        if HAS_SKLEARN and isinstance(etl, BaseEstimator):
```
stephen-hoover: Should this also have an `and _NEWEST_CIVISML_VERSION`?

civis/ml/_model.py
```diff
@@ -871,6 +924,19 @@ def train(self, df=None, csv_path=None, table_name=None,
...
+                fout = os.path.join(tempdir, 'ETL.pkl')
+                joblib.dump(etl, fout, compress=3)
+                with open(fout, 'rb') as _fout:
+                    etl_file_id = cio.file_to_civis(
+                        _fout, 'ETL Estimator', client=self._client)
+                train_args['ETL'] = str(etl_file_id)
+            finally:
+                shutil.rmtree(tempdir)
+        if self.train_template_id == max(_PRED_TEMPLATES.keys()):
+            self._etl_train = etl  # Keep the estimator
```
stephen-hoover: Why not put this directly after setting `train_args['ETL']`?

civis/ml/_model.py
elsander: The user could call `from_existing` before initializing any `ModelPipeline`s, so we can't assume that the global is defined. But I can refactor to use the `_set_template_version` method.

Do you think this logic is worth keeping after external release? If so, I might design some tests for this, if only to document the possible cases and expected behavior.

civis/ml/_model.py
elsander: Right now we default to 1. Do we want to do that? Why not automatically parallelize?

civis/ml/_model.py
stephen-hoover: You're calling the `ModelPipeline` constructor in the `cls` call above, so I think this method is covered.

The machinery you're adding here can apply to future release cycles, so I can see us keeping them.

civis/ml/_model.py
elsander: For that matter, `cross_validation_parameters` is also in the constructor. Almost everything is in the constructor. Should we put `validation_data` there too?

stephen-hoover: If the selection were only between "train" and "skip", I'd say yes. But at some point we should add the option to input a new dataset. I think it's reasonable to keep it here. The `ModelPipeline` is not a scikit-learn Estimator, but its construction is similar. So the `train` parameters should be those which are data-dependent.

elsander: Yeah, it wasn't clear to me how to choose where to put the arguments as I was writing the code, but that makes sense. Data-dependent parameters should go in the methods, parameters that can be treated as features of the model should go in the constructor.

civis/ml/_model.py
elsander: Okay, now I remember why I didn't use `_NEWEST_CIVISML_VERSION` here. The user might have access to the newest templates, but still be constructing a model from an older version of CivisML. This means that we need to test the version of the model, not the newest version the user has access to. It's testing something different than `_NEWEST_CIVISML_VERSION`.

civis/ml/_model.py
elsander: What if we default to 4, so that the user always gets the benefit of parallelized cross-validation?

civis/ml/_model.py
stephen-hoover: Is there a reason to set `_etl_train` here? It should already be set in the constructor. Does it matter if we set it even when users don't have access to v2 templates?

civis/ml/_model.py
stephen-hoover: Sounds reasonable to me.

civis/ml/_model.py
elsander: We don't actually ever change `etl` or `_etl_train`, so I don't know that we need `_etl_train` at all.

elsander:
I also added a NotImplementedError in case anyone tries to pass an ETL transformer before they have access to the new templates (the other new arguments have non-None defaults, so I wasn't sure it made sense to raise an error in those cases).

elsander:
Not sure what's causing the Python 3.4 error. It doesn't seem like a test that my code should have affected. I'll look into that more.

elsander:
I can't replicate that error locally (after building a minimal conda env with python 3.4). It's not just a one-time freakout in Travis though, because it's thrown that error on multiple commits. Any ideas, @stephen-hoover?

elsander:
Ah, never mind-- apparently running `pytest` was enforcing python 3.6, while `py.test` used the version in my conda env. I'm replicating the error now.

elsander:
Ok-- looks like some tests got re-added here as the result of a merge. Tests are passing, it should be good for another look.

civis/ml/tests/test_model.py
```diff
@@ -404,18 +404,6 @@ def test_table(mock_res, mock_lt):
                                     index_col=0, client=c)
 
 
-@mock.patch.object(_model, "_load_table_from_outputs", autospec=True)
-@mock.patch.object(_model.ModelFuture, "result")
-@mock.patch.object(_model.ModelFuture, "_set_model_exception", mock.Mock())
-def test_table_None(mock_res, mock_lt):
```
stephen-hoover: Why did you remove this test?

civis/ml/tests/test_model.py
```diff
@@ -540,25 +528,6 @@ def test_metrics_training(mock_file_id_from_run_output):
 
 
 @mock.patch.object(_model.cio, "file_id_from_run_output", autospec=True)
-@mock.patch.object(_model.cio, "file_to_json")
-def test_metrics_training_None(mock_file_to_json, mock_file_id_from_run_output):
```
stephen-hoover: Why remove this test?

civis/ml/tests/test_model.py
```diff
@@ -757,25 +751,6 @@ def test_modelpipeline_train_from_estimator(mock_ccr, mock_f2c):
     assert mock_f2c.call_count == 1  # Called once to store input Estimator
 
 
-@pytest.mark.skipif(not HAS_SKLEARN, reason="scikit-learn not installed")
-@mock.patch.object(_model, "APIClient", mock.Mock())
-@mock.patch.object(_model.cio, "file_to_civis")
-@mock.patch.object(_model.ModelPipeline, "_create_custom_run")
-def test_modelpipeline_train_custom_etl(mock_ccr, mock_f2c, mp_setup):
```
stephen-hoover: Why remove this test?

elsander:
Ah, I feel silly-- I wrote those tests a while ago to test skipping validation. I didn't recognize them, so I thought that they were tests that someone else had removed, and that I had kept them on my branch during a merge. I had been getting some unexpected merge issues in my fork, so I thought that might be the cause of them. I should have `git blame`d them to check.

For some reason, one of those tests is failing with python 3.4, so I'll need to figure out what's going on there before merging.

elsander:
Weird gotcha I discovered-- apparently, in some versions of Python, autospecced mocks don't raise exceptions properly. I don't need the autospec in that test anyway, so removing it should fix the issue.

stephen-hoover:
Should you also restore the `test_modelpipeline_train_custom_etl` test?

elsander:
Weird. I reverted the commit, so I thought I added back all of the tests. I'll put it back in.

elsander:
Ah-- so actually, I removed that one because `etl` is passed to the constructor rather than to the train function now. But I suppose I could still use the test to check that the etl estimator is being stored using `file_to_civis`.

PR #140, opened 2017-09-12
ENH Bind optional arguments to endpoints by name

stephen-hoover:
Previously, optional arguments were all absorbed in a `**kwargs` blob and pulled out later. This changes to set all of them explicitly in the function signature. Doing this improves the readability of the docstrings, and also allows other code (e.g. `mock`'s autospeccer) to read expected function signatures. Note that the actual function code, in `resources._resources.create_method` remains the same. It takes `*args, **kwargs` inputs. The function's signature gets patched after it's created. Because the code remains the same, we never actually transmit default values to the API.

I've set default values to the string "DEFAULT" in an attempt to indicate to users that the defaults in the function signature are different from the defaults in the API.

I'm running `raise_for_unexpected_kwargs` only on keyword arguments. `sig.bind` will only give one unexpected keyword argument even if many are present, so I've preserved our own error handling for that. `sig.bind` will take care of errors for positional arguments and collisions of positional arguments and keyword arguments.

In Python 2, optional arguments will now be able to be given positionally. The previous, keyword-only behavior, is preserved in Python 3.

I've deleted one test from `test_resources` which fails now; it doesn't seem like it should have ever passed. `iterator` wasn't in the spec, so trying to call it should have made an error.

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -9,6 +9,7 @@ except ImportError:
 
 from jsonref import JsonRef
 import requests
+import six
 
 from civis.base import Endpoint, get_base_url
 from civis.compat import lru_cache
@@ -145,7 +146,7 @@ def iterable_method(method, params):
     return (method.lower() == 'get' and params_present)
 
 
-def create_signature(args, kwargs):
+def create_signature(args, optional_args, kwargs=False):
     """ Dynamically create a signature for a function from strings.
 
     This function can be used to create a signature for a dynamically
@@ -156,8 +157,10 @@ def create_signature(args, kwargs):
     ----------
     args : list
         List of strings that name the required arguments of a function.
-    kwargs : list
+    optional_args : list
         List of strings that name the optional arguments of a function.
+    kwargs : bool, optional
+        If True, the function can also accept arbitrary keyword arguments
 
     Returns
     -------
@@ -166,6 +169,12 @@ def create_signature(args, kwargs):
         a dynamically created function.
     """
     p = [Parameter(x, Parameter.POSITIONAL_OR_KEYWORD) for x in args]
+    if six.PY3:
+        opt_par_type = Parameter.KEYWORD_ONLY
+    else:
+        opt_par_type = Parameter.POSITIONAL_OR_KEYWORD
+    p += [Parameter(x, opt_par_type, default='DEFAULT')
+          for x in optional_args]
     if kwargs:
         p.append(Parameter('kwargs', Parameter.VAR_KEYWORD))
     return Signature(p)
@@ -173,7 +182,7 @@ def create_signature(args, kwargs):
 
 def split_method_params(params):
     args = []
-    kwargs = []
+    optional_args = []
     body_params = []
     query_params = []
     path_params = []
@@ -182,14 +191,14 @@ def split_method_params(params):
         if param["required"]:
             args.append(name)
         else:
-            kwargs.append(name)
+            optional_args.append(name)
         if param["in"] == "body":
             body_params.append(name)
         elif param["in"] == "query":
             query_params.append(name)
         elif param["in"] == "path":
             path_params.append(name)
-    return args, kwargs, body_params, query_params, path_params
+    return args, optional_args, body_params, query_params, path_params
 
 
 def create_method(params, verb, method_name, path, doc):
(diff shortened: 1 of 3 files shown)


civis/tests/test_resources.py
```diff
@@ -42,20 +42,6 @@
...
-            {"name": 'page_num', "in": 'query', "required": False, "doc": ""},
-            {"name": 'order', "in": 'query', "required": False, "doc": ""},
-            {"name": 'order_by', "in": 'query', "required": False, "doc": ""}]
-    method = _resources.create_method(args, 'get', 'mock_name', '/objects',
-                                      'fake_doc')
-    mock_endpoint = mock.MagicMock()
-
-    method(mock_endpoint, iterator=True)
-    mock_endpoint._call_api.assert_called_once_with(
-        'get', '/objects', {}, {}, iterator=True)
```
keithing: I think this test is actually okay.  `iterator` is in the spec for `_call_api`, since `_call_api` accepts `**kwargs` (not sure if this is actually for the best, though).   [Here](https://github.com/civisanalytics/civis-python/blob/5c135f9482ed6ca00c83871fb40b3cd35fc3045c/civis/base.py#L109) is the corresponding line.

However, this test should fail if `{"name": 'limit', ...}` and `{"name": 'page_num'}` isn't included in `args`, since this would cause [`iterable_method`](https://github.com/civisanalytics/civis-python/blob/master/civis/resources/_resources.py#L227) to return`False`, and then [`raise_for_unexpected_kwargs`](https://github.com/civisanalytics/civis-python/blob/master/civis/resources/_resources.py#L250) should raise a `TypeError`.

This is pretty confusing, does that trail of breadcrumbs follow your understanding as well?

stephen-hoover: Ah, I see. I'd gotten confused about what the `iterator` argument was doing. I thought it was part of the Platform spec, but it's a feature of the API client. I needed to modify `create_method` to pop out `iterator` before binding the arguments, but the test is restored. This means that `iterator` is a keyword-only argument in both Python 2 and Python 3, and is the only keyword-only argument in Python 2.

civis/resources/_resources.py
```diff
@@ -166,14 +169,20 @@ def create_signature(args, kwargs):
         a dynamically created function.
     """
     p = [Parameter(x, Parameter.POSITIONAL_OR_KEYWORD) for x in args]
+    if six.PY3:
+        opt_par_type = Parameter.KEYWORD_ONLY
```
keithing: Why `KEYWORD_ONLY` here?  IIUC, this would mean a function like this

```python
def f(a, b, c=3):
    return a + b + c
```

Could not be called like `f(1, 2, 3)` and would have to be called with `f(1, 2, c = 3)`.  Granted, since we've been using kwargs up until now, this won't break any code to use `KEYWORD_ONLY`.  But I think most users would expect the `POSTITIONAL_OR_KEYWORD` calling convention.

stephen-hoover: If you're defining the function yourself, it would be
```python
def f(a, b, *, c=3):
    return a + b + c
```
and yes, you couldn't call it as `f(1, 1, 1)`.

The `KEYWORD_ONLY` preserves the existing behavior. Is the order of the optional arguments guaranteed to be stable? New arguments get added to API endpoints sometimes. Are new arguments guaranteed to always come after any existing arguments? If so, it seems okay to use `POSITIONAL_OR_KEYWORD`. However, I think that enforcing keyword only seems closer to the way the API is used, and more future-proof.

keithing: Yeah, I agree that it is a cleaner API to have `KEYWORD_ONLY` given that we don't have guarantees that the order of keyword args is not going to change.  I do think this will confuse some users, but I suppose it would be better to give a confusing error during development than have code break later because argument order changed.

civis/resources/_resources.py
```diff
@@ -145,7 +146,7 @@ def iterable_method(method, params):
     return (method.lower() == 'get' and params_present)
 
 
-def create_signature(args, kwargs):
+def create_signature(args, optional_args, kwargs=False):
```
keithing: do we ever use `kwargs` here?

stephen-hoover: No. Shall I remove it?

keithing: Sure. We can add it back if we need it later.

PR #141, opened 2017-09-12
TST Provide autospecced client mock for testing

stephen-hoover:
This PR provides an autospecced mock of the full `APIClient` object, suitable for testing. The PR includes some refactoring around fetching and retrieving the cached spec, which we need for creating the mock client object.


stephen-hoover:
@keithing , thoughts? Some or all of the refactoring could be rolled back if you prefer. This PR includes an update to one of the stored API specs in the repo, but only the "_channels" spec. Updating the other one causes many tests to fail because VCR doesn't expect the channels endpoint to be present.

keithing:
> Updating the other one causes many tests to fail because VCR doesn't expect the channels endpoint to be present.

The VCR tests do seem a little fragile.  Do you know if there is anything we could change so we don't need to commit two different specs? That will probably increase confusion. I don't know the scope of this issue here.

stephen-hoover:
The two specs are pre-existing. We had one with the "channels" endpoint and one without. I updated the specs with the "channels" endpoint. I think that it would be a good idea to clean up / update the VCR tests, but I think it's out of scope for this PR. I'm not sure what it would take to do that.

keithing:
> The two specs are pre-existing. 

Oh, I missed that.  If it's already in then it fine to keep it in.  Eventually I think we should just have one to be consistent.

> I think that it would be a good idea to clean up / update the VCR tests, but I think it's out of scope for this PR

Oh yes, most certainly.

Issue #143, opened 2017-09-18
Pandas Column Typing


gdj0nes:
**Overview:** Add functionality to get column types, including date, and sortkeys from Redshift meta info for Pandas IO

I modified the client to allow me to pass arguments to the pd.read_csv on line 227 in civis/io/_tables.py. I gather the column types using client.tables.get after which I map the SQL types to their corresponding Python types. Doing this eliminates the automatic str to int conversion, e.g. zipcodes, and automatically identifies the date columns.  Is there any interest in adding this in a more thoughtful way to the code?

beckermr:
Yes this is a good idea! I am not sure who has time to get to it, but if you want to make a PR feel free!

stephen-hoover:
+1 ! You can tag me for review. I've dealt with similar issues in my own code.

gdj0nes:
I have the time right now but I'm not super familiar with the code. Basically what's the fastest way to get the table meta given the database and table names because that seems to be readily available in the existing code?  I currently do the following in my code:

1. Get the table in the database by searching over the tables:
    `client.tables.list(database_id=DATABASE_META['id'], limit=1000)`
2. Use the table id to get column data
    `table_cols = client.tables.get(table['id'])['columns']`

beckermr:
Hmmm. Those calls might require the table scanner. You could query the database for the info directly.

stephen-hoover:
The types will be populated without a full table scanner run. You can use something like
```python
db_id = client.get_database_id(database)
tables = client.tables.list(database_id=db_id, schema=schema, name=table)
columns = client.tables.get(tables[0].id).columns
dtypes = {col.name: redshift_to_py_type(col.sql_type)
              for col in columns}
```

gdj0nes:
Here is the function I plan on using:

`from civis.APIClient import get_table_id`

gdj0nes:
Initial code where table_id is passed from `read_civis` into `read_civis_sql`
```python
    if use_pandas:
        table_meta = client.tables.get(table_id)
        # The split is to handle e.g. DECIMAL(17,9)
        redshift_types = {col['name']: col['sql_type'].split('(')[0]
                              for col in table_meta['columns']}
        py_types = {name : SQL_PANDAS_MAP[type_]
                    for (name, type_) in redshift_types 
                    if type_ not in DATE_TYPES}
        date_cols = [name for (name, type_) in redshift_types 
                     if type_ in DATE_TYPES]
                     
[...]

PR #144, opened 2017-09-19
Pandas IO Typing

gdj0nes:
The new code provides a typed Pandas Dataframe based on the Redshift analogue when `use_pandas=True ` with `civis.io.read_civis` based on #143. The last remaining component is to verify the testing tools. This entails updating the sql table on platform with columns with the types we would like to test.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -3,6 +3,7 @@ import csv
 import io
 import six
 import warnings
+import numpy as np
 
 from civis import APIClient
 from civis.io import civis_to_file
@@ -10,8 +11,6 @@ from civis._utils import maybe_get_random_name
 from civis.base import EmptyResultError
 from civis.futures import CivisFuture
 from civis.utils._deprecation import deprecate_param
-import numpy as np
-from pprint import pprint
 
 import requests
 
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -10,6 +10,8 @@ from civis._utils import maybe_get_random_name
 from civis.base import EmptyResultError
 from civis.futures import CivisFuture
 from civis.utils._deprecation import deprecate_param
+import numpy as np
+from pprint import pprint
 
 import requests
 
@@ -32,22 +34,35 @@ DELIMITERS = {
     '|': 'pipe',
 }
 
-SQL_PANDAS_MAP = {
-                  'numeric' : np.float,
-                  'real': np.float,
-                  'smallint': np.int, 
-                  'integer' : np.int,
-                  'bigint' : np.int,
-                  'deicmal':np.float,
-                  'real':np.float,
-                  'double':np.float,a
-                  'boolean' : np.bool,
-                  'char': np.str,
-                  'character varying': np.str
-}
+
+SQL_PANDAS_MAP = {'smallint': np.int,
+               'int': np.int,
+               'integer': np.int,
+               'bigint': np.int,
+               'decimal': np.float,
+               'numeric': np.float,
+               'float': np.float,
+               'float4': np.float,
+               'float8': np.float,
+               'real': np.float,
+               'double precision': np.float,
+               'boolean': np.bool,
+               'bool': np.bool,
+               'int2': np.int,
+               'int4': np.int,
+               'int8': np.int,
+               'char': np.str,
+               'character': np.str,
+               'nchar': np.str,
+               'bpchar': np.str,
+               'varchar': np.str,
+               'character varying': np.str,
+               'nvarchar': np.str,
+               'text': np.str}
 
 # Checkout timestamptz
-DATE_TYPES = ['date', 'timestamp', 'timestamptz']
+DATE_TYPES = ['date', 'timestamp', 
+              'timestamptz', 'timestamp without time zone']
 
 
 @deprecate_param('v2.0.0', 'api_key')
@@ -134,7 +149,7 @@ def read_civis(table, database, columns=None, use_pandas=False,
         # Instantiate client here in case users provide a (deprecated) api_key
         client = APIClient(api_key=api_key, resources='all')
 
-    
+
     table_id = client.get_table_id(table=table, database=database)
 
     sql = _get_sql_select(table, columns)
@@ -247,22 +262,13 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
                                .format(script_id))
     url = outputs[0]["path"]
     if use_pandas:
-        table_meta = client.tables.get(table_id)
-        # The split is to handle e.g. DECIMAL(17,9)
-        redshift_types = {col['name']: col['sql_type'].split('(')[0]
-                              for col in table_meta['columns']}
-        py_types = {name : SQL_PANDAS_MAP[type_]
-                    for (name, type_) in redshift_types 
-                    if type_ not in DATE_TYPES}
-        date_cols = [name for (name, type_) in redshift_types 
-                     if type_ in DATE_TYPES]
-                     
-        sortkeys = table_meta['sortkeys'].split(',')
+
+        sortkeys, py_types, date_cols = _redshift_to_py(client, table_id)
         # Not sure about the best way to handle null fields
-        # or how they may appear
         data = pd.read_csv(url, parse_dates=date_cols,
-                                index_col=sort_keys,
-                                converters=py_types)
+                                index_col=sortkeys,
+                                converters=py_types,
+                                engine='python')
     else:
         r = requests.get(url)
         r.raise_for_status()
@@ -765,3 +771,17 @@ def _import_bytes(buf, database, table, client, max_errors,
 
         fut.add_done_callback(f)
     return fut
+
+def _redshift_to_py(civ_client, tableid):
+        table_meta = civ_client.tables.get(tableid)
+        # The split is to handle e.g. DECIMAL(17,9)
+        redshift_types = {col['name']: col['sql_type'].split('(')[0]
+                              for col in table_meta['columns']}
+        py_types = {name : SQL_PANDAS_MAP[type_]
+                    for (name, type_) in redshift_types.items() 
+                    if type_ not in DATE_TYPES}
+        date_cols = [name for (name, type_) in redshift_types.items() 
+                     if type_ in DATE_TYPES]
+
+        sortkeys = table_meta['sortkeys'].split(',')
+        return sortkeys, py_types, date_cols
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -67,8 +67,7 @@ DATE_TYPES = ['date', 'timestamp',
 @deprecate_param('v2.0.0', 'api_key')
 def read_civis(table, database, columns=None, use_pandas=False,
                job_name=None, api_key=None, client=None, credential_id=None,
-               polling_interval=None, archive=False, hidden=True,
-               pandas_kwargs=None, **kwargs):
+               polling_interval=None, archive=False, hidden=True, **kwargs):
     """Read data from a Civis table.
 
     Parameters
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -147,9 +147,7 @@ def read_civis(table, database, columns=None, use_pandas=False,
         # Instantiate client here in case users provide a (deprecated) api_key
         client = APIClient(api_key=api_key, resources='all')
 
-
     table_id = client.get_table_id(table=table, database=database)
-
     sql = _get_sql_select(table, columns)
     data = read_civis_sql(sql=sql, database=database, use_pandas=use_pandas,
                           job_name=job_name, client=client, 
@@ -260,7 +258,6 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
                                .format(script_id))
     url = outputs[0]["path"]
     if use_pandas:
-
         sortkeys, py_types, date_cols = _redshift_to_py(client, table_id)
         # Not sure about the best way to handle null fields
         data = pd.read_csv(url, parse_dates=date_cols,
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -32,11 +32,29 @@ DELIMITERS = {
     '|': 'pipe',
 }
 
+SQL_PANDAS_MAP = {
+                  'numeric' : np.float,
+                  'real': np.float,
+                  'smallint': np.int, 
+                  'integer' : np.int,
+                  'bigint' : np.int,
+                  'deicmal':np.float,
+                  'real':np.float,
+                  'double':np.float,a
+                  'boolean' : np.bool,
+                  'char': np.str,
+                  'character varying': np.str
+}
+
+# Checkout timestamptz
+DATE_TYPES = ['date', 'timestamp', 'timestamptz']
+
 
 @deprecate_param('v2.0.0', 'api_key')
 def read_civis(table, database, columns=None, use_pandas=False,
                job_name=None, api_key=None, client=None, credential_id=None,
-               polling_interval=None, archive=False, hidden=True, **kwargs):
+               polling_interval=None, archive=False, hidden=True,
+               pandas_kwargs=None, **kwargs):
     """Read data from a Civis table.
 
     Parameters
@@ -115,9 +133,14 @@ def read_civis(table, database, columns=None, use_pandas=False,
     if client is None:
         # Instantiate client here in case users provide a (deprecated) api_key
         client = APIClient(api_key=api_key, resources='all')
+
+    
+    table_id = client.get_table_id(table=table, database=database)
+
     sql = _get_sql_select(table, columns)
     data = read_civis_sql(sql=sql, database=database, use_pandas=use_pandas,
-                          job_name=job_name, client=client,
+                          job_name=job_name, client=client, 
+                          table_id=table_id,
                           credential_id=credential_id,
                           polling_interval=polling_interval,
                           archive=archive, hidden=hidden, **kwargs)
@@ -126,9 +149,9 @@ def read_civis(table, database, columns=None, use_pandas=False,
 
 @deprecate_param('v2.0.0', 'api_key')
 def read_civis_sql(sql, database, use_pandas=False, job_name=None,
-                   api_key=None, client=None, credential_id=None,
-                   polling_interval=None, archive=False,
-                   hidden=True, **kwargs):
+                   api_key=None, client=None, table_id=None,
+                   credential_id=None, polling_interval=None, 
+                   archive=False, hidden=True, **kwargs):
     """Read data from Civis using a custom SQL string.
 
     Parameters
@@ -224,7 +247,22 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
                                .format(script_id))
     url = outputs[0]["path"]
     if use_pandas:
-        data = pd.read_csv(url, **kwargs)
+        table_meta = client.tables.get(table_id)
+        # The split is to handle e.g. DECIMAL(17,9)
+        redshift_types = {col['name']: col['sql_type'].split('(')[0]
+                              for col in table_meta['columns']}
+        py_types = {name : SQL_PANDAS_MAP[type_]
+                    for (name, type_) in redshift_types 
+                    if type_ not in DATE_TYPES}
+        date_cols = [name for (name, type_) in redshift_types 
+                     if type_ in DATE_TYPES]
+                     
+        sortkeys = table_meta['sortkeys'].split(',')
+        # Not sure about the best way to handle null fields
+        # or how they may appear
+        data = pd.read_csv(url, parse_dates=date_cols,
+                                index_col=sort_keys,
+                                converters=py_types)
     else:
         r = requests.get(url)
         r.raise_for_status()
--- a/civis/tests/test_io.py
+++ b/civis/tests/test_io.py
@@ -159,14 +159,16 @@ class ImportTests(CivisVCRTestCase):
                   'char': np.str,
                   'character varying': np.str
                   }
-        expected = pd.DataFrame([[1., 2., 3, 4, 5, 6., 7., 8., True, 'f', 'bar', '05/07/1995']], 
-                                 columns=['numeric', 'real', 'smallint', 
+        expected = pd.DataFrame(data=[[1., 2., 3, 4, 5, 6., 7., 8.,
+                                  True, 'f', 'bar', '05/07/1995']], 
+                                columns=['numeric', 'real', 'smallint', 
                                  'integer', 'bigint', 'deicmal', 'real', 
                                  'double', 'boolean', 'char', 
                                  'character varying', 'date'
                                  ],
-                     parse_dates=['date'],
-                     converters=col_types)
+                                parse_dates=['date'],
+                                converters=col_types)
+        
         df = civis.io.read_civis('scratch.api_client_test_fixture',
                                  'redshift-general', use_pandas=True,
                                  polling_interval=POLL_INTERVAL)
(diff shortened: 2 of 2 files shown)


civis/io/_tables.py
```diff
@@ -33,6 +34,36 @@
...
+               'int4': np.int,
+               'int8': np.int,
+               'char': np.str,
+               'character': np.str,
+               'nchar': np.str,
+               'bpchar': np.str,
+               'varchar': np.str,
+               'character varying': np.str,
+               'nvarchar': np.str,
+               'text': np.str}
```
beckermr: We might want to pick the numpy type that is appropriately sized for the redshift type. For example, for 'float8' we should prob use `np.float64`, etc. Ditto for the various int types and the rest of the float types.

civis/io/_tables.py
```diff
@@ -33,6 +34,36 @@
...
+
+# Checkout timestamptz
```
beckermr: Can you make this comment more verbose?

civis/io/_tables.py
```diff
@@ -126,9 +160,9 @@ def read_civis(table, database, columns=None, use_pandas=False,
 
 @deprecate_param('v2.0.0', 'api_key')
 def read_civis_sql(sql, database, use_pandas=False, job_name=None,
-                   api_key=None, client=None, credential_id=None,
-                   polling_interval=None, archive=False,
-                   hidden=True, **kwargs):
+                   api_key=None, client=None, table_id=None,
```
beckermr: You need to add documentation for any new parameters.

civis/io/_tables.py
```diff
@@ -224,7 +258,12 @@ def f(x):
                                .format(script_id))
     url = outputs[0]["path"]
     if use_pandas:
-        data = pd.read_csv(url, **kwargs)
+        sortkeys, py_types, date_cols = _redshift_to_py(client, table_id)
```
beckermr: You need to test here if `table_id` is `None`. If it is, then we should revert back to pandas type inference.

civis/io/_tables.py
```diff
@@ -224,7 +258,12 @@ def f(x):
...
+        # Not sure about the best way to handle null fields
+        data = pd.read_csv(url, parse_dates=date_cols,
+                                index_col=sortkeys,
+                                converters=py_types,
+                                engine='python')
```
beckermr: Can you explain this particular invocation of `pd.read_csv`? Also, we should not be setting the index of the output DF here. The convention used here IIUIC is that the DF row index is always integers and any dist or sort keys in redshift are left as normal columns. Finally, I don't think we need to revert to the python engine here, but I could be wrong.

gdj0nes: I'll drop the sort keys. The python engine is much more robust and sometimes you get weird errors trying to make type conversions with the c engine, furthermore the performance difference seems negligible. When you say invocation what do you mean, the keyword args? The original was `pd.read_csv(url)`

civis/io/_tables.py
gdj0nes: I can just delete this comment because I got all of the aliases

civis/io/_tables.py
gdj0nes: I had considered that but I've found in a private repo of ours that has similar functionality and they convert all numerical to np.float but I thought it would be fine to leave the ints.

civis/io/_tables.py
beckermr: Yeah I just meant the overall function signature and the choice of keywords. Have you tested performance with very big tables? I think this might be where that matters but I am not sure.

civis/io/_tables.py
beckermr: I think here we want to preserve the type sizes properly since we don't know what users are doing with the data. We can chat about the other stuff offline. I think I understand the confusion.

stephen-hoover:
I recommend not making any changes to `io.read_civis_sql`. That function executes arbitrary SQL, so I don't think we can have any expectation of knowing the types that come out. Stick to grabbing types for `io.read_civis`. You can use the `kwargs` in `read_civis_sql` to pass in a `dtype` parameter for `read_csv`.

stephen-hoover:
I'm still concerned about the point that @keithing raised in #143 . Users can alter the type of columns in `read_civis`. I don't know what the best way to handle that would be. Perhaps we don't read Redshift types if `columns` is non-`None`?

stephen-hoover:
Another comment -- in the future, please make PRs from a fork, not from a branch of this repo.

beckermr:
Good point @stephen-hoover. RE the column transformation issue, I think the use case of asking only for some columns, but no transformations and wanting types makes a lot of sense. So I am bit worried about the guard rail as you described it.

beckermr:
I like the idea of passing the kwargs to read_civis_sql.

gdj0nes:
@stephen-hoover sorry for my inexperience with the conventions, I'll keep that in mind in the future.  I also like the kwargs approach.

stephen-hoover:
Then maybe we need a switch to turn off automatic type setting? Alternately, users can already pass in `dtype` to `read_civis` if they know the types they expect. We won't be able to handle everything, and if users pass a non-`None` value to `columns`, we have no guarantees that the types in the Redshift table will be the same as the output types.

beckermr:
I am ok with a switch. Something like `exact_types=False` or `infer_types=True`. It might be good to try and match the pandas API?

civis/io/_tables.py
gdj0nes: I have a table that's over 1.5GB and I can't say I noticed any difference switching between the two but it did crash when there was an integer column that had some funny formatting when using the C engine. See preformance comparison: [here](https://stackoverflow.com/questions/29678154/why-is-numpy-pandas-parsing-of-a-csv-file-with-long-lines-so-slow)

gdj0nes:
@stephen-hoover if we want users to pass the dtypes manually should we create a helper function that grabs the column dtypes for them outside of `read_civis` to expedite the process?

@beckermr if we use your approach would we add this keyword to the `read_civis` function and then if it is `True` grab the table meta in `read_civis` and then pass it as kwargs to `read_civis_sql`?

Is there a consensus on how to go forward with this?

stephen-hoover:
I don't think it's necessary to add a public helper function here.

What about the following: Add an explicit `dtype` parameter to `read_civis`. Default it to "redshift". If the value is "redshift", then replace it with data types read from the specified table. Then pass `dtype` (possibly replaced with types read from Redshift, possibly not) to `read_civis_sql`.

beckermr:
I like this suggestion @stephen-hoover. Do we want to default to None?

keithing:
Would it be a breaking change if we default a new `dtype` parameter to `redshift`?

stephen-hoover:
It might change the function's output, and it could break previously functioning code if users were using `columns` to modify output types. Maybe it should default to `None`. Reading types from Redshift seems like a nicer default, though -- perhaps we can make a note to switch the default to "redshift" if/when we make a v2 release?

gdj0nes:
I'm worried that if we have the default being redshift and a user is currently passing their own dtypes through kwargs then we overwrite their dtypes without them knowing. It also seems that dtypes should boolean because it has no other use case then grabbing redshift types.

stephen-hoover:
If users are already setting `dtype`, that will override our default. My suggestion is that we add an extra option for what users could pass to `dtype`, not create a new parameter. (See https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html for documentation on the `dtype` parameter.)

beckermr:
Yep @stephen-hoover is right. Here is what we would have

- `dtype='redshift'` - infer types
- `dtype=None` - let pandas infer the types as usual (default)
- `dtype=<anything else accepted by pandas>` - let pandas do its thing (item above is a special case of this)

gdj0nes:
@beckermr If it is anything else excepted by pandas should we change it to `pd_options` or some other more descriptive name?

beckermr:
Nah. Pandads calls it dtype, so let's call it dtype. It is anything else accepted by the pandas dtype pasrameter on read_csv, so this makes sense. Sorry I was unclear about that!

gdj0nes:
Any thoughts on how to update the testing suite?

beckermr:
We should do this in person. Feel free to add 30 min to my calendar.

civis/io/_tables.py
```diff
@@ -115,9 +149,24 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+        if dtype == 'redshift':
+            table_id = client.get_table_id(table=table, database=database)
+            py_types, date_cols = _redshift_to_py(client, table_id)
+            pd_kwargs = {'parse_dates': date_cols,
+                         'dtype': py_types,
+                         'engine': 'python',
+                         'infer_datetime_format': True}
+            kwargs = kwargs.update(pd_kwargs)
+        elif type(dtype) == dict:
+            kwargs = kwargs.update(pd_kwargs)
```
beckermr: This line is wrong. You want `kwargs['dtype'] = dtype` I think.

civis/io/_tables.py
```diff
@@ -115,9 +149,24 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+        else:
```
beckermr: No need for this block.

civis/io/_tables.py
```diff
@@ -115,9 +149,24 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+        elif type(dtype) == dict:
+            kwargs = kwargs.update(pd_kwargs)
+        else: 
+            continue
+
     sql = _get_sql_select(table, columns)
     data = read_civis_sql(sql=sql, database=database, use_pandas=use_pandas,
-                          job_name=job_name, client=client,
+                          job_name=job_name, client=client, 
+                          table_id=table_id,
```
beckermr: This can be removed.

civis/io/_tables.py
```diff
@@ -127,7 +176,7 @@ def read_civis(table, database, columns=None, use_pandas=False,
 @deprecate_param('v2.0.0', 'api_key')
 def read_civis_sql(sql, database, use_pandas=False, job_name=None,
                    api_key=None, client=None, credential_id=None,
-                   polling_interval=None, archive=False,
+                   polling_interval=None, archive=False,
```
beckermr: flake8 will complain about this whitespace change...ugh :(

gdj0nes:
I've encountered an issue related to null types and integers. In the case that there is a`null` type in an integer column, the converter will fail because `NaN`s are floats in Numpy. Possible Solutions:

1. Try block reverts back to the default behavior with a warning message
2. Try block and switch to all (we don't know which column caused the problem)  ints to floats

stephen-hoover:
Ugh, that's annoying. Another option is to read the DataFrame with all integer types set to float and then try one-by-one to cast the columns back to integers. It seems like that might be faster than a try/except on reading the entire DataFrame. Still potentially slow, though.

We could even consider reading all integer types as floats and not bothering to keep them as ints.

Note that with anything involving reading integer types as floats, we need to use more bits of precision in the float types than in the Redshift integer type. Otherwise we'll end up losing precision. Some of the float's bits are used for the exponent, so it can't exactly represent all of the integers in an integer type with the same number of bits.

beckermr:
Yeah I actually like Stephen's suggestion. It is also probably the only workable one.

beckermr:
The one I meant was just reading int types as float and then casting back if we can.

gdj0nes:
Thoughts on how to modify the testing function? Should I just update `scratch.api_client_test_fixture` with a var char, bool, int, float, null int, and date?

jseabold:
Hey, cool! Before you get too far, I'd recommend at least having a look at how sqlalchemy handles this. You'll get all the standard types from postgres + any redshift specific ones via redshift-sqlalchemy. Here's some helper functions to make a pandas-ready dtype from redshift. Should cover every type, but I'm not positive I didn't stop along the way.

https://github.com/civisanalytics/modeling-dsrd/blob/dask-redshift/dask-redshift/dask_redshift/dask_redshift.py#L12

As for integers and nulls, the best you can do is [check if an integer is nullable](https://github.com/civisanalytics/modeling-dsrd/blob/dask-redshift/dask-redshift/dask_redshift/dask_redshift.py#L69) and use a float. Otherwise, we'd have to pass over all of the data.

jseabold:
I remember at least one place I stopped. The char types can have an encoding associated with them, I didn't try to use it yet.

stephen-hoover:
What exactly is the "notnull" column in pg_table_def? The docs say "True if the column has a NOT NULL constraint.", but what is a "NOT NULL" constraint? I'm a bit concerned by that conversion. Forcing ints into a float32 will make some distinct integers no longer distinct. This has caused bugs for us in the past.

jseabold:
>  I'm a bit concerned by that conversion. Forcing ints into a float32 will make some distinct integers no longer distinct. This has caused bugs for us in the past.

Ah, right.

From the docs:

>>> NOT NULL specifies that the column is not allowed to contain null values. NULL, the default, specifies that the column accepts null values. IDENTITY columns are declared NOT NULL by default.

beckermr:
Generally the client doesn’t uses direct database connection since it goes through the API. Is it possible to use sqlalchemy without a database connection? Or do you just mean look at the code for how they do this?

stephen-hoover:
I've checked some of my own tables, and I have some with integer ID columns which are marked nullable (but which don't have nulls). We should make sure to represent integer columns with enough bits of precision that we can represent all allowable integers.

civis/io/_tables.py
```diff
@@ -727,3 +762,48 @@ def f(x):
...
+        'int4': np.int32,
+        'int8': np.int64,
+        'char': np.str,
+        'character': np.str,
+        'nchar': np.str,
+        'bpchar': np.str,
+        'varchar': np.str,
+        'character varying': np.str,
+        'nvarchar': np.str,
+        'text': np.str}
```
beckermr: @jseabold I think the type map here is pretty complete already. We just have to deal with nulls. Probs best to read as float64 and then do a quick check and convert back to the right in type if there are not any nulls.

beckermr: redshift by default uses UTF-8 encoding. Thus I think we should make sure to account for that and then punt on the rest.

gdj0nes: I've added a utf-8 encoding parameter to the kwargs

gdj0nes:
It turns out that you can't cast back to integer after the import because the problem still remains that NaNs can't be converted to Int types.

beckermr:
You need to test for nans first and if there are none, then cast back, otherwise don’t.

gdj0nes:
Are there any thoughts about how the test should be incorporated?

civis/io/_tables.py
```diff
@@ -727,3 +765,48 @@ def f(x):
 
         fut.add_done_callback(f)
     return fut
+
+
+def _redshift_to_py(civ_client, tableid):
```
stephen-hoover: This library has the convention of naming `APIClient` objects as `client`. You should use that instead of `civ_client` here.

civis/io/_tables.py
```diff
@@ -115,6 +127,23 @@ def read_civis(table, database, columns=None, use_pandas=False,
     if client is None:
         # Instantiate client here in case users provide a (deprecated) api_key
         client = APIClient(api_key=api_key, resources='all')
+
+    if use_pandas == True:
+        if dtype == 'redshift':
+            if NO_NUMPY:
+                raise ImportError("`dtype='redshift'` specified but numpy not installed.")
```
stephen-hoover: Is this `numpy` check needed? If we have `pandas`, then we know that we have `numpy`.

civis/io/_tables.py
```diff
@@ -727,3 +765,48 @@ def f(x):
 
         fut.add_done_callback(f)
     return fut
+
+
+def _redshift_to_py(civ_client, tableid):
+
+    SQL_PANDAS_MAP = {
```
stephen-hoover: I suggest moving this dictionary and `DATE_TYPES` into module-level constants. They seem like something worth putting at the top of the file.

civis/tests/test_io.py
```diff
@@ -152,6 +159,61 @@ def test_read_civis_pandas(self, *mocks):
...
+
+        # Construct all of these column types
+
+        pd_data = {}
+        for num, (name, type) in enumerate(SQL_PANDAS_MAP.items()):
+            if name in ['boolean', 'bool']:
+                value = [True]
+            elif name in ['char', 'character', 'nchar', 'bpchar', 'varchar',
+                          'character varying', 'nvarchar', 'text']:
+                value = ['foo']
```
stephen-hoover: Try making this something like `"13"`. If pandas can infer types, then we aren't testing all of the new code.

civis/tests/test_io.py
```diff
@@ -152,6 +159,61 @@ def test_read_civis_pandas(self, *mocks):
...
+            else:
+                value = [num]
+            pd_data[name] = value
+
+        expected = pd.DataFrame
+
+        df = civis.io.read_civis('scratch.api_client_test_fixture',
+                                 'redshift-general', use_pandas=True,
+                                 dtype='redshift', polling_interval=POLL_INTERVAL)
+        assert df.equals(expected)
```
stephen-hoover: Doing this test in the same way as `test_read_civis_pandas` would require recording a new VCR cassette.

civis/io/_tables.py
```diff
@@ -115,6 +127,23 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+    if use_pandas == True:
+        if dtype == 'redshift':
+            if NO_NUMPY:
+                raise ImportError("`dtype='redshift'` specified but numpy not installed.")
+            else:
+                table_id = client.get_table_id(table=table, database=database)
+                py_types, date_cols = _redshift_to_py(client, table_id)
+                pd_kwargs = {'parse_dates': date_cols,
+                             'dtype': py_types,
+                             'engine': 'python',
```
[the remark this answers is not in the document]
gdj0nes: I defaulted to this because it is more robust and the performance benefits of the C engine seem negligible for our data

civis/io/_tables.py
```diff
@@ -115,6 +127,23 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+                             'infer_datetime_format': True,
+                             'encoding': 'utf8'}
```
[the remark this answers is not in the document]
gdj0nes: see previous comment

civis/io/_tables.py
```diff
@@ -115,6 +127,23 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+                kwargs.update(pd_kwargs)
+        elif type(dtype) == dict:
+            kwargs['dtype'] = dtype
```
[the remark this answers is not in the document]
gdj0nes: That equality check is good to know

I'm not sure what you mean by allow string types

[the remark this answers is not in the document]
stephen-hoover: I mean that valid inputs include both `float` and `"float"`.

civis/io/_tables.py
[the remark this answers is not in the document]
stephen-hoover: I'm not comfortable altering this default, especially since it would only apply to a special subset of circumstances and it's hidden from users. Is there a specific problem you're solving with this? Users have the option to modify the "engine" parameter themselves if they so choose.

civis/io/_tables.py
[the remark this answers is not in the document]
stephen-hoover: You mean "because it is more robust"? Can you give an example of a problem that this solves?

civis/io/_tables.py
```diff
@@ -224,7 +252,17 @@ def f(x):
                                .format(script_id))
     url = outputs[0]["path"]
     if use_pandas:
+        user_dtypes = kwargs.pop('dtype')
+        # Only for types for str and bool
+        kwargs['dtype'] = {k:v for k,v in user_dtypes.items()
+                           if v in [np.str, np.bool]}
```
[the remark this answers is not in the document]
gdj0nes: In addition to the subclass should it search for "int" provided as a string?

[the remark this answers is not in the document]
stephen-hoover: No, `np.issubdtype` will handle that for you!

civis/io/_tables.py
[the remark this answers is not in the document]
gdj0nes: I remember having to change to it but I can't remember why I had to make this change. This is not the first time I have encountered this problem either. If we encounter the error that is raised when the C engine fails it would probably be nice to raise a flag that the user try the python engine

civis/io/_tables.py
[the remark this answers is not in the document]
gdj0nes: Learning so much here

civis/io/_tables.py
```diff
@@ -224,7 +252,17 @@ def f(x):
...
         data = pd.read_csv(url, **kwargs)
+        dtype_cols = [c for c in list(data) if c in set(user_dtypes.keys())]
+        for col in dtype_cols:
+            are_nulls = data[col].isnull().any()
+            if not are_nulls:
+                data[col] = data[col].astype(user_dtypes[col])
```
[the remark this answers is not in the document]
gdj0nes: Good point, I'll drop it

civis/io/_tables.py
```diff
@@ -32,11 +39,41 @@
...
+    'bigint': np.int64,
+    'decimal': np.float64,
+    'numeric': np.float64,
+    'float': np.float64,
+    'float4': np.float32,
+    'float8': np.float64,
+    'real': np.float32,
+    'double precision': np.float64,
+    'boolean': np.bool,
+    'bool': np.bool,
```
beckermr: Doesn't redshift output 't' and 'f' for boolean values? http://docs.aws.amazon.com/redshift/latest/dg/r_Boolean_type.html

I think that means this would fail. I thought @stephen-hoover ran into this in CivisML?

beckermr: Yeah this won't work.

```python
In [1]: np.bool('t')
Out[1]: True

In [2]: np.bool('f')
Out[2]: True
```

beckermr:
@gdj0nes @stephen-hoover Can we get this into the 2.0 release?

I talked with @gdj0nes about this today and we agreed that the 2.0 release is a good time since it might be a big change.

beckermr:
Right. My thought was that if people call `read_civi` with no column arguments, then we infer the types for that case. Anything is too hard I agree. This might too weird of a special case to have wrapped up in the client I agree. 

We could add a function to return table types that are numpy compatible. I now cargo cult this around everywhere.

stephen-hoover:
Scanning through this PR again -- I like the approach of forcing users to opt-in to this with an explicit call to the `dtypes` parameter. I think there is a version of this PR which would work. I'd have to read through in more detail to say how close this is. If someone were to take it up again, it looks like there's still an outstanding issue with handling booleans, and it needs to be rebased on the current master.

civis/io/_tables.py
```diff
@@ -115,6 +156,20 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+            table_id = client.get_table_id(table=table, database=database)
+            py_types, date_cols = _redshift_to_py(client, table_id)
+            pd_kwargs = {'parse_dates': date_cols,
+                         'dtype': py_types,
+                         'engine': 'python',
+                         'infer_datetime_format': True,
+                         'encoding': 'utf8'}
+            kwargs.update(pd_kwargs)
+        else:
+            kwargs['dtype'] = dtype
```
keithing: Will this work if the user passes `columns`?  I think it will at least break if the users passes sql into the columns.

beckermr: It would not work with columns in general no.

beckermr:
I’m ok with optin and only on read_civis. I say we just let people do what they want wrt columns.

beckermr:
Alright. I am looking at this and thinking it is not worth it. We could add a function to the client to extract table types but given how we do exports this is not going to work. 

If we start to use parquet or some other typed format on the backend, this would solve this issue and speed up the export etc. 

I vote we close this.

beckermr:
Alright. Closing. Feel free to reopen if you feel differently OFC!

PR #146, opened 2017-09-21
[CIVP-12717] ENH add remote backend capability

beckermr:
This PR adds the ability for containers launched via the joblib backend to themselves launch more containers.


beckermr:
Tagging you for 30% CR here. I need to add tests, but I am going to try and take this for a test drive first.

stephen-hoover:
I think this looks like the right approach.

beckermr:
OK @stephen-hoover This is ready for another look. Some notes.

1. The tests for the new stuff are pretty slim. 
2. The backend now pickles(!), and this should be compatible with any changes in the executor that powers it.
3. I have tested both nested calling into the civis backend (`remote_backend='civis'`) and a nested call where the second call was set to sequential (`remote_backend='sequential'`). Both versions work great!

civis/parallel.py
```diff
@@ -680,25 +726,44 @@ def __init__(self, setup_cmd=_DEFAULT_SETUP_CMD,
                  from_template_id=None,
                  max_submit_retries=0,
                  client=None,
+                 remote_backend='civis',
                  **executor_kwargs):
-        if max_submit_retries < 0:
+        self.setup_cmd = setup_cmd
+        self.from_template_id = from_template_id
+        self.max_submit_retries = max_submit_retries
+        self.client = client
```
stephen-hoover: What's the motivation for changing this from a private attribute to a public one?

civis/run_joblib_func.py
```diff
@@ -37,9 +46,20 @@ def worker_func(func_file_id):
...
     try:
-        func = _robust_pickle_download(func_file_id, client=client,
-                                       n_retries=5, delay=0.5)
-        result = func()
+        func, remote_backend = _robust_pickle_download(
+            func_file_id, client=client, n_retries=5, delay=0.5)
+
+        _backend = _setup_remote_backend(remote_backend)
+
+        # graceful nested context managers are ~hard across python versions,
```
stephen-hoover: There should be a comment here about why we need nested context managers.

civis/parallel.py
beckermr: To make things pickle. There is already public client attached in other places. Thus I just kept it public here so that when I go remove it for pickling I don't have to check both places.

beckermr:
@stephen-hoover I think the comments I added should do it. Back to you!

civis/parallel.py
stephen-hoover: Are you saying we were attaching both `client` and `_client` to this object?

stephen-hoover: That makes sense. I think there's no reason this should be public, but we do have it as a public attribute on other objects. Consistency is good.

PR #150, opened 2017-09-29
[DE-15] Multipart uploads with Civis client

jkulzick:
Update file_to_civis to use multipart uploads
Add retry decorator to file_to_civis
Add new function civisfile_to_civis to import an existing file from files endpoint
Update csv_to_civis and dataframe_to_civis to use file_to_civis and civisfile_to_civis instead of _import_bytes

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -2,6 +2,7 @@ from collections import OrderedDict
 import io
 import json
 import logging
+import math
 import os
 import re
 import six
@@ -13,6 +14,7 @@ from civis import APIClient, find_one
 from civis.base import CivisAPIError, EmptyResultError
 from civis.compat import FileNotFoundError
 from civis.utils._deprecation import deprecate_param
+from civis.utils._jobs import retry
 try:
     from requests_toolbelt.multipart.encoder import MultipartEncoder
     HAS_TOOLBELT = True
@@ -24,12 +26,19 @@ try:
 except ImportError:
     HAS_PANDAS = False
 
+MIN_PART_SIZE = 5 * 2 ** 20 # 5MB
+MAX_FILE_SIZE = 5 * 2 ** 40 # 5TB
+MAX_PART_SIZE = 5 * 2 ** 30 # 5GB
+RETRY_EXCEPTIONS = (requests.HTTPError, requests.ConnectionError, requests.ConnectTimeout)
+
 log = logging.getLogger(__name__)
 __all__ = ['file_to_civis', 'civis_to_file', 'file_id_from_run_output',
            'file_to_dataframe', 'file_to_json']
 
 
 def _get_aws_error_message(response):
+    # Amazon gives back informative error messages
+    # http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
     # NOTE: This is cribbed from response.raise_for_status with AWS
     # message appended
     msg = ''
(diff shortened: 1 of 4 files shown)


civis/io/_files.py
```diff
@@ -49,6 +60,135 @@ def _get_aws_error_message(response):
...
+            pass
+        else:
+            return os.fstat(fileno).st_size
+
+    if hasattr(buf, 'getvalue'):
+        # e.g. BytesIO, cStringIO.StringIO
+        return len(buf.getvalue())
+
+    log.warning('Could not determine length of file. Defaulting to single put '
+                'instead of multipart upload. If file is >5GB put will fail.')
```
keithing: very minor, but putting this log message here is coupling this function to where it is being used.  I think it would be cleaner to call this warning from `file_to_civis`.  i.e.

```python
file_size = _buf_len(...)
if not file_size:
    log.warning(...)
```

civis/io/_tables.py
```diff
@@ -629,11 +641,133 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+        advanced_options=options)
+
+    run = client.jobs.post_runs(import_job.id)
+    fut = CivisFuture(client.jobs.get_runs,
+                      (import_job.id, run['id']),
+                      polling_interval=polling_interval,
+                      client=client,
+                      poll_on_creation=False)
+
+    if archive:
```
keithing: We've deprecated the `archive` pattern in favor of `hidden`.

civis/io/_tables.py
```diff
@@ -629,11 +641,133 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+                                 sortkey1=sortkey1, sortkey2=sortkey2,
+                                 delimiter=delimiter, headers=headers,
+                                 credential_id=credential_id,
+                                 polling_interval=polling_interval,
+                                 archive=archive, hidden=hidden)
+    return fut
+
+
+@deprecate_param('v2.0.0', 'api_key')
+def civisfile_to_civis(file_id, database, table, api_key=None, client=None,
```
keithing: The name for this function is a little awkward, do you have any other ideas for what to call this?

keithing: Here are some ideas: 

* `civis_file_to_table`,
* `civis_file_to_database`
* `remote_file_to_civis`

I can see how `civisfile_to_civis` follows the pattern we've been using, but I think it looks strange to have a function like `civis.io.civisfile_to_civis`...

jkulzick: Yeah, totally agree. I was trying to stick with the current naming convention, but it is confusing. I like civis_file_to_table.

civis/io/_files.py
jkulzick: +1

civis/io/_tables.py
jkulzick: +1

jkulzick:
@keithing 

Ready for review!

civis/io/_files.py
```diff
@@ -89,52 +223,43 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
...
-    # http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-post-example.html
-    # Note that the payload must have "key" first and "file" last.
-    form = file_response.upload_fields
-    form_key = OrderedDict(key=form.pop('key'))
-    form_key.update(form)
-    form_key['file'] = buf
+    file_size = _buf_len(buf)
+    if not file_size:
+        log.warning('Could not determine file size; defaulting to '
+                    'single post. Files over 5GB will fail.')
```
keithing: Will the typical use case

```python
with open("my_file.csv", "r") as f:
    file_to_civis(f, "my_file")
```

throw a warning?

jkulzick: it should not. this would only happen in rare cases where someone was using a file-like object that does not have a len attribute or support os.fstat.

An example of one scenario is in the test_io in the test_zip_member_to_civis test function, where someone opened a zip file and opened an archive member in memory instead of extracted it to disk. The zipfile class implementation does not support os.fstat on archive members.

civis/io/_tables.py
```diff
@@ -629,11 +643,122 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+                                  sortkey1=sortkey1, sortkey2=sortkey2,
+                                  delimiter=delimiter, headers=headers,
+                                  credential_id=credential_id,
+                                  polling_interval=polling_interval,
+                                  hidden=hidden)
+    return fut
+
+
+@deprecate_param('v2.0.0', 'api_key')
+def civis_file_to_table(file_id, database, table, api_key=None, client=None,
```
keithing: Do we need to include the deprecated `api_key` here?

jkulzick: removed!

jkulzick:
updated readme to remove references to requests-toolbelt

PR #151, opened 2017-10-05
[CIVP-12839] Fixing name parsing of multiword endpoints

keithing:
Certain endpoints with multi-word names like "match_targets" were being parsed as a single word.  This changes calls like

```python
client = civis.APIClient()
client.matchtargets.list()
```

To

```python
client = civis.APIClient()
client.match_targets.list()
```

Todo:
- [x] fix parsing
- [x] update tests
- [x] update change log with changed endpoint names

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -538,11 +538,11 @@ def generate_classes_maybe_cached(cache, api_key, api_version, resources):
             raise ValueError(msg.format(type(cache)))
         spec = JsonRef.replace_refs(raw_spec)
         classes = parse_api_spec(spec, api_version, resources)
-    classes = add_no_underscore_compatibility(classes)
-    return classes
+    classes_ = _add_no_underscore_compatibility(classes)
+    return classes_
 
 
-def add_no_underscore_compatibility(classes):
+def _add_no_underscore_compatibility(classes):
     """ Add class names without underscores for compatibility.
 
     Previously, no resources had underscores in APIClient.  Parsing of
@@ -553,11 +553,12 @@ def add_no_underscore_compatibility(classes):
     be removed in v2.0.0.
     """
     new = ["bocce_clusters", "match_targets", "remote_hosts", "feature_flags"]
+    classes_ = {}
     class_names = list(classes.keys())
     for class_name in class_names:
+        if class_name != "feature_flags":
+            classes_[class_name] = classes[class_name]
         if class_name in new:
             class_name_no_us = "".join(class_name.split("_"))
-            classes[class_name_no_us] = classes[class_name]
-        if class_name == "feature_flags":
-            classes.pop(class_name)
-    return classes
+            classes_[class_name_no_us] = classes[class_name]
+    return classes_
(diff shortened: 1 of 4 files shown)


civis/resources/_resources.py
```diff
@@ -433,12 +432,12 @@ def parse_api_spec(api_spec, api_version, resources):
...
     classes = {}
     for path, ops in paths.items():
-        class_name, methods = parse_path(path, ops, api_version, resources)
-        class_name_lower = class_name.lower()
-        if methods and classes.get(class_name_lower) is None:
-            classes[class_name_lower] = type(str(class_name), (Endpoint,), {})
+        base_path, methods = parse_path(path, ops, api_version, resources)
+        class_name = to_camelcase(base_path)
+        if methods and classes.get(base_path) is None:
+            classes[base_path] = type(str(class_name), (Endpoint,), {})
```
stephen-hoover: Should this use `base_path.lower()` instead of `base_path`? Or are we guaranteed that the API endpoint name will always be lower case?

keithing: I think `base_path.lower()` is better.

keithing:
@stephen-hoover, FYI I rebased and made a few changes to ensure that we now have *both* `APIClient.bocceclusters` and `APIClient.bocce_clusters`.  This ensures we have no breaking changes.  It also means that `APIClient.featureflags` can be used to access FeatureFlags, despite the name collision.

civis/resources/_resources.py
```diff
@@ -539,4 +538,26 @@ def generate_classes_maybe_cached(cache, api_key, api_version, resources):
             raise ValueError(msg.format(type(cache)))
         spec = JsonRef.replace_refs(raw_spec)
         classes = parse_api_spec(spec, api_version, resources)
+    classes = add_no_underscore_compatibility(classes)
+    return classes
+
+
+def add_no_underscore_compatibility(classes):
```
stephen-hoover: I suggest that you make this function private. Also, the function both modifies its input and returns a value. I suggest that it do one or the other -- either it modifies the input `classes`, or it leaves `classes` unchanged and returns a new dictionary. Since the function's name starts with "add", I think it would make sense to modify the input.

keithing: > I suggest that you make this function private. 

Agreed.  Some of the other private functions here are relying on the fact that the module is prefixed with an underscore (`_resources.py`) and aren't listed in the `__all__` of `__init__.py`.  But now that you mention it, it does seem to be a good idea to be explicit.

> Since the function's name starts with "add", I think it would make sense to modify the input.

Agreed. We can't avoid modifying the input the because `feature_flag` needs to be removed.

stephen-hoover: It should be possible to avoid modifying the input, no? You could make a copy of the input dictionary and only modify the copy.

keithing: Yeah, a copy is a good idea.

stephen-hoover:
The change looks good, but there's a test failure where the function name change needs to be propagated to the test module.

PR #145, opened 2017-09-20
DOC: Fix markup

jseabold:
@jmemich Mind taking a look at this?

PR #154, opened 2017-10-13
Release bugfix 1.6.3

keithing:
@stephen-hoover, there is a lot of heavy stuff planned for `v1.7.0` and I was asked to release some of the fixes that have made it into `master`.  I'd like to release a bugfix version before we make it to `v1.7.0`.


stephen-hoover:
We should also merge #145 before making this release.

stephen-hoover:
#158  didn't put documentation in the changelog. I think you should do that here; it's worth mentioning to people.

keithing:
> If we have reason to make a release now, this should be 1.7.0.

Agreed.  #150 will likely merge before this as well.  Is there is any reason to hold back the release of 1.7.0?  @leanne73 was looking for specifically the endpoint name fix.

keithing:
Closing this as the plan now is to move to a release of `v1.7.0` (although possibly not immediately).

PR #155, opened 2017-10-13
added "civis notebooks download" CLI command

mheilman:
This makes it easier to download a notebook to a local file.

### DIFF
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -7,7 +7,9 @@ Additional commands to add to the CLI beyond the OpenAPI spec.
 import os
 
 import click
+import requests
 
+import civis
 from civis.io import file_to_civis, civis_to_file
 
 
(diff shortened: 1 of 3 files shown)


mheilman:
I need to update the cached spec and possibly the cassettes.

civis/cli/_cli_commands.py
```diff
@@ -62,11 +64,27 @@ def files_upload_cmd(path, name, expires_at):
...
-    """Download a Civis File to the specified local path."""
+    """Download a Civis File to a specified local path."""
     with open(path, 'wb') as f:
         civis_to_file(file_id, f)
 
 
+@click.command('download')
+@click.argument('notebook_id', type=int)
+@click.argument('path')
+def notebooks_download_cmd(notebook_id, path=None):
```
stephen-hoover: It looks like `path` should be required here.

PR #156, opened 2017-10-13
update API client specification

mheilman:
This adds all publicly available resources to the list of base resources in the API client, for ease of use.


mheilman:
I need to update the cached spec and possibly the cassettes.

mheilman:
@stephen-hoover or @keithing, would one of you please take a look at this?

keithing:
> Do you remember what happens to endpoints with a hyphen? Does that get converted to an underscore, or removed entirely?

Hyphens might give us trouble, I'm going to write a test to verify that.  I think converting to an underscore makes the most sense, do you agree @stephen-hoover ?

stephen-hoover:
Yes, that does make sense. I think we're already doing that at https://github.com/civisanalytics/civis-python/blob/30d4591c31ccfd41eabd2aa231d6deb1551d15d4/civis/resources/_resources.py#L372 . I took a quick look through the `test_resources` module and didn't see a test for hyphenated endpoint names there. It would be worth adding one.

civis/resources/_resources.py
```diff
@@ -19,9 +19,41 @@
...
+    'credentials',
+    'databases',
+    'endpoints',
+    'enhancements',
+    'exports',
+    'files',
+    'groups',
+    'imports',
+    'jobs',
+    'match-targets',
```
keithing: I'm seeing `match_targets` in my endpoints list at `https://api.civisanalytics.com/endpoints`.  Is that the same for you?

mheilman: yeah, these should use underscores not hyphens.  The CLI-related test should convert them to hyphens (that's what the CLI does to follow UNIX-y norms).  I'll fix this.

mheilman:
Do we need a test for whether the `APIClient` has the expected resources given the API spec that's in the repo?  Should I add that in this PR?  I think I've addressed the other comments.

stephen-hoover:
I think it would be worthwhile to add a test which checks that all of the base resources are in the test API spec. That way we won't miss testing any new resources we add.

keithing:
> Do we need a test for whether the APIClient has the expected resources given the API spec that's in the repo? Should I add that in this PR? I think I've addressed the other comments.

I think what we should do going forward is to remove `BASE_RESOURCES` and default to giving users all the endpoints they have access to.  We then use the cached API spec only for tests / documentation.  That way, us changing the cached api spec should never cause changes in behavior for the user.

stephen-hoover:
Oh, good point. The test is less useful then. LGTM!

PR #160, opened 2017-10-20
[DE-23] Parallel Unload with Civis Client

jkulzick:
This implements the following:

civis_to_csv and read_civis_sql
- parallel unload by retrieving headers separately
- always requesting gzip and decompressing it as a stream using zlib, if needed
- serves back the correct file compression, i.e. none/gzip/zip (for civis_to_csv only)

civis_to_file
- shutil.copyfileobj instead of iterating of chunks; slightly better memory usage


civis/io/_tables.py
```diff
@@ -782,18 +830,76 @@ def _get_sql_select(table, columns=None):
...
+        with open(local_path, 'wb') as fout:
+            fout.write(headers)
+            _decompress_stream(url, fout)
+
+        return
+
+    # if headers are requested or zip compression is requested
+    # first decompress the stream and write headers
+    # then re-compress the file contents
+    with tempfile.NamedTemporaryFile() as temp:
```
keithing: We had trouble with `NamedTemporaryFile` for Windows users before.  Is there a way to do this without using that function?

https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

keithing:
Meta Question: what sort of functionality here is overlapping with `boto3`?  I don't know if we want to add that dependency, but I'm curious your thoughts about that.

civis/io/_tables.py
jkulzick: @keithing yikes! darn windows...

at the bottom of the page it outlines a solution, which is to pass delete=False and then us os.unlink after you're done with the file. still feels more straightforward than tempfile.mkstemp().

also, solution is repeated on SO [https://stackoverflow.com/questions/94153/how-do-i-persist-to-disk-a-temporary-file-using-python](here)

thoughts?

stephen-hoover: You could use a temporary directory instead. We do that in several other places in the client. You need to import the version from `compat`; the `TemporaryDirectory` isn't in Python 2.

jkulzick:
good question. since we use pre-signed urls we can't use boto3. otherwise, we would be able to establish a connection with boto3 and not do some of this and also we wouldn't have needed to implement the multipart upload PR, but alas... we would all need s3 credentials with appropriate permissions for that.

civis/io/_tables.py
jkulzick: great. that looks better.

also, i think i need to fix the use of `TemporaryDirectory` in the multipart-upload PR then
https://github.com/civisanalytics/civis-python/blob/master/civis/io/_files.py#L163

keithing: Is there a way to avoid the second pass over the data in the first place?

mheilman:
> since we use pre-signed urls we can't use boto3

Bummer.  Probably related: https://github.com/boto/boto3/issues/798

Maybe it's worth adding a link to that issue or something else describing why we can't use boto, so that in case boto ever makes this possible, we'll remember why we didn't use boto.

civis/io/_tables.py
```diff
@@ -209,8 +215,18 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
     if archive:
         warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
                       "Use `hidden` instead.", FutureWarning)
+
+    if isinstance(database, str):
+        database = client.get_database_id(database)
+
+    credential_id = credential_id or client.default_credential
+    headers = _get_headers(client, sql, database, credential_id)
```
keithing: @jmemich,  if I recall correctly you were able to return headers in `civis_to_multifile_csv`.  Do you know how that was implemented?  I'd like to avoid an extra query to retrieve headers if there is some optimization I am unaware of.

civis/io/_tables.py
jkulzick: not really... parallel unload will only return no compression, gzip or bzip2.

if we want to be able to return gzip or zip files then we'll have to download the file while decompressing the stream, write headers, then zip/gzip it.

the issue is that we're including headers. i'm not sure how to get around that.

civis/io/_tables.py
jkulzick: `civis_to_multifile_csv` uses `force_multifile=True` and returns the json from the manifest file. Civis API writes the headers to the manifest, but only when using the force_multifile option. headers are not provided by the Civis API for other export options.

the challenge with this approach is that it also doesn't combine files so you get back multiple files. we could download those in parallel using threads and then recombine them locally to still return one file to the user.

civis/io/_tables.py
stephen-hoover: You can concatenate gzipped files into a larger gzipped file.

civis/io/_tables.py
keithing: It's possible that there is a side-effect in the query a user passes to `read_civis_sql`.  I.e. a user might pass a query like

```sql
INSERT INTO table VALUES ...;
SELECT * FROM table
```

And the user wouldn't expect that query to run twice (which would happen if we ran a query to get the headers).  In the R client we add `BEGIN READ ONLY` to `read_civis` calls, which we could do here but it would constitute a breaking change.

It's awkward, but I think we need to wrap calls here in `BEGIN READ ONLY`, catch exceptions when the query isn't read only and unload with `PARALLEL OFF` (warning the user) in order to get the headers.

civis/io/_tables.py
```diff
@@ -782,18 +830,76 @@ def _get_sql_select(table, columns=None):
     return sql
 
 
-def _download_file(url, local_path):
-    response = requests.get(url, stream=True)
-    response.raise_for_status()
+def _get_headers(client, sql, database, credential_id):
+    sql = 'select * from ({}) limit 1'.format(sql)
```
keithing: If we have a complex query, it is possible it will be slower to run the query twice than unload with `PARALLEL OFF` ?

jkulzick: limit 1 should make it very fast

stephen-hoover: If there's an order by, it could still be very slow, no?

civis/io/_tables.py
stephen-hoover: Modifying queries has implications for table caching. The read-only queries won't use the same cache as the unmodified queries. That doesn't seem like a big deal here, but it seems worth pointing out.

keithing: > Modifying queries has implications for table caching.

That's a bummer.  I don't see how we can get around this though, given the problem of possible side effects.  All ideas are welcome on this!

civis/io/_tables.py
jkulzick: agreed. i'll check for order bys and we'll do a regular export. platform will not do a parallel unload for these anyway so there is no reason for the client to retrieve headers separately

civis/io/_tables.py
jkulzick: we can wrap the query for getting headers in `begin read only;` and if that fails we will proceed with a regular export. since the query for getting headers has a limit 1 it will be returning different results anyway so i don't think we need to worry about table caching here.

stephen-hoover: Oh, right -- I was thinking that the "BEGIN READ ONLY" would be in the regular query as well, but it would only be to get the headers, wouldn't it?

civis/io/_tables.py
jkulzick: - removed tempfile and switched to civis.compat.TemporaryDirectory
- simplified code for gzips by just writing headers as gzipped bytes since gzipped files can be concatenated

civis/io/_tables.py
jkulzick: - added 'begin read only;' to _get_headers function
- added separate function _include_header to determine if we should request headers

jkulzick:
@keithing updated per comments. let me know if you think i should add some additional tests since we have some branching now in the code.

jkulzick:
@keithing @stephen-hoover 

I want to bring the following to your attention since it could impact the current experience. Please let me know what you think.

Note: I have removed use of shutil.copyfileobj from civis.io.civis_to_file for now since it uses response.raw and because of the below discussion

- When Console executes a parallel unload with gzip, files are created with content-type application/x-gzip.

- When Console cannot execute a parallel unload, the CsvWriter generate the files and upload them to S3. When specifying gzip, the files are created with content-type text/<something> depending on the delimiter and more important content-encoding gzip.

- requests.iter_content will decompress the file assuming stream=True and content-encoding appears in the headers.

- based on the above, the use of iter_content will produce different results

- the civis_to_csv function uses iter_content and will never return a gzip file. is this by design? a bug? a breaking change?

Based on the above, I explored the following options. I've gone with #1 for now but we should probably go with #3 if we don't consider the last point above a bug

1. Use requests.raw
- this allows us to use the same _decompress_stream function for both scenarios since we're getting the raw response

2. Check content-type in headers
- by doing this would could use native functionality of iter_content to decompress the stream when content-encoding is in place
- this would require a separate _download_file function since this changes how you generate files

3. Check content-type in headers + Never return gzip
- similar to 2 but the flow would be simplified a bit since we will never write gzip files

civis/io/_tables.py
```diff
@@ -209,8 +217,24 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
     if archive:
         warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
                       "Use `hidden` instead.", FutureWarning)
+
+    if isinstance(database, str):
+        database = client.get_database_id(database)
```
keithing: `client.get_database_id` should take care of handling whether an int or str is passed [1].  Is there a specific reason to check `isinstance` here?

https://github.com/civisanalytics/civis-python/blob/master/civis/civis.py#L45

jkulzick:
removed isinstance call for setting database per feedback

last open item seems to be what to do about gzip compression

jkulzick:
I update `civis.io.civis_to_csv` [here](https://github.com/civisanalytics/civis-python/pull/160/files#diff-afeca7e651e5cf116e70dcbfaff99364R343) to reset gzip compression to none if requested

jkulzick:
@keithing ready to go!

stephen-hoover:
I have some comments I'll get to you soon.

civis/io/_tables.py
```diff
@@ -781,18 +845,85 @@ def _get_sql_select(table, columns=None):
...
+        # if _get_headers throws an error then assume sql is not read only
+        headers = _get_headers(client, sql, database, credential_id)
+        include_header = False
+    except:  # NOQA
+        pass
+
+    return include_header, headers
+
+
+def _decompress_stream(response, buf, bytes=True):
```
stephen-hoover: `bytes` is a Python built-in type name. Could you change this to maybe "write_bytes"?

civis/io/_tables.py
```diff
@@ -807,6 +938,8 @@ def callback(future):
             return
         else:
             url = outputs[0]["path"]
-            return _download_file(url, filename)
+            file_id = outputs[0]["file_id"]
+            log.info('Exported results to Civis file %s', file_id)
```
stephen-hoover: I think this should be a debug-level emit.

civis/io/_tables.py
```diff
@@ -209,8 +217,22 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
     if archive:
         warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
                       "Use `hidden` instead.", FutureWarning)
+
+    database = client.get_database_id(database)
```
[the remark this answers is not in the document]
jkulzick: done

civis/io/_tables.py
```diff
@@ -303,12 +337,38 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
...
         client = APIClient(api_key=api_key, resources='all')
 
+    database = client.get_database_id(database)
+    credential_id = credential_id or client.default_credential
+
+    # don't fix bug that would cause breaking change for now
+    # when gzip compression is requested, a gzip file is not actually returned
+    # instead the gzip file is decompressed during download
+    if compression == 'gzip':
+        compression = 'none'
```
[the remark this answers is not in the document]
keithing: I like the idea of a deprecation warning.  I thought this might be noisy in logs, but if you're pro deprecation warning I like it.

civis/io/_tables.py
```diff
@@ -226,13 +248,25 @@ def f(x):
     if not outputs:
         raise EmptyResultError("Query {} returned no output."
                                .format(script_id))
+
     url = outputs[0]["path"]
+    file_id = outputs[0]["file_id"]
+    log.info('Exported results to Civis file %s (%s)',
+             outputs[0]["output_name"], file_id)
```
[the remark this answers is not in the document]
jkulzick: i guess that spirit of logging as info was to make the file created more discoverable so that people could interact with the file more easily if they wanted to.

in order to do that they would have to submit this job as hidden=False, then find the script id, then run id, then look at run outputs.

civis/io/_tables.py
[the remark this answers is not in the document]
stephen-hoover: Could there be some way to use the non-deprecated behavior? I don't think we should have a warning if there's no way to turn it off.

civis/io/_tables.py
```diff
@@ -226,13 +248,25 @@ def f(x):
...
+
     if use_pandas:
-        data = pd.read_csv(url, **kwargs)
+        data = pd.read_csv(url, names=headers, compression='gzip', **kwargs)
```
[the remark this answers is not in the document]
jkulzick: great tip. thanks!

civis/io/_tables.py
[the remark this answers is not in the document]
keithing: Do you mean to turn off the warning other than set `civis_to_csv(..., compression='none')`  (the default)?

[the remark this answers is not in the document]
jkulzick: we should probably warn users that compression=gzip should be set to compression=none if they do not want a gzip file returned. what's the best way to implement that?

civis/io/_tables.py
```diff
@@ -760,16 +821,19 @@ def civis_file_to_table(file_id, database, table, client=None,
 def _sql_script(client, sql, database, job_name, credential_id, hidden=False,
                 csv_settings=None):
     job_name = maybe_get_random_name(job_name)
-    db_id = client.get_database_id(database)
-    cred_id = credential_id or client.default_credential
+    database = client.get_database_id(database)
```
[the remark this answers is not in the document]
jkulzick: done

civis/io/_tables.py
jkulzick: done

civis/io/_tables.py
```diff
@@ -781,18 +845,85 @@ def _get_sql_select(table, columns=None):
...
+    # can't do a parallel unload when sql contains an order by
+    if not include_header or re.search('order\s+by', sql, re.I | re.M):
+        return include_header, headers
+
+    try:
+        # if _get_headers throws an error then assume sql is not read only
+        headers = _get_headers(client, sql, database, credential_id)
+        include_header = False
+    except:  # NOQA
+        pass
```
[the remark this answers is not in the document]
jkulzick: done

civis/io/_tables.py
```diff
@@ -303,12 +337,38 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
...
+
+    # determine if we can request headers separately; if we can then Platform
+    # will perform a parallel unload which is significantly more performant
+    ovrd_include_header, headers = _include_header(client, sql, include_header,
+                                                   database, credential_id)
+
+    # format headers so we can write them to the csv
+    if headers:
+        if not unquoted:
+            headers = ['"{}"'.format(x) for x in headers]
```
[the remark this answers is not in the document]
jkulzick: I will double check out headers like this are returned

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: please confirm that debug is preferred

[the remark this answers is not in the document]
stephen-hoover: Hm. I don't like the idea of making a file ID an "output" of this function. For a couple of reasons -- one is that it feels a bit like an implementation detail that the export becomes a Civis file, and we shouldn't commit to having this function always provide a file. The second is that printing something to the screen doesn't seem like a great way to surface information which we want the user to pick up and use somewhere else.

Given that Python offers very fine grained control of log outputs, I guess it's not that big of a deal. But I'd advocate for having all (or almost all) log emits from the client as debug level.

@keithing , do you have a preference here?

jkulzick:
@stephen-hoover 

- I updated the docstrings with a note. Let me know what you think
- If a user requests zip compression and we don't manage it then the API will not perform a parallel unload so requesting gzip, decompressing the gzip stream and recompressing as zip is the only way. FWIW, the performance will be at least 4x faster with parallel unload.

If we wanted to get complex I could first execute the query as an explain, then parse the explain to determine the number of rows and columns expected to be returned and then use the normal method if that was below some threshold.

- I updated the logging to debug level

Aside from the ^^, the only outstanding item is whether or not we should issue a deprecation warning.

civis/io/_tables.py
jkulzick: done

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: done

civis/io/_tables.py
```diff
@@ -303,12 +337,38 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
     if client is None:
         client = APIClient(api_key=api_key, resources='all')
 
+    database = client.get_database_id(database)
```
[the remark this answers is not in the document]
jkulzick: done

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: updated to debug for now

jkulzick:
@stephen-hoover @keithing 

let me know what you think about the deprecation warning. it would be good to get through this today so that I can coordinate some user tests with Wade before the tentative release next week

stephen-hoover:
Changes look good! For now, how about we document the current and planned future behavior of `compression="gzip"` in the docstrings and the changelog?

jkulzick:
Sweet!

PR #161, opened 2017-10-23
DOC: Use real civis_file_to_table example

seancquinn:
We were using a csv_to_civis example in the civis_file_to_table definition Switching to a correct example.


seancquinn:
@jkulzick - quick doc fix for your new civis_file_to_table - mind doing a quick review?

jkulzick:
@seancquinn i'll add this change to [DE-108] Fix use of TemporaryDirectory in civis.io.file_to_civis. there is a CI error with the code i need to fix for the changes to go in

PR #162, opened 2017-10-23
[DE-108] Fix use of TemporaryDirectory in civis.io.file_to_civis

jkulzick:


civis/io/_files.py
```diff
@@ -160,14 +159,15 @@ def _upload_part(item):
 
     # upload each part
     try:
-        with tempfile.TemporaryDirectory() as tmp_dir:
+        with TemporaryDirectory() as tmp_dir:
+            tmp_path = os.path.join(tmp_dir, 'file_to_civis')
```
keithing: is it possible multiple calls might be made simultaneously?  Should we use a uuid instead of "file_to_civis" as the string here?

keithing: Or will `TemporaryDirectory` guaranteed to be a unique folder name?

keithing:
We talked about this offline, but this looks like it's breaking because of the (unrelated) bare Exception linter error.  Are you planning to add the `noqa` to this PR or a different one?

stephen-hoover:
I ran into the same linter error. I opened #164 to fix it.

jkulzick:
great. thanks!

stephen-hoover:
@jkulzick , if you merge the master branch into this PR, your tests should pass.

jkulzick:
@stephen-hoover has the api spec been updated under civis/tests/civis_api_spec.json?

i don't see endpoints in there for files/multipart/post or files/multipart/{id}/complete

stephen-hoover:
I haven't updated the spec. I didn't realize this PR was waiting on that. If that's the case, then I guess it's not ready yet.

stephen-hoover:
This PR doesn't look like it needs an updated API spec. Are you thinking of #160 ?

jkulzick:
This PR needs the updated spec b/c I fixed the test for multipart-upload, test_large_file_to_civis, so that is actually using the new code and the new multipart API endpoint.

As a result, that test now fails b/c of the spec

The reason we didn't catch this 2.7 compatibility issue was b/c the test_large_file_to_civis was not actually testing what I thought it was

stephen-hoover:
@jkulzick , the new API spec is in. You can continue with this PR now.

jkulzick:
@stephen-hoover +1 i'll update it today

jkulzick:
@keithing @stephen-hoover 

Ready for your review!

jkulzick:
fixed changelog

civis/io/_files.py
```diff
@@ -148,7 +147,7 @@ def _gen_chunks(part_buf, max_bytes, chunk_size=32 * 1024):
     def _upload_part(item):
         part_num, part_url = item[0], item[1]
         log.debug('Uploading file part %s', part_num)
-        file_out = os.path.join(tmp_dir, str(part_num) + '.csv')
+        file_out = tmp_path.format(i)
```
keithing: Does `tmp_path` need to be passed to this function? I'm not clear where it is defined.

jkulzick: the _upload_part func is inline so its within the same scope

keithing: So `tmp_path` is defined on L164, *after* this function is defined? I have feeling this might be confusing later.  Could we do this with a closure instead?  Something like

```python
def _create_upload_part(tmp_dir):
    def _upload_part(item):
        ...
    return _upload_part
```

stephen-hoover: Or you could make `tmp_path` an input to this function and use functools.partial to include it.

jkulzick: updated; implemented use of functools.partial

jkulzick:
removed CivisFutureTests cassettes per ^^

civis/io/_files.py
```diff
@@ -160,25 +160,28 @@ def _upload_part(item):
...
-            with Pool(MAX_THREADS) as pool:
-                pool.map(_upload_part, enumerate(urls))
+            _upload_part = partial(_upload_part_base, part_path=tmp_path)
+            pool.map(_upload_part, enumerate(urls))
 
     # complete the multipart upload; an abort will be triggered
     # if any part except the last failed to upload at least 5MB
     finally:
         client.files.post_multipart_complete(file_response.id)
+        pool.terminate()
```
stephen-hoover: Could you put the `pool.terminate()` as the first line of the `finally`? It seems less likely to error than the client call. Alternately, the pool creation and destruction could be in its own small try/finally block where the context manager used to be.

jkulzick: +1

moved

PR #165, opened 2017-10-25
BUG Don't error when cancelling completed ContainerFutures

stephen-hoover:
If you call `scripts.post_cancel` on a completed script, you get a `CivisAPIError` with a 404 status code. We'd gated this call behind an `if not self.done()`, but it's possible for jobs to complete before the done status updates. This is most prominent when using polling to update job status. This patch causes all 404 errors on `post_cancel` to return `False` for the cancelled status (since they obviously weren't cancelled).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Fixed parsing of multiword endpoints. Parsing no longer removes underscores
   in endpoint names.
+- In ``civis.futures.ContainerFuture``, return ``False`` when users attempt to cancel
+  an already-completed job. Previously, the object would sometimes give a ``CivisAPIError``
+  with a 404 status code. This fix affects the executors and joblib backend, which
+  use the ``ContainerFuture``.
 
 ### Added
 - ``civis.resources.cache_api_spec`` function to make it easier to record the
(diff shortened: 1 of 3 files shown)


keithing:
> it's possible for jobs to complete before the done status updates

Could you talk a bit more about this? Would it be possible to fix this rather than add an exception?

stephen-hoover:
No, that's not an error. If you're checking job status by polling at 30-second intervals, then you can imagine a situation where you check status at second 0 and see the job is running. The job completes at second 10. At second 20, the user requests a job cancellation. The `CivisFuture` considers itself to be still in state "running", so it sends a cancel request. The Civis API responds with a 404 error because the job has completed.

stephen-hoover:
This PR is narrowly calling out 404 status codes. Maybe it should catch all `CivisAPIError` exceptions?

keithing:
> This PR is narrowly calling out 404 status codes. Maybe it should catch all CivisAPIError exceptions?

I like that.  Maybe turn the exception into a warning?  Returning `False` seems fair to me, I don't have a strong opinion on that.

I think we could mitigate this problem if we forced a polling request at the time of cancellation.

stephen-hoover:
@keithing , I changed the `raise` to a warning.

I don't think it would help to force a polling call before trying to cancel. It would reduce the chance that we try to cancel a completed script, but not completely eliminate it. And I think it's reasonable to catch an exception and treat it as an answer to the question "did we cancel this job".

PR #166, opened 2017-10-26
DOC Add use case to parallel doc page

stephen-hoover:
The documentation on the Civis joblib backend assumed that users were already familiar with `joblib` and embarrasingly parallel computation, but not everyone is. Talk about the use case for the Civis joblib backend so that users can decide if this tool is useful for them or not.


stephen-hoover:
@johnm232 , could you review this documentation addition? Do you think this makes clear when users would want to consider using the Civis joblib backend?

stephen-hoover:
Right, "wall time" is more jargon than I should be using. I changed it to "time your job takes to finish". How's that, @johnm232 ?

PR #167, opened 2017-10-27
REF DRY polling thread restarts and always start after pubnub

stephen-hoover:
Refactor the Futures code so that we handle thread starts and resets from the same function. Now `_begin_tracking` is the one function which sets (or resets) both status polling in the thread and PubNub channel listening. Make sure that we always subscribe to PubNub before we poll. This is important because otherwise we might poll, see a running job, and miss that the job completed in between polling and subscribing.


stephen-hoover:
@keithing , are you able to review this?

civis/futures.py
```diff
@@ -120,9 +120,15 @@ def __init__(self, poller, poller_args,
...
-            self._pubnub = self._subscribe(config, channels)
+    def _begin_tracking(self):
+        # Be sure to subscribe to the PubNub channel before polling.
+        # Otherwise the job might complete after polling and before
+        # we subscribe, causing us to miss the notification.
+        with self._condition:
+            if hasattr(self.client, 'channels') and not self.subscribed:
+                config, channels = self._pubnub_config()
+                self._pubnub = self._subscribe(config, channels)
+            super()._begin_tracking()
```
mheilman: I think this would be a bit easier to read if there were a comment here like, "The superclass PollableResult manages polling."

stephen-hoover: Comment added.

stephen-hoover:
@mheilman , I had some trouble thinking of a test. The most relevant thing seems to be testing that the first poll happens after subscribing. When I tried to write that test, I realized that we had been subscribing before polling to begin with. We were setting up the polling thread before subscribing, but we only started the polling thread running the first time that someone asked for status. That can only happen after object creation.

This refactor changes the way thread starting is handled such that we always start the result thread (and therefore the polling) at the same time as we create the thread. So I think this PR doesn't solve any problems, but I still think the refactor is useful.

stephen-hoover:
@mheilman , as we discussed in person, I restored the performance optimization to only start threads when we need to find job status. Tests passing.

civis/polling.py
```diff
@@ -188,10 +195,14 @@ def cleanup(self):
...
         with self._condition:
-            if self._polling_thread.is_alive():
+            if (getattr(self, '_polling_thread', None) is not None and
+                    self._polling_thread.is_alive()):
                 self._polling_thread.cancel()
             self.polling_interval = polling_interval
             self._polling_thread = _ResultPollingThread(self._check_result, (),
                                                         polling_interval)
+            if start_thread:
+                self._polling_thread.start()
```
mheilman: Could/should this line just be added wherever we need to start the thread rather than having an extra kwarg?  This way seems a little difficult to read, but it's very possible I'm missing something

stephen-hoover: I added it here so that we centralize handling of the private attribute `_polling_thread`. There's only one place I'd need to add it (in the `ContainerFuture`) if it weren't here. Do you think that would be better?

mheilman: Hmm... I think either way is fine.

PR #171, opened 2017-11-02
 BUG SQL jobs should poll on creation

stephen-hoover:
We've set the `CivisFuture`'s `poll_on_creation` parameter to `False` when we know that we just started the jobs. This saves a Platform API call, which can matter when starting hundreds of jobs at once. However, it's possible for SQL to fail fast enough that the job has already completed by the time we can subscribe to the notification endpoint. If that happens, we miss the completion notice and need to wait for the fallback poll in another 9.5 minutes. Switch `CivisFuture` objects connected to SQL scripts to not skip the first poll. Continue to use `poll_on_creation=False` for imports and container scripts. Those should take longer to complete.


keithing:
When we subscribe, can we start the subscription as of 9 minutes ago (or some other time)?

stephen-hoover:
It looks like we can! I'm closing this in favor of #174 .

PR #173, opened 2017-11-03
PERF Don't keep an open session when we don't need to

### LINKED ISSUE #168: Connections Not Closed when No Client Passed to civis.ml.ModelPipeline
I've got some code like:

```python
for i in range(500):
    model = civis.ml.ModelPipeline(clf)
    model.train()
```

Running this on my system eventually results in errors due to running out of file handles (my system's default limit is 1024.)
Each call to model.train() results in three connections to the civis platform which are never closed, consuming file handles (you can observe this using `lsof -i` and `strace` on linux.)

This similar bit of code results in connections being closed and no file handle leaks:

```python
for i in range(500):
    client = civis.APIClient()
    model = civis.ml.ModelPipeline(clf, client=client)
    model.train()
```

I have vague suspicions about the first case creating its own APIClient object since one was not supplied, and its getting garbage-collected somehow prevents connections from being closed.

stephen-hoover:
We'd been holding open sockets in two places. This uses system file handles, which isn't a problem for a few things at a time, but which can eat up lots of resources if you're getting to hundreds of client objects and/or `CivisFuture`s.

First, we opened a `requests.Session` with every new client and kept it open. This wasn't getting closed even when the `APIClient` went out of scope. We can ensure that this session closes by only opening it when needed. There are performance implications here, but we send API requests infrequently enough that it's probably worth it to open a new session each time. I measured 0.1 ms to open and close a session in rough testing.

We're now storing the user's API key directly on the `APIClient` instance, so explicitly raise an error when pickling. This will ensure we don't accidentally leave keys where they shouldn't be.

The second place which keeps open sockets is PubNub, used in the `CivisFuture`. PubNub opens two sockets (I still don't know why there's two) when we subscribe, and it never gives them up, even after unsubscribing. PubNub does not provide a way to close the session in public methods, but it's possible to close the session directly on the private `_request_handler` attribute. I found that PubNub will then reopen one socket (not sure why), but I can avoid this by immediately removing the `Session` object. I then remove the `PubNub` object since it's no longer usable. (We would recreate it if we wanted to resubscribe anyway.) Note that `PubNub` is written with lots of circular references, so the Python garbage collector isn't going to be able to remove it. This represents a (small) memory leak.

Closes #168 .

TODO
- [x] Add changelog
- [x] Make sure that keys aren't serialized


stephen-hoover:
@waltaskew , here's my solution to your issue in #168 . Connections will still be held open for any `CivisFuture` which is waiting for its job to complete. What do you think?

Other possible solutions:

We could potentially re-use sockets for the pubnub connection. If the client object had a `PubNub` object which it passed to every new `CivisFuture`, we could use the same sockets for each of them. That way, 500 `CivisFuture` objects created from the same client would all use the same `PubNub` object and add their own Listeners as needed. I'd have to rely on an `APIClient` destructor to close that connection. I don't know how well that would work.

Another idea -- we encourage users to let us auto-create new `APIClient` objects as needed. We could create a global `APIClient` which is returned to each user, instead of creating a fresh one each time. Then we could keep an open Session for API calls and a shared `PubNub` object.

civis/base.py
```diff
@@ -93,8 +94,9 @@ def _make_request(self, method, path=None, params=None, data=None,
         url = self._build_path(path)
 
         with self._lock:
-            response = self._session.request(method, url, json=data,
-                                             params=params, **kwargs)
+            with open_session(**self._session_kwargs) as sess:
```
mheilman: we can't just use `requests.request` because the session [needs the httpadapter for retries](https://github.com/civisanalytics/civis-python/blob/0d9c886570a78bbbdb2f45a53c3ca660db6f89ce/civis/_utils.py#L58-L61), right?

waltaskew: I was wondering about this myself -- it looks like we're losing the performance benefits of sessions if we're opening and closing them with each requests, so I was wondering if we could stop using them entirely to avoid a little of the complexity they bring.

waltaskew:
> Another idea -- we encourage users to let us auto-create new APIClient objects as needed. We could create a global APIClient which is returned to each user, instead of creating a fresh one each time. Then we could keep an open Session for API calls and a shared PubNub object.

This might feel a bit un-appetizing, but we could also go the other way and require users to explicitly pass in APIClient objects in the spots  such as the ModelPipeline object where we currently automatically create them if one is not passed. I guess I'm assuming users would tend to create the minimum number of required APIClient objects if left to their own devices :)

waltaskew:
> Another idea -- we encourage users to let us auto-create new APIClient objects as needed. We could create a global APIClient which is returned to each user, instead of creating a fresh one each time. Then we could keep an open Session for API calls and a shared PubNub object.

This might feel a bit un-appetizing, but we could also go the other way and require users to explicitly pass in APIClient objects in the spots  such as the ModelPipeline object where we currently automatically create them if one is not passed. I guess I'm assuming users would tend to create the minimum number of required APIClient objects if left to their own devices :)

waltaskew:
>  PubNub opens two sockets (I still don't know why there's two) when we subscribe, and it never gives them up, even after unsubscribing. PubNub does not provide a way to close the session in public methods

Do you think this is something we might point out to folks upstream?

civis/base.py
waltaskew: We could also go the other direction and make the session a singleton so all requests would share the same connection

stephen-hoover: @mheilman , yes, but also, the `request` function has exactly two lines:
```python
    with sessions.Session() as session:
        return session.request(method=method, url=url, **kwargs)
```

It doesn't look like creating a new `Session` has much of a performance penalty. I don't think the Civis API supports querying at a rate high enough where it would matter.

civis/futures.py
```diff
@@ -141,6 +141,19 @@ def cleanup(self):
             if hasattr(self, '_pubnub'):
                 self._pubnub.unsubscribe_all()
 
+                # Pubnub doesn't close its open session, so that ourselves
```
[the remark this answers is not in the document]
stephen-hoover: Fixed.

Pubnub has issues turned off on their repo, so I couldn't file anything on GitHub. They ask for questions to be sent to customer support.

civis/base.py
waltaskew: > It doesn't look like creating a new `Session` has much of a performance penalty. I don't think the Civis API supports querying at a rate high enough where it would matter.

I meant to refer to the performance benefits of re-using session objects, not the cost of creating new session objects.

PR #175, opened 2017-11-08
DOC Mention pickling in ModelPipeline doc string

### LINKED ISSUE #169: document serialization of client variable in civis.ml.ModelPipeline
`civis.ml.ModelPipeline` has a `client` init argument that lets users specify an `APIClient` object.  This could be used, for example, to run with restricted permissions.  When `ModelPipeline` is serialized, however, the member variable `_client` is discarded, and a new one with the default API key is created during deserialization ([code](https://github.com/civisanalytics/civis-python/blob/c98ffbf788a88516e44d9cfd569ad74c4c15d2be/civis/ml/_model.py#L642-L649)). 

We probably don't want to serialize the API client objects since they have API keys (i.e., credentials), but it'd be helpful to document this behavior a little better so that users don't have wrong assumptions about what API keys will get used.

stephen-hoover:
Closes #169 .


stephen-hoover:
@elsander , do you think this clarifies the pickling behavior of the ModelPipeline?

stephen-hoover:
Good point. I added the same comment to the `ModelFuture`.

elsander:
I think this probably applies farther down the inheritance chain (ContainerFuture, PollableResult, etc). It's probably overkill to document in each of them, but are there any other user-facing classes in the inheritance chain where a note might be helpful?

stephen-hoover:
I think the `ModelFuture` is the only one with a `__getstate__` or `__setstate__` function, so we don't need a comment anywhere else.

PR #180, opened 2017-11-14
make it easier to upload local files

mheilman:
This facilitates uploads from or downloads to local files by allowing users to specify paths rather than file/buffer objects.  This particular type of polymorphism seems pretty well-accepted (cf. [pandas.read_csv](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html)).

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -192,8 +192,9 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
 
     Parameters
     ----------
-    buf : file-like object
-        The file or other buffer that you wish to upload.
+    buf : file-like object or str
+        The file or other buffer that you wish to upload. Strings will be
+        treated as paths to local files to open.
     name : str
         The name you wish to give the file.
     api_key : DEPRECATED str, optional
@@ -213,6 +214,8 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
 
     Examples
     --------
+    >>> # Upload file at a given path on the local filesystem.
+    >>> file_id = file_to_civis("my_data.csv", 'my_data')
     >>> # Upload file which expires in 30 days
     >>> with open("my_data.csv", "r") as f:
     ...     file_id = file_to_civis(f, 'my_data')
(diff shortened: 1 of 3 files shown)


civis/io/_files.py
```diff
@@ -235,6 +237,13 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
     Small or non-seekable file-like objects will be uploaded with a
     single post.
     """
+    if isinstance(buf, str):
+        with open(buf, 'rb') as f:
+            return _file_to_civis(
+                f, name, api_key=api_key, client=client, **kwargs)
```
keithing: What happens when `buf` is not a `str` here?

keithing:
I like this.  I think polymorphism is okay here.

stephen-hoover:
Our current read/write functions resist polymorphism, which is like the standard library functions (e.g. `pickle.load`). On the other hand, reading from files on disk is incredibly common, and third-party libraries I've seen (e.g. `joblib` and `pandas`) are polymorphic in exactly the way this PR proposes. I think this PR is a good idea. I know I've written this wrapper myself more than once.

mheilman:
Haha oops.  I'll fix that!

civis/io/_files.py
mheilman: This should be fixed now.

civis/io/_files.py
```diff
@@ -235,6 +238,16 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
     Small or non-seekable file-like objects will be uploaded with a
     single post.
     """
+    if isinstance(buf, str):
```
stephen-hoover: You should check against `six.string_types` instead of `str`, for Python 2 compatibility.

civis/io/_files.py
```diff
@@ -290,9 +303,26 @@ def civis_to_file(file_id, buf, api_key=None, client=None):
...
     >>> with open("my_file.txt", "wb") as f:
     ...    civis_to_file(file_id, f)
+    >>> # Download a file as a bytes object.
+    >>> import io
+    >>> buf = io.BytesIO()
+    >>> civis_to_file(file_id, buf)
+    >>> # Note that s could be converted to a string with s.decode('utf-8').
+    >>> s = buf.read()
     """
+    if isinstance(buf, str):
```
stephen-hoover: The check should be on `six.string_types`.

stephen-hoover:
LGTM!

PR #182, opened 2017-11-15
[CIVP-13523] Escaping in read_civis has changed from v1.6.0 to v1.7.0 of the python client

jkulzick:


civis/io/_tables.py
```diff
@@ -234,6 +234,11 @@ def read_civis_sql(sql, database, use_pandas=False, job_name=None,
                                                    db_id, credential_id,
                                                    polling_interval)
 
+    # if we retrieved headers then we are performing a parallel unload
+    # in which case we need to specify backslash as the escapechar
+    if headers:
```
keithing: Could we do `if headers is not None:` here?  In case we ever return a numpy array or something else that might cause this to break?

jkulzick: updated

jkulzick:
@keithing merged the other PR and had to resolve a conflict with the CHANGELOG

PR #185, opened 2017-11-16
[CIVP-13534] Escaping in civis_to_csv has changed from v1.6.0 to v1.7.0 of the python client

jkulzick:


civis/io/_tables.py
```diff
@@ -393,7 +382,7 @@ def civis_to_csv(filename, sql, database, job_name=None, api_key=None,
 
     # always set include_header to False and compression to gzip to
```
stephen-hoover: Should this comment about setting include_header to False change now?

jkulzick: updated comment; we're no longer setting include_header and just passing through what the user provided

stephen-hoover:
LGTM!

PR #186, opened 2017-11-20
[CIVP-13581] add diststyle param to IO functions

kbenker:
This PR adds this new parameter to the IO functions. It should also be added to advanced options in the post_syncs endpoint, but that's autogenerated.

Looking at the existing test coverage here, it looks like we don't typically have tests for each specific parameter, but let me know if I should add some for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-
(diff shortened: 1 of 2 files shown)


keithing:
> Looking at the existing test coverage here, it looks like we don't typically have tests for each specific parameter, but let me know if I should add some for this.

Yes, I think we're okay without a test for this specific param.

kbenker:
@keithing Back to you for the CHANGELOG update I previously forgot. :woman_facepalming:

PR #187, opened 2017-11-27
BUG Relax requirement on cloudpickle version

stephen-hoover:
The requirements file states that this library needs cloudpickle~=0.2, but the setup.py file states `cloudpickle>=0.2.0,<=0.3.99`. We don't need to have such a low upper limit in the setup.py file. Relax it to match the requirements.txt file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,13 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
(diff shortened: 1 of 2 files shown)


stephen-hoover:
I'm not sure if this or #188 is better.

stephen-hoover:
I re-read https://packaging.python.org/discussions/install-requires-vs-requirements/ and decided this PR is the better one.

beckermr:
Yay! This just made my life annoying...

PR #188, opened 2017-11-27
BUG Use requirements.txt in setup.py

stephen-hoover:
Instead of repeating requirements in the `setup.py` file, read from `requirements.txt`. This fixes the error that we were being too restrictive on the `cloudpickle` version, and makes sure we don't diverge in requirements again.


stephen-hoover:
Re-read https://packaging.python.org/discussions/install-requires-vs-requirements/ ... okay, let's keep the explicit requirements in `setup.py`.

PR #189, opened 2017-11-30
[CIVP-13616] doubled disk usage in multipart upload

jkulzick:
- `dataframe_to_civis` now writes the dataframe to disk instead of an in memory buffer
- `file_to_civis` will re-write buffers that are not instances of io.BufferedReader to disk
- `_multipart_upload` will no longer re-write file parts to disk; instead it will open new file handles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - ``civis.io.dataframe_to_civis``, ``civis.io.csv_to_civis``, and ``civis.io.civis_file_to_table`` functions now support the `diststyle` parameter.
 
-### Changed
-- ``civis.io.dataframe_to_civis`` writes dataframes to disk instead of using an in memory buffer
-
 ### Fixed
 - Relaxed requirement on ``cloudpickle`` version number (#187)
+
+### Performance Enhancements
 - ``civis.io.file_to_civis`` now uses additional file handles for multipart upload instead of writing to disk to reduce disk usage
+- ``civis.io.dataframe_to_civis`` writes dataframes to disk instead of using an in memory buffer
 
 
 ## 1.7.1 - 2017-11-16
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - ``civis.io.dataframe_to_civis``, ``civis.io.csv_to_civis``, and ``civis.io.civis_file_to_table`` functions now support the `diststyle` parameter.
 
+### Changed
+- ``civis.io.dataframe_to_civis`` writes dataframes to disk instead of using an in memory buffer
+
 ### Fixed
 - Relaxed requirement on ``cloudpickle`` version number (#187)
+- ``civis.io.file_to_civis`` now uses additional file handles for multipart upload instead of writing to disk to reduce disk usage
 
 
 ## 1.7.1 - 2017-11-16
--- a/civis/_utils.py
+++ b/civis/_utils.py
@@ -129,3 +129,19 @@ def retry(exceptions, retries=5, delay=0.5, backoff=2):
         return f_retry
 
     return deco_retry
+
+
+class BufferedPartialReader(object):
+    def __init__(self, buf, max_bytes):
+        self.max_bytes = max_bytes
+        self.bytes_read = 0
+        self.buf = buf
+        self.len = max_bytes
+
+    def read(self, size=-1):
+        if self.bytes_read >= self.max_bytes:
+            return b''
+        bytes_left = self.max_bytes - self.bytes_read
+        data = self.buf.read(min(size, bytes_left))
+        self.bytes_read += len(data)
+        return data
--- a/civis/_utils.py
+++ b/civis/_utils.py
@@ -133,15 +133,19 @@ def retry(exceptions, retries=5, delay=0.5, backoff=2):
 
 class BufferedPartialReader(object):
     def __init__(self, buf, max_bytes):
+        self.buf = buf
         self.max_bytes = max_bytes
         self.bytes_read = 0
-        self.buf = buf
         self.len = max_bytes
 
     def read(self, size=-1):
         if self.bytes_read >= self.max_bytes:
             return b''
         bytes_left = self.max_bytes - self.bytes_read
-        data = self.buf.read(min(size, bytes_left))
+        if size < 0:
+            bytes_to_read = bytes_left
+        else:
+            bytes_to_read = min(size, bytes_left)
+        data = self.buf.read(bytes_to_read)
         self.bytes_read += len(data)
         return data
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -25,7 +25,7 @@ try:
 except ImportError:
     HAS_PANDAS = False
 
-MIN_MULTIPART_SIZE = 10 * 2 ** 20  # 50MB # >>>
+MIN_MULTIPART_SIZE = 50 * 2 ** 20  # 50MB
 MIN_PART_SIZE = 5 * 2 ** 20  # 5MB
 MAX_PART_SIZE = 5 * 2 ** 30  # 5GB
 MAX_FILE_SIZE = 5 * 2 ** 40  # 5TB
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -229,7 +229,7 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
     # if buf is not a file handle or if current position is not 0
     # then write to a file
     if not isinstance(buf, io.BufferedReader) or buf.tell() != 0:
-       with TemporaryDirectory() as tmp_dir:
+        with TemporaryDirectory() as tmp_dir:
             tmp_path = os.path.join(tmp_dir, 'file_to_civis.csv')
             try:
                 with open(tmp_path, 'wb') as fin:
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -10,12 +10,11 @@ import re
 import shutil
 import six
 import requests
-import tempfile
 from requests import HTTPError
 
 from civis import APIClient, find_one
 from civis.base import CivisAPIError, EmptyResultError
-from civis.compat import FileNotFoundError
+from civis.compat import FileNotFoundError, TemporaryDirectory
 from civis.utils._deprecation import deprecate_param
 from civis._utils import BufferedPartialReader, retry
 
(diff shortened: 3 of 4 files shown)


civis/io/_files.py
```diff
@@ -242,6 +228,21 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
         with open(buf, 'rb') as f:
             return _file_to_civis(
                 f, name, api_key=api_key, client=client, **kwargs)
+
+    # if buf is not a file handle or if current position is not 0
+    # then write to a file
+    if not isinstance(buf, io.BufferedReader) or buf.tell() != 0:
+        with TemporaryDirectory() as tmp_dir:
+            tmp_path = os.path.join(tmp_dir, 'file_to_civis.csv')
+            try:
```
jkulzick: Let me know if you have any ideas on a more elegant way of doing this. I found that buffers produced by some modules, such as zipfile, have a mode attribute that is misleading in that the mode is 'r' but only bytes are ever returned. I figured this was the safest way given we don't know what mode the buffer was opened in.

civis/_utils.py
```diff
@@ -129,3 +129,19 @@ def f_retry(*args, **kwargs):
...
+
+
+class BufferedPartialReader(object):
+    def __init__(self, buf, max_bytes):
+        self.max_bytes = max_bytes
+        self.bytes_read = 0
+        self.buf = buf
+        self.len = max_bytes
+
+    def read(self, size=-1):
```
stephen-hoover: This defaults `size` to `-1`, but doesn't seem to handle the size=-1 case. `size=-1` should set `size` to `bytes_left`, right?

civis/io/_files.py
```diff
@@ -132,25 +134,21 @@ def _multipart_upload(buf, name, file_size, client, **kwargs):
...
+        offset = part_size * part_num
+        num_bytes = min(part_size, file_size - offset)
+
         log.debug('Uploading file part %s', part_num)
-        file_out = part_path.format(part_num)
-        with open(file_out, 'rb') as fout:
-            part_response = requests.put(part_url, data=fout)
+        mode = 'r'
+        if 'b' in buf.mode:
+            mode += 'b'
```
stephen-hoover: Why only open as bytes if the original buffer did? Wouldn't you always want to open the file as bytes?

civis/io/_files.py
```diff
@@ -132,25 +134,21 @@ def _multipart_upload(buf, name, file_size, client, **kwargs):
...
+        with open(buf.name, mode) as fout:
```
stephen-hoover: I find it confusing to name a read-only buffer `fout`. I use that name for files I'm writing. Maybe that's just me? ... Oh, I see you have naming conventions reversed from what I use. It looks like this repo currently uses both conventions.

civis/io/_files.py
```diff
@@ -268,11 +272,12 @@ def _file_to_civis(buf, name, api_key=None, client=None, **kwargs):
     elif file_size > MAX_FILE_SIZE:
         msg = "File is greater than the maximum allowable file size (5TB)"
         raise ValueError(msg)
-    elif not is_seekable and file_size > MAX_PART_SIZE:
-        msg = "Cannot perform multipart upload on non-seekable files. " \
+    elif not multipart_upload and file_size > MAX_PART_SIZE:
+        msg = "Cannot perform multipart upload on non-seekable files or " \
```
stephen-hoover: We do perform multi-part uploads on non-seekable files, don't we? We write those to disk first.

civis/io/_files.py
```diff
@@ -256,10 +257,13 @@ def _file_to_civis(buf, name, api_key=None, client=None, **kwargs):
...
                     'single post. Files over 5GB will fail.')
 
-    # determine if file-like object is seekable
+    # determine if file-like object will support multipart upload
+    multipart_upload = False
     try:
         buf.seek(buf.tell())
         is_seekable = True
+        if hasattr(buf, 'name') and hasattr(buf, 'mode'):
+            multipart_upload = True
```
stephen-hoover: Could you use only these hasattr checks instead of trying to seek? Shouldn't every buffer in this function be seekable?

civis/io/_tables.py
```diff
@@ -622,17 +622,16 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
...
-    else:
-        txt = buf
-    df.to_csv(txt, encoding='utf-8', index=False, **kwargs)
-    txt.flush()
-    buf.seek(0)
+
+    with TemporaryDirectory() as tmp_dir:
+        tmp_path = os.path.join(tmp_dir, 'dataframe_to_civis.csv')
+        with open(tmp_path, 'w') as fin:
+            df.to_csv(fin, encoding='utf-8', index=False, **kwargs)
```
stephen-hoover: Is there a reason why you open the file yourself rather than giving the filename to `pandas`?

civis/io/_tables.py
```diff
@@ -622,17 +622,16 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
...
+        with open(tmp_path, 'rb') as fout:
+            name = table.split('.')[-1]
+            file_id = file_to_civis(fout, name, client=client)
```
stephen-hoover: You can give `tmp_path` directly to `file_to_civis` now.

CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - ``civis.io.dataframe_to_civis``, ``civis.io.csv_to_civis``, and ``civis.io.civis_file_to_table`` functions now support the `diststyle` parameter.
 
+### Changed
+- ``civis.io.dataframe_to_civis`` writes dataframes to disk instead of using an in memory buffer
+
 ### Fixed
 - Relaxed requirement on ``cloudpickle`` version number (#187)
+- ``civis.io.file_to_civis`` now uses additional file handles for multipart upload instead of writing to disk to reduce disk usage
```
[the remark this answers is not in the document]
jkulzick: updated

civis/_utils.py
jkulzick: good catch; fixed

civis/io/_files.py
```diff
@@ -34,6 +35,7 @@
                     requests.ConnectTimeout)
 
 log = logging.getLogger(__name__)
+CHUNK_SIZE = 32 * 1024
```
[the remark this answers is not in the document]
jkulzick: I've added a comment though I will note that I'm utilizing the chunk size that was already in place in the repo

civis/io/_files.py
jkulzick: overthought this one; you are correct, we should always open as bytes

civis/io/_files.py
jkulzick: I agree! I think your convention makes more sense. I've updated references to use `fin` for read and `fout` for write.

civis/io/_files.py
jkulzick: Theoretically I would get right of this whole try block and the use of `is_seekable` and `multipart_upload` variables because in `file_to_civis` we are checking if the buffer `isinstance(buf, io.BufferedReader)` which should ensure that it is seekable and has these attributes.

I have come across file like objects that have a seek method, but fail when you call it, which makes me sad.

@stephen-hoover Let me know what you think.

civis/io/_files.py
jkulzick: True, we now write these to disk. I think this is related to the ^^ comment and discussion.

civis/io/_tables.py
jkulzick: No; good catch

civis/io/_tables.py
jkulzick: Another good catch!

civis/io/_files.py
stephen-hoover: I just realized that `is_seekable` shouldn't be necessary any longer because we write files to disk if `.tell() != 0`. So we should be able to remove the `.seek` call, correct? 

We'd still be unable to do a multipart upload if we can't find a `name`.  Do you think it's better to handle that by writing to disk and doing a multipart upload, or by reverting to `_single_upload` as we are now? I think the `_single_upload` is better, but I could easily be convinced to write to disk and treat everything the same.

jkulzick: I think we still need the `.seek` for retries and to seek to different positions on the buffer for multipart upload.

I struggled a bit with deciding what checks should go in `file_to_civis` versus `_file_to_civis`. The `isinstance(buf, io.BufferedReader)` check is designed to catch in memory buffers. If that check is satisfied, then the file like object should have a `name`, unless someone did something silly, and it should be seekable no? As long as `file_to_civis` fails and returns the underlying error should we try to handle edge cases like `is_seekable` or `name`?

Regardless, I feel like the `isinstance` check is us changing how the `file_to_civis` function works, but making the function backwards compatible so the `file_to_civis` location makes sense.

The `is_seekable` check is for determining if we can do retries or multipart upload (which requires seeking), which impacts what branch of the code is executed so it feels like that belongs in `_file_to_civis`.

Checks on `buf.tell` and `buf.name` are guardrails to handle possible edge cases.

stephen-hoover: It looks like the `io.BufferedReader` is guaranteed to have `seek`, but not guaranteed to have `name`.

The multipart upload only seeks on the buffer we open ourselves from the file on disk, correct? It doesn't need the input buffer to be seekable. You're right about retries, though. But we only need to seek for retries if we're uploading directly from an in-memory buffer. So in which cases do we want to upload from the in-memory buffer, and in which cases do we read from disk and do a multipart upload?

jkulzick: Thats correct, for multipart upload we seek on the buffer we open ourselves.

I don't think we're allowing uploads from an in-memory buffer, since we re-write in-memory buffers to disk if users pass them in `file_to_civis` before calling `_file_to_civis`. Given that we could arguable strip out `seekable` check altogether couldn't we? Although, I don't know if its good to keep it for visibility.

civis/io/_files.py
```diff
@@ -132,25 +135,18 @@ def _multipart_upload(buf, name, file_size, client, **kwargs):
...
-            if length <= 0:
-                break
-            bytes_read += length
-            data = part_buf.read(length)
-            yield data
-
     # upload function wrapped with a retry decorator
     @retry(RETRY_EXCEPTIONS)
-    def _upload_part_base(item, part_path):
+    def _upload_part_base(item, buf, part_size, file_size):
```
stephen-hoover: You only need the file name here, correct? It's probably better to make that the argument instead of the whole buffer. That will avoid any possible thread safety or pickling issues when putting this function in a pool.

civis/io/_files.py
stephen-hoover: Since we're making strong assumptions about what goes into this function already, I think we should leverage that to make the code as simple as possible.

jkulzick:
alright, I removed the use of `is_seekable` and checks for `name` attribute. this simplified the if block in `_file_to_civis` and the `_single_upload` function.

I put a comment in `file_to_civis` around the expected attributes of the `buf` passed to `_file_to_civis`

PR #191, opened 2017-12-04
add example of how to create civis futures to user guide

mheilman:
The issue of how to create futures objects has come up a few times recently, so I think it'd be worth adding an example to the documentation.

### DIFF
--- a/docs/source/user_guide.rst
+++ b/docs/source/user_guide.rst
@@ -53,6 +53,45 @@ calling ``result()`` on ``fut`` above forces the program to wait for the
 job started with :func:`~civis.io.dataframe_to_civis` to finish and
 returns the result.
 
+You can create :class:`CivisFuture <civis.futures.CivisFuture>` objects for
+many tasks (e.g., scripts, imports):
+
+.. code-block:: python
+
+   >>> import civis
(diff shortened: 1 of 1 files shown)


mheilman:
@stephen-hoover or @keithing, could one of you please take a look?

docs/source/user_guide.rst
```diff
@@ -53,6 +53,45 @@ calling ``result()`` on ``fut`` above forces the program to wait for the
 job started with :func:`~civis.io.dataframe_to_civis` to finish and
 returns the result.
 
+You can create :class:`CivisFuture <civis.futures.CivisFuture>` objects for
+many tasks (e.g., scripts, imports):
```
keithing: This is great! If you don't mind, it would be nice to have 2 or so sentences here just describing high level what the code below is doing.  Perhaps something like

> Here we will be creating a container script that does the simple task of printing the text "HELLO WORLD".  We will be creating the script, executing it and then waiting for the script to finish.

mheilman:
@keithing, does ad3b944 work?

keithing:
> does ad3b944 work?

Yeah, thanks!

mheilman:
rebased

PR #192, opened 2017-12-04
[CIVP-13740] Bug in buffer rewrite in file_to_civis

jkulzick:
Two small changes:
1. If the file is opened as text it will be an instance of io.TextIOWrapper. We were previously re-writing those files. There is no reason to do that.
2. More importantly, we were not returning `buf` to its original position in the `Except` block so when trying to read files opened as text we were writing partial files


jkulzick:
@stephen-hoover for viz

civis/io/_files.py
```diff
@@ -225,15 +225,18 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
 
     # we should only pass _file_to_civis a file-like object that is
     # on disk, seekable and at position 0
-    if not isinstance(buf, io.BufferedReader) or buf.tell() != 0:
+    if not isinstance(buf, (io.BufferedReader, io.TextIOWrapper)) or \
+            buf.tell() != 0:
+        # read some bytes to determine proper mode for writing
+        mode = 'w'
+        data = buf.read(5)
```
mheilman: would `buf.read(0)` work?   if so, you could put that in the `isinstance` and get rid of this line and  `fout.write(data)` below.

civis/io/_files.py
```diff
@@ -225,15 +225,18 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
 
     # we should only pass _file_to_civis a file-like object that is
     # on disk, seekable and at position 0
-    if not isinstance(buf, io.BufferedReader) or buf.tell() != 0:
+    if not isinstance(buf, (io.BufferedReader, io.TextIOWrapper)) or \
```
mheilman: Can we just check `not isinstance(buf, io.IOBase) or buf.tell() != 0`?  ([IOBase docs](https://docs.python.org/3/library/io.html#io.IOBase))

```
In [16]: isinstance(io.TextIOWrapper(io.StringIO()), io.IOBase)
Out[16]: True

In [17]: isinstance(io.BufferedReader(io.StringIO()), io.IOBase)
Out[17]: True
```

civis/tests/test_io.py
```diff
@@ -107,11 +107,10 @@ def test_zip_member_to_civis(self, *mocks):
 
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_file_to_civis(self, *mocks):
-        with tempfile.NamedTemporaryFile() as tmp:
-            tmp.write(b'a,b,c\n1,2,3')
-            tmp.flush()
-            tmp.seek(0)
-            result = civis.io.file_to_civis(tmp, tmp.name)
+        buf = StringIO()
```
mheilman: At a glance, it seems like we ought to also have a bytes-version of this in here:

```
buf = BytesIO()
buf.write(b'a,b,c\n1,2,3')
buf.seek(0)
result = civis.io.file_to_civis(buf, 'somename')
assert isinstance(result, int)
```

mheilman:
Could you please explain why we were writing partial files a bit more?  Is it that the `copyfileobj` call in the try part advanced the position in the text file before triggering the exception?

civis/io/_files.py
jkulzick: that works. good idea.

civis/io/_files.py
jkulzick: I don't think so, because `io.StringIO` and `io.BytesIO` both inherit from `io.IOBase` and the goal is to re-write those

civis/tests/test_io.py
jkulzick: I'll rename this to `test_text_file_to_civis` and add a `test_bytes_file_to_civis`.

Also, for transparency, I discussed these tests with Stephen and we agreed that we should add integration tests similar to what have been implemented in the R client where we actually test different files and validate not only that they were successfully uploaded and but also that they were uploaded with the correct contents. Those tests are not within the scope of this PR though.

jkulzick:
That's correct. `copyfileobj` advances the buffer in the try block which means that we need to seek back to the original position in the exception block. However, we can't assume that the `buf` will be seekable, which is why I switched to this method. One example of this is if someone uses the `zipfile` package to open a member of a zip archive. This scenario is captured in `test_io.test_zip_member_to_civis`.

PR #194, opened 2017-12-06
[CIVP-14018] civis_to_csv doesn't return a gzip when requested for include_header=False

jkulzick:
- only override gzip compression for include_header=True
- add tests for gzip compression and include_header=True/False

The Civis API will return different results when exporting data with compression='gzip' when include_header=True vs include_header=False. Specifically, include_header=True will not return a gzipped file and include_header=False will. This is due to differences in how headers are being applied to the file urls by the API.


jkulzick:
@keithing @mheilman not sure which is available to review this. Its to address [PARO-545](https://civisanalytics.atlassian.net/browse/PARO-545)

keithing:
Looks like there is a python2.7 hiccup!  Do you mind addressing that?

jkulzick:
fixed. `gzip.compress` is not available in python2.7.

PR #195, opened 2017-12-06
[CIVP-14018] civis_to_csv doesn't return a gzip when requested for include_header=False

jkulzick:
The Civis API will return different response-content-type headers in file urls when exporting data with compression='gzip' when include_header=True vs include_header=False. As a result prior to version 1.7.1 of civis-python, civis_to_csv with compression='gzip' has the following behavior:
- include_header=True returns a decompressed file
- include_header=False returns a gzip compressed file

In order to maintain backwards compatibility, we need to ensure we still return a gzip compressed file when include_header=False.


civis/io/_tables.py
```diff
@@ -925,8 +925,9 @@ def _download_file(url, local_path, headers, compression):
 
     # gzipped buffers can be concatenated so write headers as gzip
     if compression == 'gzip':
-        with open(local_path, 'wb') as fout:
-            fout.write(gzip.compress(headers))
+        with gzip.open(local_path, 'wb') as fout:
+            fout.write(headers)
```
keithing: Double checking:  does headers need a `headers.encode('utf-8')` here?

keithing: > headers are being passed into this function as bytes 

Great.  This is fine as is then.

jkulzick:
sounds good. docstring updated.

PR #197, opened 2017-12-11
ENH CLI commands to open notebooks

stephen-hoover:
Add commands to the CLI to allow users to more easily open new notebooks. New commands are:
- launch : Create a new notebook, deploy it, and open it in a new brower tab.
- up : Start an existing notebook running and open it in a new browser tab.
- down : Stop a running notebook.
- open : Open an existing notebook in a new browser tab (if it's not running, don't start it).

These are usability improvements to make Civis Platform as a notebook backend even more transparent to users.

### DIFF
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -86,7 +86,7 @@ def notebooks_download_cmd(notebook_id, path):
             f.write(lines)
 
 
-@click.command('launch')
+@click.command('new')
 @click.argument('language', type=click.Choice(['python3', 'python2', 'r']),
                 default='python3')
 @click.option('--mem', type=int, default=None,
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -92,7 +92,7 @@ def notebooks_download_cmd(notebook_id, path):
 @click.option('--mem', type=int, default=None,
               help='Memory allocated for this notebook in MiB.')
 @click.option('--cpu', type=int, default=None,
-              help='CPU available for this notebook in 1/1000 of a core.')
+              help='CPU available for this notebook in 1/1024 of a core.')
 def notebooks_launch_cmd(language='python3', mem=None, cpu=None):
     """Create a new notebook and open it in the browser."""
     client = civis.APIClient()
@@ -109,7 +109,7 @@ def notebooks_launch_cmd(language='python3', mem=None, cpu=None):
 @click.option('--mem', type=int, default=None,
               help='Memory allocated for this notebook in MiB.')
 @click.option('--cpu', type=int, default=None,
-              help='CPU available for this notebook in 1/1000 of a core.')
+              help='CPU available for this notebook in 1/1024 of a core.')
 def notebooks_up(notebook_id, mem=None, cpu=None):
     """Start an existing notebook and open it in the browser."""
     client = civis.APIClient()
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -93,7 +93,7 @@ def notebooks_download_cmd(notebook_id, path):
               help='Memory allocated for this notebook in MiB.')
 @click.option('--cpu', type=int, default=None,
               help='CPU available for this notebook in 1/1000 of a core.')
-def notebooks_launch_cmd(language='python3', mem=None, cpu=None):
+def notebooks_new_cmd(language='python3', mem=None, cpu=None):
     """Create a new notebook and open it in the browser."""
     client = civis.APIClient()
     kwargs = {'memory': mem, 'cpu': cpu}
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -8,6 +8,7 @@ import os
 
 import click
 import requests
+import webbrowser
 
 import civis
 from civis.io import file_to_civis, civis_to_file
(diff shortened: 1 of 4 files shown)


stephen-hoover:
Instead of "civis notebooks launch", how about making the main startup command "civis notebook", to make it even more similar to "jupyter notebook"? I'm also not sure how useful the other three commands are , "up", "down", and "open". They're really easy to add, so I put them there in this PR.

keithing:
> how about making the main startup command "civis notebook"

It's somewhat awkward because we already have `civis notebooks`, but I think that makes a lot of sense.  We could be sure to document that `civis notebooks` is for raw API access.  This way we could subset set `down` with `--down`, etc.

mheilman:
`civis notebooks launch` seems ok to me, but a couple alternatives come to mind:

* `civis notebooks up` without an ID
* `civis notebooks init`, which I think follows [vagrant](https://www.vagrantup.com/intro/getting-started/index.html), which also uses up/down.

mheilman:
In a future PR, it'd be nice to make an .ipynb file an optional arg of the launch/new/init command.  I'd be happy to add that.  (My current workaround is `civis notebooks patch 12345 --file-id $(civis files upload localnotebook.ipynb)`.)

civis/cli/_cli_commands.py
```diff
@@ -85,6 +86,71 @@ def notebooks_download_cmd(notebook_id, path):
             f.write(lines)
 
 
+@click.command('launch')
+@click.argument('language', type=click.Choice(['python3', 'python2', 'r']),
+                default='python3')
+@click.option('--mem', type=int, default=None,
+              help='Memory allocated for this notebook in MiB.')
+@click.option('--cpu', type=int, default=None,
+              help='CPU available for this notebook in 1/1000 of a core.')
```
mheilman: is it 1024 of a core?  cf. https://platform.civisanalytics.com/#/api#post_scripts_containers

stephen-hoover: Yes, I think it is. Fixed.

stephen-hoover:
@keithing , talking with Mike pushed me back away from "civis notebook". It's probably too confusing, given that all the other notebook commands are under "civis notebooks". I changed "launch" to "new" to make it slightly shorter. "init" doesn't seem quite right, since the command also starts and opens the notebook. How's this set of names for commands?
- new
- up
- down
- open

beckermr:
As a quick comment on this, I wrote a similar CLI for workflows here https://github.com/civisanalytics/modeling-dsrd/blob/civisflow/civisflow/cli.py. 

Are we open to a similar PR for that?

civis/cli/_cli_commands.py
```diff
@@ -85,6 +86,71 @@ def notebooks_download_cmd(notebook_id, path):
...
+              help='Memory allocated for this notebook in MiB.')
+@click.option('--cpu', type=int, default=None,
+              help='CPU available for this notebook in 1/1024 of a core.')
+def notebooks_launch_cmd(language='python3', mem=None, cpu=None):
+    """Create a new notebook and open it in the browser."""
+    client = civis.APIClient()
+    kwargs = {'memory': mem, 'cpu': cpu}
+    kwargs = {k: v for k, v in kwargs.items() if v is not None}
+    new_nb = client.notebooks.post(language=language, **kwargs)
+    print("Created new {} notebook with ID {} .".format(language, new_nb.id))
```
beckermr: You might consider using `click.echo` here, BTW. http://click.pocoo.org/6/quickstart/#echoing

They claim it is better between python 2 and 3. I am not sure if we care.

civis/cli/_cli_commands.py
```diff
@@ -85,6 +86,71 @@ def notebooks_download_cmd(notebook_id, path):
             f.write(lines)
 
 
+@click.command('new')
+@click.argument('language', type=click.Choice(['python3', 'python2', 'r']),
+                default='python3')
+@click.option('--mem', type=int, default=None,
+              help='Memory allocated for this notebook in MiB.')
+@click.option('--cpu', type=int, default=None,
+              help='CPU available for this notebook in 1/1024 of a core.')
```
beckermr: I conferred with Saranga and for K8S which powers notebooks, the units are 1/1000 of a core. Container scripts are 1/1024.

stephen-hoover: Ugh. Okay, changed back.

civis/cli/_cli_commands.py
```diff
@@ -85,6 +86,71 @@ def notebooks_download_cmd(notebook_id, path):
...
 
 
+@click.command('new')
+@click.argument('language', type=click.Choice(['python3', 'python2', 'r']),
+                default='python3')
+@click.option('--mem', type=int, default=None,
+              help='Memory allocated for this notebook in MiB.')
+@click.option('--cpu', type=int, default=None,
+              help='CPU available for this notebook in 1/1000 of a core.')
+def notebooks_launch_cmd(language='python3', mem=None, cpu=None):
```
mheilman: probably worth changing this to `notebooks_new_cmd`, for consistency

stephen-hoover: Thanks for catching that. Done.

mheilman:
@beckermr, re: a workflows command, I don't have a strong opinion either way.  It looks pretty concise and potentially useful for some folks.  Maybe wait for @keithing to chime in on that?

beckermr:
sure @mheilman

stephen-hoover:
If there's further discussion on a workflows command, it should move off of this PR, maybe into an issue.

PR #200, opened 2018-01-08
ENH Use feather to send data to CivisML, if possible

stephen-hoover:
When users provide a `DataFrame` for model training or prediction, try first to send data with feather. This requires an additional package installation, but it's faster and it guarantees that we get the type correct for columns of categorical data. Require that we use feather for data with `pd.Categorical` columns, since we can't guarantee that we get the types correct when going through CSVs.
  
Feather does have a couple of limitations. It can't write column names which aren't strings, and it can't write object-type columns with non-string data. I'm not sure this is a problem -- CSVs have the same limitation, but they fail silently. Non-string column names will be read back in as strings, and object columns will either be read as entirely strings or entirely numbers, depending on how they can be coerced.


keithing:
>quote Require that we use feather for data with pd.Categorical columns

It looks like `feather-format` is a soft import.  Should we raise a warning rather than an Exception for this situation?

stephen-hoover:
One of the driving motivations for this change is that users who wish for CivisML to treat a column of integer features as a categorical might (reasonably) assume that they could cast the column in their `pd.DataFrame` as a `pd.Categorical`. With the current release of the API client, the DataFrame would be written as a CSV, which loses the type information. The column would be interpreted as a numerical column and not expanded.

I think the only way to be certain that we model with the data intended by the user is to use a type-aware format, such as feather. Users who don't need the column to be treated as a categorical could convert it to another type if they don't want to install `feather-format`. I'm uncomfortable with letting a categorical go by with only a warning, since we'd be losing information about types when sending it to CivisML.

keithing:
> The column would be interpreted as a numerical column and not expanded . . . 

That would be a problem.  That sounds like a good reason for requiring it to me.  What happens to `object` type data?

stephen-hoover:
I added a note on a couple of extra feather limitations to the OP. Feather (at least as it currently is) doesn't support `object` columns which aren't entirely strings. Object columns of numbers or mixed numbers and strings cause an error when I try to save them. This is non-ideal, but it still seems like an improvement over CSVs. For CSVs, writing will succeed, but we'll read in different types on the other end.

stephen-hoover:
@keithing , I realized I'd forgotten to put a version check in here. We want to continue writing CSVs if the user is creating a train or predict job from versions <= 2.0. Could you review the new commit?

PR #203, opened 2018-01-11
ENH default to dynamically calculating number of training jobs in CivisML

elsander:
This PR changes the default `n_jobs` in `ModelPipeline.train` from 4 to `None`, so that it will be calculated dynamically by default. I also updated the docs to reflect that change, and documented `ModelPipeline`'s soft dependency on `glmnet` and `civisml-extensions` for accessing model objects.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -909,7 +909,8 @@ class ModelPipeline:
         n_jobs : int, optional
             Number of jobs to use for training and validation. Defaults to
             `None`, which allows CivisML to dynamically calculate an
-            appropriate number of workers to use.
+            appropriate number of workers to use (in general, as many as
+            possible without using all resources in the cluster).
             Increase n_jobs to parallelize over many hyperparameter
             combinations in grid search/hyperband, or decrease to use fewer
             computational resources at once.
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -284,6 +284,7 @@ class ModelFuture(ContainerFuture):
     civis.futures.ContainerFuture
     concurrent.futures.Future
     """
+
     def __init__(self, job_id, run_id, train_job_id=None, train_run_id=None,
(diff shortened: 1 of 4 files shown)


docs/source/index.rst
```diff
@@ -37,17 +37,22 @@ set ``use_pandas=True`` in functions that accept that parameter.  To install
    pip install pandas
 
 Machine learning features in the ``ml`` namespace have a soft dependency on
-``scikit-learn`` and ``pandas``. Install ``scikit-learn`` to
+``scikit-learn``, ``pandas``, ``civisml-extensions``, and
+``glmnet``. Install ``scikit-learn`` to
```
stephen-hoover: I think it's not quite right to call our relationship with `civisml-extensions` and `glmnet` a soft dependency. Installing those won't really change what the API client code will do for you, while installing, say, scikit-learn does add new capabilities.

I think it's better to have a separate paragraph noting that various CivisML models have open-source dependencies beyond scikit-learn, and that if users want to download their model, they would need to install those dependencies. The full description is probably more appropriate for `ml.rst`, but a mention here isn't a bad idea. 

If you're documenting these libraries, you should also include `muffnn`. The MLP models use that library.

civis/ml/_model.py
```diff
@@ -907,7 +908,8 @@ def train(self, df=None, csv_path=None, table_name=None,
             for validation; and `'skip'`, which skips the validation step.
         n_jobs : int, optional
             Number of jobs to use for training and validation. Defaults to
-            4, which allows parallelization over the 4 cross validation folds.
+            `None`, which allows CivisML to dynamically calculate an
+            appropriate number of workers to use.
```
stephen-hoover: I see the specifics are in the full docs, but is it worth mentioning here that "appropriate" generally means "as many as possible without using your entire cluster"?

PR #204, opened 2018-01-12
ENH Re-raise warnings from CivisML

stephen-hoover:
When retrieving metadata, check the "warnings" key (present on all jobs) for any warnings which were recorded during the CivisML run. Reraise them as exactly as possible. We always attempt to download metadata after a job completes so that we can check for any errors, so we'll always see warnings from jobs as they complete.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -1,5 +1,6 @@
 """Run CivisML jobs and retrieve the results
 """
+import builtins
 from builtins import super
 from collections import namedtuple
 import io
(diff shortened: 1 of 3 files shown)


stephen-hoover:
FYI @elsander .

civis/ml/tests/test_model.py
```diff
@@ -451,12 +471,12 @@ def test_table_None(mock_res, mock_lt):
...
 @mock.patch.object(_model.cio, "file_id_from_run_output",
                    mock.Mock(return_value=11, spec_set=True))
-@mock.patch.object(_model.cio, "file_to_json", return_value='bar')
+@mock.patch.object(_model.cio, "file_to_json", return_value={'foo': 'bar'})
 @mock.patch.object(_model.ModelFuture, "_set_model_exception")
 def test_metadata(mock_spec, mock_f2j):
     c = setup_client_mock(3, 7)
     mf = _model.ModelFuture(3, 7, client=c)
-    assert mf.metadata == 'bar'
+    assert mf.metadata == {'foo': 'bar'}
```
keithing: Why make this change?  Is this because the test before the change raises a warning after these changes?

stephen-hoover: The metadata property now assumes that the metadata is always in the form of a dictionary. It calls `get` on the object. This is in L523 of `_model.py`.

civis/ml/_model.py
```diff
@@ -476,6 +520,7 @@ def metadata(self):
             fid = cio.file_id_from_run_output('model_info.json', self.job_id,
                                               self.run_id, client=self.client)
             self._metadata = cio.file_to_json(fid, client=self.client)
+            _show_civisml_warnings(self._metadata.get('warnings', []))
```
stephen-hoover: This line requires the change in `test_metadata`.

PR #208, opened 2018-01-18
ENH Allow missing model validation metadata

stephen-hoover:
With CivisML v2, users can choose to skip the validation step. If that happens, there will be no validation metadata stored. This change lets the `ModelFuture` handle missing validation metadata files without issuing an error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   speed improvement; it reads and writes faster than CSVs and produces smaller files (#200).
 - ``ModelFuture`` objects will emit any warnings which occurred during their
   corresponding CivisML job (#204)
+- ``ModelFuture`` objects will return an empty dictionary if validation metadata
+  are missing, rather than issuing an exception or returning ``None`` (#208)
 
 ### Fixed
 - Restored the pre-v1.7.0 default behavior of the ``joblib`` backend by setting the ``remote_backend``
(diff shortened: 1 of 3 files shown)


stephen-hoover:
FYI @elsander

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   speed improvement; it reads and writes faster than CSVs and produces smaller files (#200).
 - ``ModelFuture`` objects will emit any warnings which occurred during their
   corresponding CivisML job (#204)
+- ``ModelFuture`` objects will return an empty dictionary if validation metadata
```
jmemich: From what I can tell below `metrics` defaults to [`None`](https://github.com/civisanalytics/civis-python/pull/208/files#diff-f3b644e69a0d393e0ee202a864d7e190R557).

stephen-hoover: I meant this line to refer to `validation_metadata`, not `metrics`, but it's obviously too vague. You raise a good point, though. It seems better to stay consistent here. I modified the code so that it would return `None` for both `validation_metadata` and `metrics` if the validation metadata were missing. This makes the behavior the same as it was before.

(Previously, skipping validation would store `None` in a file of validation metadata. In v2.1, we won't write the file at all.)

elsander:
@stephen-hoover Did you run a Platform job to test that this works? When validation is skipped, I'm still getting a failure on `train.result()`, even though the Platform job itself succeeds. This is the acceptance testing notebook, with the `civis` version pinned to the master commit where this merged: https://platform.civisanalytics.com/#/notebooks/3057, and the associated successful training job: https://platform.civisanalytics.com/#/scripts/custom/9208581

elsander:
Actually, it looks like I can't replicate this locally. There must have been a dependency issue in the notebook.

PR #210, opened 2018-01-19
In Which We Allow Callers to Over-Write Default to_csv kwargs in dataframe_to_civis

waltaskew:
I hit a bug where I couldn't pass `index=True` to write my datframe's index with dataframe_to_civis.
This allows callers to overwrite the default `to_csv` kwargs if they'd like.

@keithing

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -626,7 +626,9 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
 
     with TemporaryDirectory() as tmp_dir:
         tmp_path = os.path.join(tmp_dir, 'dataframe_to_civis.csv')
-        df.to_csv(tmp_path, encoding='utf-8', index=False, **kwargs)
+        to_csv_kwargs = {'encoding': 'utf-8', 'index': False}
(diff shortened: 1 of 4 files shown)


civis/io/_tables.py
```diff
@@ -626,7 +626,9 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
 
     with TemporaryDirectory() as tmp_dir:
         tmp_path = os.path.join(tmp_dir, 'dataframe_to_civis.csv')
-        df.to_csv(tmp_path, encoding='utf-8', index=False, **kwargs)
+        to_csv_kwargs = {'encoding': 'utf-8', 'index': False}
+        to_csv_kwargs.update(kwargs)
+        df.to_csv(tmp_path, **to_csv_kwargs)
```
keithing: @jkulzick, is there a reason to prevent users from changing `encoding` or `utf-8` here?

jkulzick: I can't think of one, we should probably let the user override these

waltaskew:
I just had to rebase some merge conflicts on the CHANGELOG, so I'm afraid you'll need to re-LGTM.

PR #212, opened 2018-01-22
ENH Increase max CivisML template to v2.1

stephen-hoover:
Also modify the template fallback logic such that users get the most recent template to which they have access, in case there's more than one such template.


stephen-hoover:
FYI, @elsander .

PR #215, opened 2018-01-24
remove extra space in the notebooks new CLI command output

mheilman:
@stephen-hoover, did you have this space there intentionally?  I just used `civis notebooks new`, and the space seemed odd.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Changed
+
+- Reworded output of `civis notebooks new` CLI command (#215)
 
 ## 1.8.0 - 2018-01-23
 ### Added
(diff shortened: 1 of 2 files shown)


stephen-hoover:
That was intentional. That ID number is the kind of thing which I personally would like to copy-paste. If the number is surrounded by spaces, then I can highlight it with a double-click. If there's a period immediately after, then the double-click highlights the period as well. I'd prefer to keep the number surrounded by spaces -- maybe there's a way to rewrite the output so that it's not at the end of the sentence? Or if this issue is particular to me, I don't mind removing the space.

mheilman:
Ah, how about this?

```
print("The new {} notebook with ID {} was created.".format(language, new_nb.id))
```

stephen-hoover:
Hm... the passive voice doesn't look good to me. What about adding "in Civis Platform" to the end of the existing sentence?

mheilman:
How's that?  I added the URL, too, so the output is like this:

```
Created new python3 notebook with ID 3095 in Civis Platform (https://platform.civisanalytics.com/#/notebooks/3095).
```

PR #216, opened 2018-01-31
Add script for smoke tests

mheilman:
This adds a script for running some simple tests that hit the live Civis platform.

### DIFF
--- a/tools/smoke_tests.py
+++ b/tools/smoke_tests.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python3
+
+"""
+This script tests end-to-end functionality using the Civis Python client.
+It uses the live Civis API and Redshift, so a valid CIVIS_API_KEY is needed.
+
+This is based on a similar script for the R client:
+https://github.com/civisanalytics/civis-r/blob/master/tools/integration_tests/smoke_test.R
+"""
+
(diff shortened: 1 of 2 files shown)


tools/smoke_tests.py
```diff
@@ -0,0 +1,95 @@
...
+    logger.info('Testing reading from redshift...')
+    sql = "SELECT * FROM datascience.iris"
+    df1 = civis.io.read_civis_sql(
+        sql=sql, database=database, use_pandas=True, client=client)
+    df2 = civis.io.read_civis(
+        table="datascience.iris", database=database, use_pandas=True,
+        client=client
+    )
+    assert df1.shape == (150, 6)
+    pd.testing.assert_frame_equal(df1, df2)
```
keithing: is it possible the ordering to the two tables comes back differently from the database and this results in `false`?

mheilman: oh, good point.  i haven't seen that, but i'll add an order by or sort after downloading

mheilman: @patr1ckm, the r client test does the same thing [here](https://github.com/civisanalytics/civis-r/blob/master/tools/integration_tests/smoke_test.R#L29), right?

PR #217, opened 2018-01-31
BUG Fix multipart upload log strings

stephen-hoover:
Without this "%s", we see an error message for every fragment of multipart uploads. Also correct the order of two values in a multipart upload debug emit.


civis/io/_files.py
```diff
@@ -122,7 +122,7 @@ def _multipart_upload(buf, name, file_size, client, **kwargs):
     num_parts = int(math.ceil((file_size) / float(part_size)))
 
     log.debug('Uploading file with %s bytes using %s file parts with a part '
-              'size of %s bytes', file_size, part_size, num_parts)
+              'size of %s bytes', file_size, num_parts, part_size)
```
keithing: Nice find.

PR #220, opened 2018-02-01
Adding robust parsing

### LINKED ISSUE #219: Parsing <schema.table> correctly where schema/table may contain periods
Apparently, the period is a valid character in a (standard) identifier on Redshift ([docs](https://docs.aws.amazon.com/redshift/latest/dg/r_names.html)). This creates parsing issues when there's a period in either "schema" or "table" for a given table `schema.table` and the Civis Python client attempts to parse out "schema" and "table" with code like this:

```python
schema, table = name.split('.')
```

It looks like there are three places in the Civis Python client code base where this strategy is used:

https://github.com/civisanalytics/civis-python/blob/5becc2437d80a68b2f1e29a239a084efed30dda1/civis/civis.py#L211
https://github.com/civisanalytics/civis-python/blob/4b94eb8f1d8208d4b9af9b7a8a13bcb71f440c6d/civis/io/_tables.py#L632
https://github.com/civisanalytics/civis-python/blob/4b94eb8f1d8208d4b9af9b7a8a13bcb71f440c6d/civis/io/_tables.py#L813

I've confirmed that one can successfully run SQL queries on Civis Platform for a table whose name is in the form of `foo."bar.baz"` (with double quotes as the necessary delimiters because of a period in the table `bar.baz`). We should update the client to handle periods (and delimiters) correctly.

keithing:
Fixes #219.  This PR allows users to pass in tables that contain periods in the schema and/or tablename, so long as they are in double quotes (as is the SQL convention).  For instance `schema."my.table"` or `"sch.ma"."tab.le"`.

If a user passes `table` without a schema, there will now be a more helpful error as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
+- Added more robust parsing for tablename parsing in io.  You may now
+  pass in tables like schema."tablename.with.periods".
 
 ## 1.8.1 - 2018-02-01
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-### Changed
+### Fixed
 - Added more robust parsing for tablename parsing in io.  You may now
   pass in tables like schema."tablename.with.periods".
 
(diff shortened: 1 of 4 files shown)


civis/io/_tables.py
```diff
@@ -973,3 +973,12 @@ def callback(future):
             return _download_file(url, filename, headers, compression)
 
     return callback
+
+
+def _robust_schema_table_split(table):
+    schema_name_tup = next(csv.reader(StringIO(table), delimiter="."))
```
stephen-hoover: Wow, this is a great find! It might be worth explicitly setting some other parameters instead of leaving them to the defaults. It looks like `doublequote` and `quotechar` are the two most important.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
```
stephen-hoover: Fixed?

PR #224, opened 2018-02-14
DOC Document civis.find and civis.find_one

### LINKED ISSUE #170: document the civis.find function
The `civis.find` function currently has no documentation. That's all.

jacksonllee:
This PR resolves #170.

### DIFF
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -18,8 +18,8 @@ def find(object_list, filter_func=None, **kwargs):
     Parameters
     ----------
     object_list : iterable
-        An iterable of arbitrary objects, particularly those with attributes
-        that can be targeted by the filters in ``kwargs``.
+        An iterable of arbitrary objects, particularly dicts and objects with
+        attributes that can be targeted by the filters in ``kwargs``.
     filter_func : callable, optional
         A one-argument function. If specified, ``kwargs`` are ignored.
         An ``object`` from the input iterable is kept in the returned list
@@ -38,6 +38,15 @@ def find(object_list, filter_func=None, **kwargs):
     Returns
     -------
     list
+
+    Examples
+    --------
+    >>> import civis
+    >>> client = civis.APIClient()
+    >>> # creds is a list of dicts representing credentials
+    >>> creds = client.credentials.list()
+    >>> # target_creds contains dicts with the key-value ('name', 'username')
+    >>> target_creds = find(creds, name='username')
 
     See Also
     --------
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -30,7 +30,7 @@ def find(object_list, filter_func=None, **kwargs):
         in conjunction with ``filter_func``. All keys must be strings.
         For an ``object`` from the input iterable to be included in the
         returned list, all the ``key``s must be attributes of ``object``, plus
-        one of the following conditions for a given ``key``:
+        any one of the following conditions for a given ``key``:
         - ``value`` is a one-argument function and
           ``bool(value(getattr(object, key)))`` is ``True``
         - ``value`` is ``True``
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -28,13 +28,13 @@ def find(object_list, filter_func=None, **kwargs):
     **kwargs
         Key-value pairs for more fine-grained filtering; they cannot be used
         in conjunction with ``filter_func``. All keys must be strings.
-        An ``object`` from the input iterable is kept in the returned
-        list, unless one of the following conditions is met:
-        - ``key`` is not an attribute of ``object``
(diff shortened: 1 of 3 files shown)


civis/civis.py
```diff
@@ -13,6 +13,34 @@
 
 
 def find(object_list, filter_func=None, **kwargs):
+    """Return the elements from ``object_list`` that satisfy the filters.
```
mheilman: An example would be nice.  It doesn't seem very clear what  `object_list : iterable` means as is.

jacksonllee: The function as-is seems pretty generic, and can take an arbitrary iterable of any objects. I'll try to come up with something and put it in the `Examples` section of the docstring.

jacksonllee:
It's turned out that the previous build failures were due to incompatibility between the latest PyPI releases of Sphinx (1.7.0) and numpydoc (0.7.0); Sphinx has introduced breaking changes very recently. As a workaround, I've forced Sphinx to be < 1.7.0. More at #226.

civis/civis.py
```diff
@@ -13,6 +13,45 @@
...
+        if and only if ``bool(filter_func(object))`` is ``True``.
+    **kwargs
+        Key-value pairs for more fine-grained filtering; they cannot be used
+        in conjunction with ``filter_func``. All keys must be strings.
+        An ``object`` from the input iterable is kept in the returned
+        list, unless one of the following conditions is met:
+        - ``key`` is not an attribute of ``object``
+        - ``value`` is a one-argument function and
+          ``bool(value(object[key]))`` is ``False``
+        - ``value`` is ``False``
```
stephen-hoover: Wow, this is confusing, and even seems buggy to me. This means that if `obj.spam = False`, and you run `civis.find([obj], spam=False)`, you get back an empty list, not the object for which `spam` is `False`.

If you provide a boolean `value`, you're nominally filtering on whether or not an object contains the `key` as an attribute. Except that's very different than what happens if you provide a non-boolean `value` (we check for equality). And it doesn't even work, because if you provide a key which isn't an attribute and the value `False`, then the object still isn't included. It looks like providing a `False` as the value to any kwarg will always result in an empty list being returned.

civis/civis.py
```diff
@@ -13,6 +13,45 @@
...
+        An ``object`` from the input iterable is kept in the returned list
+        if and only if ``bool(filter_func(object))`` is ``True``.
+    **kwargs
+        Key-value pairs for more fine-grained filtering; they cannot be used
+        in conjunction with ``filter_func``. All keys must be strings.
+        An ``object`` from the input iterable is kept in the returned
+        list, unless one of the following conditions is met:
+        - ``key`` is not an attribute of ``object``
+        - ``value`` is a one-argument function and
+          ``bool(value(object[key]))`` is ``False``
```
stephen-hoover: I don't think this is correct. It's `bool(value(getattr(object, key)))`, isn't it?

civis/civis.py
```diff
@@ -13,6 +13,45 @@
...
+        - ``value`` is ``False``
+        - ``object[key]`` is not equal to ``value``
```
stephen-hoover: It's looking for attributes, not items.

civis/civis.py
```diff
@@ -13,6 +13,45 @@
 
 
 def find(object_list, filter_func=None, **kwargs):
+    """Return the objects from ``object_list`` that satisfy the filters.
+
+    Parameters
+    ----------
+    object_list : iterable
+        An iterable of arbitrary objects, particularly dicts and objects with
```
stephen-hoover: This function won't filter dictionaries at all. It looks for attributes. That's one of the more confusing things about it, especially with the way the `Response` is written.

civis/civis.py
```diff
@@ -13,6 +13,45 @@
...
+
+    Returns
+    -------
+    list
+
+    Examples
+    --------
+    >>> import civis
+    >>> client = civis.APIClient()
+    >>> # creds is a list of dicts representing credentials
```
stephen-hoover: It's a list of `Response` objects. That's an important distinction, because this example won't work on a list of dictionaries.

civis/civis.py
```diff
@@ -13,6 +13,45 @@
...
+    >>> creds = client.credentials.list()
+    >>> # target_creds contains dicts with the key-value ('name', 'username')
```
stephen-hoover: "dict" -> "``civis.response.Response``s"
"key-value ('name', 'username')" -> "attribute 'name' == 'username'"

jacksonllee:
Ah, I was probably thrown off by the fact that the credentials "look" like dicts -- I'll remove the dict language. Also +1 for using positive language for how filtering works if possible. Updates are coming.

PR #225, opened 2018-02-14
ENH Public schema/tablename-splitting function

### LINKED ISSUE #223: Make _robust_schema_table_split public
The private function `civis.io._tables._robust_schema_table_split` is pretty useful. I find myself needing to do that in other pieces of code. I suggest making it public, perhaps with a name like `split_schema_table`. We may also want to handle the case where a table name is provided without a schema, e.g. for a temporary table.

stephen-hoover:
The `_robust_schema_table_split` is pretty useful, and users may wish to use it themselves. Promote to a public function with a clearer name. Also modify the function so that a single table name, without a schema, is a valid input. In such a case, return `None` for the schema name.

Closes #223 .

I'm open to name suggestions, or if this function should live in a different namespace. Or if you think that the function shouldn't be public after all.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1015,4 +1015,4 @@ def split_schema_tablename(table):
         raise ValueError("Cannot parse schema and table. "
                          "Does '{}' follow the pattern 'schema.table'?"
                          .format(table))
-    return schema_name_tup
+    return tuple(schema_name_tup)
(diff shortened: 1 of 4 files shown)


civis/io/_tables.py
```diff
@@ -632,7 +632,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
         to_csv_kwargs = {'encoding': 'utf-8', 'index': False}
         to_csv_kwargs.update(kwargs)
         df.to_csv(tmp_path, **to_csv_kwargs)
-        _, name = _robust_schema_table_split(table)
+        _, name = split_schema_tablename(table)
```
keithing: I think this needs to be updated in `civis/civis.py` as well

stephen-hoover:
This is ready to go, but it's blocked on the Sphinx version update in #224 .

PR #230, opened 2018-02-16
DOC Update the dependency set-up for building the docuemntation

jacksonllee:
The version of numpydoc as pinned down through git-submodules is outdated (circa 2015), and it is what conflicts with the latest PyPI release of Sphinx; this out-of-date version of numpydoc imports `sphinx.util.compat` which has just been removed from Sphinx. This caused the recent build failures in test building the documentation ([example](https://travis-ci.org/civisanalytics/civis-python/builds/341720752?utm_source=github_status&utm_medium=notification)). This PR removes git submodules and switches to pip-installing the dependencies for building the documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,6 +75,12 @@ to see if it will succeed.
 Build Documentation Locally
 ---------------------------
 
+To install dependencies for building the documentation::
+
+    pip install Sphinx
+    pip install sphinx_rtd_theme
+    pip install numpydoc
+
 To build the API documentation locally::
 
     cd docs
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,6 +14,7 @@
 # serve to show the default.
 
 import os
+import datetime
 
 # -- General configuration ------------------------------------------------
 
@@ -53,8 +54,9 @@ source_suffix = '.rst'
 master_doc = 'index'
 
 # General information about the project.
+current_year = datetime.datetime.now().year
 project = 'Civis Client'
-copyright = '2016, Civis Analytics'
+copyright = '2016-%d, Civis Analytics' % current_year
 author = 'Civis Analytics'
 
 # The version info for the project you're documenting, acts as replacement for
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -13,16 +13,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-import sys
 import os
-import shlex
-
-# If extensions (or modules to document with autodoc) are in another directory,
-# add these directories to sys.path here. If the directory is relative to the
-# documentation root, use os.path.abspath to make it absolute, like shown here.
-sys.path.insert(0, os.path.abspath('../sphinxext'))
-themedir = os.path.join(os.pardir, 'sphinx_rtd_theme')
-sys.path.insert(0, themedir)
 
 # -- General configuration ------------------------------------------------
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -404,7 +404,7 @@ class GeneratedAutosummary(Autosummary):
             if public_only and name.startswith('_'):
                 continue
             try:
-                documenter = get_documenter(safe_getattr(obj, name), obj)
+                documenter = get_documenter(None, safe_getattr(obj, name), obj)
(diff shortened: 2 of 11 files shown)


jacksonllee:
@keithing Will removing the git submodules and switching to pip install-ing the doc-related dependencies impact the doc building process on the readthedocs site?

keithing:
> Will removing the git submodules and switching to pip install-ing the doc-related dependencies impact the doc building process on the readthedocs site?

We should be okay so long as `dev-requirements.txt` and `docs/source/conf.py` are set up correctly.

jacksonllee:
The builds have passed. I've also tried to build the documentation locally, and everything worked and looked as expected.

README.rst
```diff
@@ -75,6 +75,12 @@ to see if it will succeed.
 Build Documentation Locally
 ---------------------------
 
+To install dependencies for building the documentation::
+
+    pip install Sphinx
+    pip install sphinx_rtd_theme
+    pip install numpydoc
+
```
keithing: since we have specific version requirements, it might be easier to suggest `pip install dev-requirements.txt` here.

docs/source/conf.py
```diff
@@ -413,7 +406,7 @@ def get_members(obj, typ, public_only=True):
             if public_only and name.startswith('_'):
                 continue
             try:
-                documenter = get_documenter(safe_getattr(obj, name), obj)
+                documenter = get_documenter(None, safe_getattr(obj, name), obj)
```
keithing: What is this change fore?

README.rst
jacksonllee: I thought about that, but `dev-requirements.txt` would also install dependencies for running tests, which this section of the README isn't about.

jacksonllee: Or maybe we should have a "Local Development" instead? That way, we can point to `dev-requirements.txt` and explain how to run tests and build the doc locally.

docs/source/index.rst
```diff
@@ -98,12 +98,12 @@ Then add the following line, replacing ``api_key_here`` with your key::
...
 **Linux**
 
-Open ``.bash_profile`` in your favorite editor (nano is used here):
+Open ``.bashrc`` in your favorite editor (nano is used here):
 
 .. code-block:: bash
 
    cd ~/
-   nano .bash_profile
+   nano .bashrc
```
keithing: we mention `.bash_profile` on the README, so we should make this consistent with that.  I don't have a preference for one or the other.

README.rst
keithing: I thinking adding a section on local development is out of scope for this.  I can see how the irrelevant stuff in `dev-requirements.txt` might make it hard to install so we can leave this as is.

docs/source/index.rst
keithing: Actually, if this is specific to Linux it's fine if the README talks about .bash_profile (which might be more common on Mac?)

docs/source/conf.py
jacksonllee: AFAICT, there's no behavioral change here. Sphinx v1.7.0 has changed the function signature of `get_documenter`. It now requires three positional args:

https://github.com/sphinx-doc/sphinx/blob/5558b63315438cfaba1d6f45e1ce59da8551f94a/sphinx/ext/autosummary/__init__.py#L163

...compared to < v1.7.0:

https://github.com/sphinx-doc/sphinx/blob/645044d7f03c7fcb4c39a8edf40aa66f6d6c2b64/sphinx/ext/autosummary/__init__.py#L160

If we set the new arg `app` as `None`, it will have the same effect as < v1.7.0. `app` is passed into `get_documenters` (note it's another function):

https://github.com/sphinx-doc/sphinx/blob/5558b63315438cfaba1d6f45e1ce59da8551f94a/sphinx/ext/autosummary/__init__.py#L190

https://github.com/sphinx-doc/sphinx/blob/5558b63315438cfaba1d6f45e1ce59da8551f94a/sphinx/ext/autodoc/__init__.py#L1491-L1497

...compared to < v1.7.0:

https://github.com/sphinx-doc/sphinx/blob/645044d7f03c7fcb4c39a8edf40aa66f6d6c2b64/sphinx/ext/autosummary/__init__.py#L188

PR #231, opened 2018-02-16
DOC: Adding missing function documentation

keithing:
Looks like we forgot to add a line to the docs about this function.  I imagine there is a way to programmatically add all function from a namespace, we might want to look into that in the future.


keithing:
> What if we pointed the docs at the private modules rather than `civis.io`

That seems like a good idea.

PR #234, opened 2018-02-20
WIP: Adding in cache invalidation for get_api_spec

### LINKED ISSUE #233: Caching confusion in the client
Heard of a use case recently where a user had a notebook instance open for a while and ran into an issue with caching.  The user called `APIClient(resources='all')` but did not see new endpoints they expected to see (because they already called it earlier and the results were cached before the endpoints were created).

My impression is that it's not trivial to refresh `lru_cache` in smart way to avoid this and since this is a rare occurrence, we should address it with documentation.

keithing:
This adds a timer to the api spec cache in `generate_classes`.  Effectively this means that calls to `APIClient()` will now poll for a new api spec if the call is made 10 hours after the first call.

Resolves #233.

TODO:

- [ ] Pass integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+### Changed
+- The caches used in generate_classes and get_api_spec will now invalidate
+  after a period of time to prevent stale API specs in long running Python
+  instances (after 10 hours and 5 hours, respectively).
+
 ### Fixed
 - Added more robust parsing for tablename parsing in io.  You may now
   pass in tables like schema."tablename.with.periods".
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -8,6 +8,7 @@ try:
 except ImportError:
     from funcsigs import Signature, Parameter
 
+import cachetools
 from jsonref import JsonRef
 import requests
 import six
@@ -63,6 +64,11 @@ ITERATOR_PARAM_DESC = (
 MAX_RETRIES = 10
 CACHED_SPEC_PATH = os.path.join(os.path.expanduser('~'),
                                 ".civis_api_spec.json")
+
+# Note: time to live (ttl) for get_api_spec should be shorter than the
+# cache for generate_classes so generate_classes can return fresh api spec
+GENERATE_CLASSES_TTL = 36000  # 10 hours
+GET_API_SPEC_TTL = GENERATE_CLASSES_TTL / 2
 
 
 def exclude_resource(path, api_version, resources):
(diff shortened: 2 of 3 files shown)


keithing:
@stephen-hoover, before I go any further with this

1. Do you think it's worth adding this dependency / change?  FWIW, `cachetools` is pure python, has no dependencies and is python 2 supported.
2. Do you think 10 hours is a reasonable time for validation?

stephen-hoover:
Hm. I'm not keen on adding a new dependency, especially for an edge case like this. You can clear an `@lru_cache` with the `cache_clear` method. You could either add a parameter to the `APIClient` constructor which invalidates the cache, or write your own version of `ttl_cache`. It doesn't seem like you'd need a lot of lines of code. Would it be enough to add a "clear_cache" argument to `APIClient`? That would be simplest, and gives more control to users.

keithing:
>  I'm not keen on adding a new dependency, especially for an edge case like this.

That's a good point.

> ... write your own version of ttl_cache. It doesn't seem like you'd need a lot of lines of code. 

I'm not positive it would be simple, especially around threading in `CivisFuture` and Python 2 compatibility.  @mheilman mentioned this, too, but I think we should be careful not to introduce a bug in the process of addressing a rare problem.  I'm happy to look at a solution, but it would have to be very concise / simple in order to warrant maintaining it going forward.

> Would it be enough to add a "clear_cache" argument to APIClient

That doesn't exactly address the issue that users will be confused if they can't access endpoints they know they have access to (since their `APIClient` was cached).  I'm okay with that, however, if adding `cachetools` as a dependency is too heavy.

stephen-hoover:
I pictured an implementation like
```python
def expiring_lru_cache(time_limit, *args, **kwargs):


  def timed_cache(func):
    cached_func = lru_cache(*args, **kwargs)(func)
    cache_time = time.time()


    @functools.wraps(cached_func)
    def wrapped_func(*args, **kwargs):
      if time.time() - cache_time > time_limit:
        cached_func.clear_cache()
        cache_time = time.time()
      return cached_func(*args, **kwargs)
    return wrapped_func
  return timed_cache
```
We may not want to maintain even that much. What's your concern around threading? The solutions we've been discussing would only address newly-created objects, which I think is the correct scope.

keithing:
> I pictured an implementation like ... we may not want to maintain even that much. 

Nice!  That is very succinct.  I saw the implementation in `cachetools` and I assumed it would be more complicated.

https://github.com/tkem/cachetools/blob/master/cachetools/ttl.py#L57
https://github.com/tkem/cachetools/blob/master/cachetools/func.py#L29

> What's your concern around threading

I noticed the `RLock` calls in the `cachetools` implementation and assumed there was some race condition or other problem I haven't thought of it yet.

stephen-hoover:
The `cachetools` implementation does look like a lot more than we need. It scared me when I first looked at it too. The implementation I suggested voids the entire cache after the expiration time, rather than expiring on a per-key basis. I think that's fine for the API client. 

`cachetools` looks like it's making its caches thread-safe. I don't think we need to worry about that, but it would be possible to include a lock. I don't expect that we'd be creating client objects across different threads, but a lock around the cache-clearing step shouldn't add many lines. A `threading.RLock` is the right thing to use there.

keithing:
I'm going to close this for now and circle back with a different implementation later.

PR #236, opened 2018-02-23
ENH Include parent job and run ID in children

stephen-hoover:
It can be useful for child jobs to know which job created them. This injects environment variables into child jobs which give the job and run ID of the creator. Note that the joblib backend uses the `_ContainerShellExecutor` to launch its jobs, so this change affects it as well. Custom scripts will only have parent job and run IDs if the template provides that parameter. There's no way to inject environment variables which the template creator didn't explicitly allow for.

### DIFF
--- a/civis/parallel.py
+++ b/civis/parallel.py
@@ -60,6 +60,13 @@ def infer_backend_factory(required_resources=None,
     relevant parameters (e.g. the Docker image) of the container
     it's running inside of.
 
+    Jobs created through this backend will have environment variables
+    "CIVIS_PARENT_JOB_ID" and "CIVIS_PARENT_RUN_ID" with the contents
+    of the "CIVIS_JOB_ID" and "CIVIS_RUN_ID" of the environment which
+    created them. If the code doesn't have "CIVIS_JOB_ID" and "CIVIS_RUN_ID"
+    environment variables available, the child will not have
+    "CIVIS_PARENT_JOB_ID" and "CIVIS_PARENT_RUN_ID" environment variables.
+
     .. note:: This function will read the state of the parent
               container job at the time this function executes. If the
               user has modified the container job since the run started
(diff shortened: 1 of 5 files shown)


stephen-hoover:
@keithing , this still needs some documentation updates. If the code looks good, I'll move on to documentation.

keithing:
Yes, code looks good.

@patr1ckm for viz.

stephen-hoover:
@keithing , I added some documentation. Let me know if you think it needs more, or if what I wrote isn't clear. This PR is ready for a full review now.

civis/parallel.py
```diff
@@ -392,6 +406,13 @@ def make_backend_template_factory(from_template_id,
                                   hidden=True):
     """Create a joblib backend factory that uses Civis Custom Scripts.
 
+    If your template has settable parameters "CIVIS_PARENT_JOB_ID" and
+    "CIVIS_PARENT_RUN_ID", then this executor will fill them with the contents
+    of the "CIVIS_JOB_ID" and "CIVIS_RUN_ID" of the environment which
+    created them. If the code doesn't have "CIVIS_JOB_ID" and "CIVIS_RUN_ID"
+    environment variables available, the child will not have
+    "CIVIS_PARENT_JOB_ID" and "CIVIS_PARENT_RUN_ID" environment variables.
```
keithing: This seems like a lot of documentation to copy in multiple places.  Is there a single natural place to put this rather than copying across multiple functions?

stephen-hoover: That's a fair criticism, but I don't know what single place would be best. Ideally, this will be a feature of Civis Platform more generally in the future, and users will expect it. For now, more places seems better. ... It occurs to me that users may not know that running scripts will always have a CIVIS_JOB_ID and CIVIS_RUN_ID. That /is/ a Platform feature. Should it be explained here?

keithing: I think we want to avoid documenting all Platform features here (and rely on autogenerated docs where possible).  If you can't think of a better single place to put it, it's fine by me to keep as is (and sometimes repeating is better than the wrong abstraction!).

stephen-hoover:
Oops, I almost forgot to add a changelog entry. Re-approve?

Issue #237, opened 2018-02-28
Add `query_to_civis_file` function


keithing:
@stephen-hoover I wonder if you've mentioned this before?  It seems like it might be useful to return the file_id from a SQL query without requiring the user to download the result (as in `read_civis_sql`).

cc @jkulzick

jkulzick:
I can see situations where just a file_id would be useful.

stephen-hoover:
I think that the easiest way to do this would be to make some sort of "run_sql" helper function available. That could mean promoting `civis.io._tables._sql_script` to a public function and changing its output to be a `CivisFuture`.

FWIW, this can be as few as two lines:
```python
sql_script = client.scripts.post_sql(...)
sql_fut = civis.utils.run_job(sql_script.id)
```
and the file ID is in `sql_fut.result().output[0].file_id`.

stephen-hoover:
This was added in #238 .

PR #238, opened 2018-03-02
Adding export_to_civis_file

keithing:
This adds the IO functionality mentioned in #237.  It might make sense to have this function tied to `read_civis_sql`, since we could essentially call this function inside `read_civis_sql`.  However, in that function we do some more complicated checks to force parallel unload which would be difficult to do in `query_to_civis_file`.  (that is, how can we add a header back to a file on s3?  In read_civis_sql we just pass the headers along to `pandas`).  I don't think this introduces too much repeated code and allows us to avoid adding deprecated parameters like `archive` and `api_key` to the function signature.

Some open questions:

* is `query_to_civis_file` the best name?
* Can we link to the autogenerated API functions in the docs? :func:`civis.APIClient.scripts.get_sql_runs` doesn't create a link.
* Do we want to unpack `csv_settings` dictionary?  That is, should we have `csv_settings` as a parameter or the multiple parameters `include_header`, `compression`, etc.   We're unpacking these parameters in `civis_to_csv`.


stephen-hoover:
I have a couple of global comments. One is that I don't think that `query_to_civis_file` is the best name. "Query" is a particular operation of Civis Platform, and this is using a SQL script, not a query. Looking through the functions we have already, my two suggestions are `read_civis_to_civis_file` or `export_to_civis_file`. `table_to_civis_file` would be great, but since it's not a straight dump of one table, that might be misleading.

My second comment is that I think the function should return a `CivisFuture`. This is an operation which really seems like it should be non-blocking, and we aren't reading any data into the client.

keithing:
Great suggestions @stephen-hoover.  I changed the name of the function to `export_to_civis_file` and am now returning a future instead.

civis/io/_tables.py
```diff
@@ -138,6 +140,70 @@ def read_civis(table, database, columns=None, use_pandas=False,
...
+    database : str or int
+        Execute the query against this database. Can be the database name
+        or ID.
+    job_name : str, optional
+        A name to give the job. If omitted, a random job name will be
+        used.
+    client : :class:`civis.APIClient`, optional
+        If not provided, an :class:`civis.APIClient` object will be
+        created from the :envvar:`CIVIS_API_KEY`.
+    credential_id : str or int, optional
```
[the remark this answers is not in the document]
keithing: This should be okay. A string works for me for `read_civis` which is also calling `_sql_script` internally.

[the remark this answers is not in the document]
keithing: (this is also the same documentation string used elsewhere for io functions)

keithing:
This is ready for another pass @stephen-hoover.  I did some gymnastics with the mocking library to avoid adding a cassette, let me know if you prefer that or going back to the cassette.

civis/tests/test_io.py
```diff
@@ -419,3 +433,24 @@ def test_split_schema_tablename_raises():
     s = "table.with.too.many.periods"
     with pytest.raises(ValueError):
         civis.io._tables.split_schema_tablename(s)
+
+
+@mock.patch.object(civis.io._tables, '_sql_script', autospec=True,
+                   return_value=[-1000, 1000])
```
stephen-hoover: I've done things like this myself in the past, but I've come to think that when you're mocking, it's best to use inputs and outputs which are valid for the function being mocked. The job ID and run ID are always going to be positive integers, so maybe you could use something like `return_value=[700, 1000]`?

civis/tests/test_io.py
```diff
@@ -419,3 +433,24 @@ def test_split_schema_tablename_raises():
...
+@mock.patch.object(civis.io._tables, 'APIClient', autospec=True)
```
stephen-hoover: Rather than patching the `APIClient` constructor, you could pass the mock client as an input parameter to `export_to_civis_file`.

civis/tests/test_io.py
```diff
@@ -419,3 +433,24 @@ def test_split_schema_tablename_raises():
...
+def test_export_to_civis_file(mock_APIClient, mock_sql_script):
+    expected = [{'file_id': 9844453}]
+
+    # Mock client.scripts.get_sql_runs to return successful response with data
+    # (mock_client must not have channels in its spec to avoid hitting PubNub)
```
stephen-hoover: You could try using the `civis.tests.mocks.create_client_mock` function to take care of the auto-speccing.

civis/tests/test_io.py
```diff
@@ -419,3 +433,24 @@ def test_split_schema_tablename_raises():
...
+    mock_client = mock.Mock(spec=['scripts'])
+    mock_client.scripts = mock.Mock(spec=['get_sql_runs'])
+    response = MockResponse(state='success', output=expected)
```
stephen-hoover: You should be able to use a real `Response` here and avoid creating a special object for testing.

civis/tests/test_io.py
```diff
@@ -419,3 +433,24 @@ def test_split_schema_tablename_raises():
...
+    mock_client.scripts.get_sql_runs.return_value = response
+    mock_APIClient.return_value = mock_client
+
+    sql = "SELECT 1"
+    fut = civis.io.export_to_civis_file(sql, 'fake-db',
+                                        polling_interval=POLL_INTERVAL)
+    data = fut.result()['output']
+    assert data == expected
```
stephen-hoover: Is it worth also checking that we called `mock_client.scripts.post_sql` with the input `sql`? That risks breaking if we come up with a different way to export SQL queries, but that would be a major version change of the API, so maybe it's okay.

stephen-hoover:
I think that the mocks work as well as cassettes, and avoid having to add and maintain a new cassette file. I prefer the test with mocks. Either way is tricky to really test properly.

civis/tests/test_io.py
keithing: I had forgotten about that!

civis/tests/test_io.py
keithing: Cool, that probably is easier to understand.

civis/tests/test_io.py
keithing: I added an assert on `_sql_script` rather than `post_sql` and another test for `_sql_script`, which I thought might be more in the spirit of unit testing.

keithing:
This should be ready to merge now!

PR #241, opened 2018-03-07
ENH Add "dvs_to_predict" param to ModelPipeline.predict

stephen-hoover:
CivisML v2.2 will add the ability for users to subset model predictions, as a way to save time and space. Add this parameter to `ModelPipeline.predict`.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -738,7 +738,7 @@ class ModelPipeline:
                  etl=None):
         self.model = model
         self._input_model = model  # In case we need to modify the input
-        if isinstance(dependent_variable, str):
+        if isinstance(dependent_variable, six.string_types):
             # Standardize the dependent variable as a list.
             dependent_variable = [dependent_variable]
         self.dependent_variable = dependent_variable
@@ -1133,7 +1133,8 @@ class ModelPipeline:
                 manifest=None, file_id=None, sql_where=None, sql_limit=None,
                 primary_key=SENTINEL, output_table=None, output_db=None,
                 if_exists='fail', n_jobs=None, polling_interval=None,
-                cpu=None, memory=None, disk_space=None):
+                cpu=None, memory=None, disk_space=None,
+                dvs_to_predict=None):
         """Make predictions on a trained model
 
         Provide input through one of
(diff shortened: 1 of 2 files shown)


civis/ml/_model.py
```diff
@@ -1219,6 +1220,13 @@ def predict(self, df=None, csv_path=None,
             RAM requested by the user for a single job.
         disk_space : float, optional
             disk space requested by the user for a single job.
+        dvs_to_predict : list of str, optional
+            If this is a multi-output model, you may list a subset of
+            dependent variables for which you wish to generate predictions.
+            This list must be a subset of the original `dependent_variable`
+            input. Ignoring some of the model's outputs will let predictions
```
elsander: Instead of this sentence, what about something like "Scores for all outputs will always be calculated, but only the subset will be returned, allowing predictions to complete faster and use less disk space". I think it's good to make this explicit so that users know that asking for a subset won't affect their scores.

civis/ml/_model.py
```diff
@@ -1265,6 +1273,12 @@ def predict(self, df=None, csv_path=None,
             predict_args['LIMITSQL'] = sql_limit
         if n_jobs:
             predict_args['N_JOBS'] = n_jobs
+        if dvs_to_predict:
+            if isinstance(dvs_to_predict, six.string_types):
+                dvs_to_predict = [dvs_to_predict]
+            if self.predict_template_id > 10600:
```
elsander: This is really nitpicky, but it would be clearer if this were 10583 instead of 10600, to match other parts of the code that reference the specific template id that is the cutoff for that argument.

stephen-hoover: Good point. Changed.

civis/ml/_model.py
stephen-hoover: It's true that we do currently let the model compute everything, then drop columns from the output array before writing to disk. But that seems like an implementation detail. If we were to somehow modify the models such that they only computed the requested labels, it would still look the same to the user. How about 

> The scores for the returned subset will be identical to the scores which those outputs would have had if all outputs were written, but ignoring some of the model's outputs will let predictions complete faster and use less disk space.

elsander: Good point, that is more of an implementation detail. I like your revision.

PR #242, opened 2018-03-09
ENH Register models trained outside of Civis Platform

stephen-hoover:
If you train a scikit-learn compatible estimator outside of Civis Platform, you can use this to upload it to Civis Platform and prepare it for scoring with CivisML. There's a new Custom Script which will introspect metadata necessary for CivisML and make itself appear sufficiently like a CivisML training job that it can be used as input to a scoring job.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -42,8 +42,11 @@ _PRED_TEMPLATES = {10582: 10583,  # v2.1
                    9112: 9113,    # v1.1
                    8387: 9113,    # v1.0
                    7020: 7021,    # v0.5
+                   11028: 10616,  # v2.2 registration CHANGE ME
                    }
 _CIVISML_TEMPLATE = None  # CivisML training template to use
+REGISTRATION_TEMPLATES = [11028,  # v2.2 CHANGE ME
+                          ]
 
 
 class ModelError(RuntimeError):
@@ -631,10 +634,10 @@ class ModelPipeline:
         See :func:`~civis.resources._resources.Scripts.post_custom` for
         further documentation about email and URL notification.
     dependencies : array, optional
-        List of packages to install from PyPI or git repository (i.e., Github
+        List of packages to install from PyPI or git repository (e.g., Github
         or Bitbucket). If a private repo is specified, please include a
         ``git_token_name`` argument as well (see below). Make sure to pin
-        dependencies to a specific version, since dependecies will be
+        dependencies to a specific version, since dependencies will be
         reinstalled during every training and predict job.
     git_token_name : str, optional
         Name of remote git API token stored in Civis Platform as the password
@@ -713,6 +716,8 @@ class ModelPipeline:
         global _CIVISML_TEMPLATE
         if _CIVISML_TEMPLATE is None:
             for t_id in sorted(_PRED_TEMPLATES)[::-1]:
+                if t_id in REGISTRATION_TEMPLATES:
+                    continue
                 try:
                     # Check that we can access the template
                     client.templates.get_scripts(id=t_id)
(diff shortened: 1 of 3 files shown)


stephen-hoover:
The template numbers for registration jobs should be changed when we release CivisML v2.2 and before the next civis-python release. Putting the placeholders in there seemed like the best way to make sure we remember to do it.

civis/ml/_model.py
```diff
@@ -783,6 +788,147 @@ def __setstate__(self, state):
...
+        of feature names, CivisML will have to assume that your scoring
+        table contains only the features needed for scoring (perhaps also
+        with a primary key column), in all in the correct order.
+
+        Parameters
+        ----------
+        model : sklearn.base.BaseEstimator or int
+            The model object. This must be a fitted scikit-learn compatible
+            Estimator object, or else the integer Civis File ID of a
+            pickle which stores such an object.
```
elsander: Does this have to be joblib-serialized? If so, we should mention that.

civis/ml/_model.py
```diff
@@ -783,6 +788,147 @@ def __setstate__(self, state):
...
+        dependent_variable : string or List[str], optional
+            The dependent variable of the training dataset.
+            For a multi-target problem, this should be a list of
+            column names of dependent variables.
+        features : string or List[str], optional
+            A list of column names of features which were used for training.
+            These will be used to ensure that tables input for prediction
+            have the correct features in the correct order.
+        primary_key : string, optional
+            The unique ID (primary key) of the scoring dataset..
```
elsander: Stray period here.

civis/ml/tests/test_model.py
```diff
@@ -39,6 +39,10 @@
 from civis.ml import _model
 
 
+LATEST_TRAIN_TEMPLATE = 10582
```
elsander: I'm not sure I understand what this change is for. It seems like this is another constant that we'll have to remember to update.

docs/source/ml.rst
```diff
@@ -61,7 +61,7 @@ A :class:`~sklearn.pipeline.Pipeline` allows you to combine multiple
 modeling steps (such as missing value imputation and feature selection) into a
 single model. The :class:`~sklearn.pipeline.Pipeline` is treated as a unit -- for example,
 cross-validation happens over all steps together.
- 
+
 You can define your model in two ways, either by selecting a pre-defined algorithm
```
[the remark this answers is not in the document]
stephen-hoover: What were you picturing here? This section is talking about training models; registering a model trained outside of Platform bypasses this step.

civis/ml/_model.py
stephen-hoover: The `joblib.load` will work on regular pickles. I added a clause about joblib; if you think it's confusing I can change it or take it out.

civis/ml/_model.py
stephen-hoover: Thanks. I think both of these are stray since it's not a complete sentence.

civis/ml/tests/test_model.py
stephen-hoover: Tests should fail if we don't update it, which makes it much easier to remember. The problem I had was that we'd used `max(_PRED_TEMPLATES)` to infer what I've set here as `LATEST_TRAIN_TEMPLATE`. With the way I've implemented model registration, we're no longer guaranteed that this `max` operation gives us the most recent template for training jobs. I could build in logic which find the maximum of the keys in `_PRED_TEMPLATES` which aren't also in `REGISTRATION_TEMPLATES`. I started to be concerned that adding extra logic like that would make the tests more reliable. Do you think it's worth writing a `get_latest_train_template` function for the tests?

docs/source/ml.rst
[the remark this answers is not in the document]
elsander: I was just imagining a sentence like "If you have already trained a model and wish to score it using CivisML, see [section link]." My goal is to make it so that the documentation will lead you to the right place; the section title "defining a model" could be interpreted as providing the definition of a model that's already trained, and since it's an early section, users are likely to start there. You don't have to include it if you don't want to, but users sometimes struggle to find things in our documentation, so I'm trying to find ways to make that easier.

civis/ml/tests/test_model.py
elsander: Thanks for the clarification. I think this change is fine; an obvious constant we have to change is better than having to debug a less obvious helper function.

PR #243, opened 2018-03-13
default to "all" resources

mheilman:
I don't think there's a reason to default to "base" anymore.  I believe we talked about how it'd be nice to have "all" be the default so that people didn't have to remember to set it explicitly.

I'm not sure whether this would technically be a breaking change.  My sense is that it isn't because it only affects internals (e.g., whether pubnub is used).


stephen-hoover:
Is there a reason to keep using "resources" at all? Removing it entirely would be a breaking change, but we could deprecate the parameter and remove the code which switches between the two different sets of endpoints.

keithing:
I agree with @stephen-hoover.  If you don't mind going the extra step and deprecating the parameter that would be really helpful.  It could get very noisy if we add a warning if resources is set, so I'm ambivalent on whether we just document that the parameter is deprecated or add a warning.  We'll also need to pull out all the places where we call `resources='all'` in the source code.

mheilman:
I think it's probably best to include the deprecation warning.  We can think of it as incentive to release 2.0.

I had to move `_deprecation.py` because importing it from `util/` was causing a loop.

stephen-hoover:
What I was thinking was to ignore the `resources` input everywhere and also deprecate it in downstream functions such as `generate_classes_maybe_cached`. It still needs to be part of the function signature of public functions, but I don't think it needs to actually do anything.

mheilman:
I tried to deprecate and remove `resources` references as appropriate.  Please let me know if I missed anything.

I originally didn't realize that that argument existed in other public functions.   Thanks for catching that.

keithing:
@mheilman it looks like there are some python 2 errors (which you probably have seen).  I'm assigning to keep track of where this is (and will take another look after that is resolved)

civis/civis.py
```diff
@@ -338,10 +342,17 @@ def __init__(self, api_key=None, return_type='snake',
...
+        # the functions it calls until the `resources` argument is removed.
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore",
+                category=FutureWarning,
+                module='civis')
+            classes = generate_classes_maybe_cached(local_api_spec,
+                                                    session_auth_key,
+                                                    api_version,
+                                                    resources)
```
keithing: Is this an interim fix?  Why not set a default argument to all of the functions with `resources` like `parse_api_spec` and `generate_classes_maybe_cached` and then in functions like this we simply don't pass along the `resources` parameters?  I don't believe that's not a breaking change since `resources='base'` is a subset of `resources='all'`?

mheilman: I was thinking we could just take it out when we do 2.0 rather than fiddling with it now and then taking it out later.

mheilman:
rebased

civis/civis.py
keithing: Ah, okay I see what you meant above now

mheilman:
good catch!

civis/_deprecation.py
```diff
@@ -1,7 +1,7 @@
 from functools import wraps
 import warnings
 
-from civis.compat import signature
+from .compat import signature
```
stephen-hoover: Why change this line? If the tests pass, then it should be okay, but there's some differences in how relative imports are handled between Python 2 and 3, and explicit imports tend to be more robust.

mheilman: The previous way led to a test failure because of adding the import for `deprecate_param` to `civis.py`.  I believe the issue is that this module is imported by `civis.py`, but then the `from civis.compat` version causes `civis.py` to be imported again.  The relative import doesn't lead to a cycle for some reason.


From [travis](https://travis-ci.org/civisanalytics/civis-python/builds/353073967), before I changed the line in question:

```
Traceback:
civis/__init__.py:4: in <module>
    from civis.civis import APIClient, find, find_one
civis/civis.py:7: in <module>
    from civis.resources import generate_classes_maybe_cached
civis/resources/__init__.py:1: in <module>
    from ._resources import (generate_classes,
civis/resources/_resources.py:17: in <module>
    from civis._deprecation import deprecate_param
civis/_deprecation.py:4: in <module>
    from civis.compat import signature
E   ImportError: No module named compat
```

mheilman: @stephen-hoover, does that answer your question?

stephen-hoover: Yes, thank you!

PR #244, opened 2018-03-14
BUG Include JSONs in package data

stephen-hoover:
We want to include JSON files with installations of the API client; there are JSON blobs with caches of API endpoints which are useful for testing. The `data_files` argument installs the JSON files when running `python setup.py install`, but they don't come when you run `pip install`. Setting `include_package_data=True` will include everything in the `MANIFEST.in`, including the JSON files.


stephen-hoover:
Re-reading the docs, I think you're correct that `include_package_data=True` is the right way to go. Changed.

PR #249, opened 2018-04-16
BUG In Travis, restrict pandas version downloaded

stephen-hoover:
Pandas dropped Python 3.4 support with v0.21, and our Python 3.4 tests have recently begun failing when they attempt to install the newest version. Our Travis tests were installing all of the optional packages in a separate line; move those to their own requirements file so that we can control which version is intalled where. See also https://github.com/pandas-dev/pandas/issues/20697 .


full-testing-requirements.txt
```diff
@@ -0,0 +1,7 @@
+feather-format
+numpy
+pandas; python_version >= '3.5'
+pandas<0.21; python_version == '3.4'
+pandas<=0.23; python_version == '2.7'
+scikit-learn
+scipy
```
keithing: Why not add these to `dev-requirements.txt`?

stephen-hoover:
The separate file did seem strange, but I couldn't remember why we didn't put those extra requirements in dev-requirements to begin with. I moved them.

PR #252, opened 2018-05-03
DOC: Windows api key

keithing:
This add instructions for adding the environmental variable `CIVIS_API_KEY` in Windows.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -48,12 +48,12 @@ Windows
 ~~~~~~~
 
 1. Navigate to ``Settings`` -> type "environment" in search bar ->
-   ``Edit environment variables for your account.``. This can also be found
+   ``Edit environment variables for your account``. This can also be found
    in ``System Properties`` -> ``Advanced`` -> ``Environment Variables...``.
 2. If ``CIVIS_API_KEY`` already exists in the list of environment variables,
-   click on it and press ``Edit...``. Otherwise, click ``New..```.
-3. Enter CIVIS_API_KEY as the ``Variable name``
-5. Enter your API key as the ``Variable value``.  Your API key should look
+   click on it and press ``Edit...``. Otherwise, click ``New..``.
+3. Enter CIVIS_API_KEY as the ``Variable name``.
+4. Enter your API key as the ``Variable value``.  Your API key should look
    like a long string of letters and numbers.
 
 
--- a/README.rst
+++ b/README.rst
@@ -25,28 +25,66 @@ requests directly to the Civis API. See the
 `full documentation <https://civis-python.readthedocs.io>`_ for more details.
 
 
-Installation
-------------
+API Keys
+--------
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -25,28 +25,67 @@ requests directly to the Civis API. See the
...
+
+1. Add the following to ``.bash_profile`` for bash::
 
     export CIVIS_API_KEY="alphaNumericApiK3y"
 
-4. Source your ``.bash_profile``
-5. Install the package::
+2. Source your ``.bash_profile`` (or restart your terminal).
+
+Windows
```
stephen-hoover: I think this applies to Windows 10 and possibly Windows 7. Is XP the same?

README.rst
```diff
@@ -25,28 +25,67 @@ requests directly to the Civis API. See the
...
+~~~~~~~
+
+1. Navigate to ``Settings`` -> type "environment" in search bar ->
+   ``Edit environment variables for your account``. This can also be found
+   in ``System Properties`` -> ``Advanced`` -> ``Environment Variables...``.
+2. If ``CIVIS_API_KEY`` already exists in the list of environment variables,
```
stephen-hoover: I believe there are two lists of environment variables, "User variables" and "System variables", yes? I think we should tell users to add their API key to the user variables.

README.rst
```diff
@@ -25,28 +25,67 @@ requests directly to the Civis API. See the
...
+
+You can test your installation by running
+
+.. code-block:: python
+
+    import civis
+    client = civis.APIClient()
+    print(client.users.list_me()['username'])
+
+If ``civis-python`` was installed correctly, this will print your Civis
```
stephen-hoover: "Civis" -> "Civis Platform"

README.rst
```diff
@@ -25,28 +25,67 @@ requests directly to the Civis API. See the
 `full documentation <https://civis-python.readthedocs.io>`_ for more details.
 
 
-Installation
-------------
+API Keys
+--------
+
+Usage of ``civis-python`` requires a valid Civis API key, which can be created
```
stephen-hoover: "Civis API key" -> "Civis Platform API key"?

README.rst
keithing: Agreed.  I'll add a note.

README.rst
```diff
@@ -25,28 +25,68 @@ requests directly to the Civis API. See the
...
+
+1. Add the following to ``.bash_profile`` for bash::
 
     export CIVIS_API_KEY="alphaNumericApiK3y"
 
-4. Source your ``.bash_profile``
-5. Install the package::
+2. Source your ``.bash_profile`` (or restart your terminal).
+
+Windows 10
```
mheilman: It might be worth mentioning https://docs.microsoft.com/en-us/windows/wsl/install-win10, which in my opinion is the best way to use python in windows.

keithing: I suspect users will already have a workflow for python on Windows before they install `civis`. This is pretty cool, though.

PR #256, opened 2018-05-16
Ensure DataFrame import successfully occurs

ffaristocrat:
This is related to this ongoing JIRA

https://civisanalytics.atlassian.net/browse/CIVP-12730

Platform import simply can't handle files that have a `/"` anywhere in the file.

If a double quote is escaped with a backslash, the import fails.
If a file has fields enclosed by double quotes and any field ends with a `/`, the import fails.

This PR does two things to ensure a DataFrame will be succesfully imported via API:
1. Cleans dataframe fields by inserting a space in between a backslash and double quotesand inserts a space at the end of any string that ends with a backslash.
2. Exports the DataFrame to CSV with | as the delimiter, doublequotes doubled up to escape, `\` as the escapechar and all fields enclosed in double quotes and imports it into platform with the same parameters.

There may be a broader goal of preparing any file (including already existing CSVs) in this fashion to ensure successful import to Platform but for now, this only addresses the dataframe to Civis situation.

I've encountered this bug while writing the Classy & MobileCommons integrations and wrote a modified version of dataframe_to_civis to handle the `/"` problems. This PR brings over the changes that have been succesfully importing data from those system.

I've included tests for the clean_text function.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -612,13 +612,37 @@ def civis_to_multifile_csv(sql, database, job_name=None, api_key=None,
     return unload_manifest
 
 
+def clean_text(string: str):
+    """
+    Clean up a string for upload into Redshift
+    Platform import fails if there are \" anywhere.
+    This puts a space in between.
+    If there's a \ at the end, it adds a space
+
+    Parameters
+    ----------
+    string : str
+        The string to clean for import.
+
+     """
+
+    if not string or pd.isnull(string):
+        return ''
+
+    string = string.replace(r'\"', r'\ "')
+
+    if string[-1] == '\\':
+        string += ' '
+
+    return string
+
+
 @deprecate_param('v2.0.0', 'api_key')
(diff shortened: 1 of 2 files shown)


stephen-hoover:
@ffaristocrat , this looks like a bug in Civis Platform import code; it should be addressed there, rather than building a workaround into the API client. Some of the changes here (like removing the `header` parameter -- why was that there?) seem good, but I don't like the idea of inspecting and modifying every string in a `DataFrame` that users are uploading.

ffaristocrat:
Maybe there should be a warning if the data has problems?

The rest of the upload options (|, double quote, quote all) should be standard though as they cause the least issues uploading into platform.

stephen-hoover:
Could you separate the best-practice CSV creation options into a different PR while we discuss the best approach on this one? Don't forget to update the changelog.

keithing:
Closing this per @stephen-hoover's objections.

PR #257, opened 2018-05-29
Best practices for exporting to csv and uploading to platform

ffaristocrat:
As suggested by @stephen-hoover in #256 , I've split out the best practices into its own PR.

Once this [platform issue](https://civisanalytics.atlassian.net/browse/CIVP-12730) is resolved, this will guarantee dataframes are successfully imported into platform.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -617,7 +617,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        max_errors=None, existing_table_rows="fail",
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
-                       credential_id=None, polling_interval=None,
+                       credential_id=None, headers=None, polling_interval=None,
                        archive=False, hidden=True, **kwargs):
     """Upload a `pandas` `DataFrame` into a Civis table.
 
@@ -660,6 +660,10 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
     credential_id : str or int, optional
         The ID of the database credential.  If ``None``, the default
         credential will be used.
+    headers : bool, optional
+        Whether or not the first row of the file should be treated as
+        headers. The default, ``None``, attempts to autodetect whether
+        or not the first row contains headers.
     polling_interval : int or float, optional
         Number of seconds to wait between checks for job completion.
     archive : bool, optional (deprecated)
@@ -688,7 +692,11 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
     if archive:
         warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
                       "Use `hidden` instead.", FutureWarning)
-
+    if headers is not None:
+        warnings.warn(
+            'The "headers" parameter has been deprecated and will be removed '
+            'in a future version of the API client. Setting it has no effect.',
+            FutureWarning)
     delimiter = '|'
 
     with TemporaryDirectory() as tmp_dir:
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -617,8 +617,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        max_errors=None, existing_table_rows="fail",
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
-                       headers=None, credential_id=None,
-                       polling_interval=None,
+                       credential_id=None, polling_interval=None,
                        archive=False, hidden=True, **kwargs):
     """Upload a `pandas` `DataFrame` into a Civis table.
 
@@ -658,10 +657,6 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
         The column to use as the sortkey for the table.
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
-    headers : bool, optional
-        Whether or not the first row of the file should be treated as
-        headers. The default, ``None``, attempts to autodetect whether
-        or not the first row contains headers.
     credential_id : str or int, optional
         The ID of the database credential.  If ``None``, the default
         credential will be used.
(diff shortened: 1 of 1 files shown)


civis/io/_tables.py
```diff
@@ -617,8 +617,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        max_errors=None, existing_table_rows="fail",
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
-                       headers=None, credential_id=None,
```
stephen-hoover: Rather than removing the `headers` parameter completely, you should leave it as an option in the function signature, and add a call to `warnings.warn`, something like 
```python
if headers is not None:
    warnings.warn('The "headers" parameter has been deprecated and will be removed in a future version of the API client. Setting it has no effect.', FutureWarning)
```
That way users won't have their code break if they'd been setting `headers` in this function. (Hopefully they weren't setting it to anything other than `True`, or their imports would have broken.) We can remove the parameter when we get to a v2 release.

civis/io/_tables.py
```diff
@@ -694,21 +689,26 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
...
 
+    delimiter = '|'
+
     with TemporaryDirectory() as tmp_dir:
         tmp_path = os.path.join(tmp_dir, 'dataframe_to_civis.csv')
-        to_csv_kwargs = {'encoding': 'utf-8', 'index': False}
+        to_csv_kwargs = {
+            'encoding': 'utf-8', 'index': False, 'header': True,
+            'sep': delimiter, 'quoting': csv.QUOTE_ALL, 'escapechar': '\\',
+            'doublequote': True
```
jkulzick: doublequote defaults to True so we shouldn't need this

civis/io/_tables.py
```diff
@@ -694,21 +689,26 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
         warnings.warn("`archive` is deprecated and will be removed in v2.0.0. "
                       "Use `hidden` instead.", FutureWarning)
 
+    delimiter = '|'
```
jkulzick: Pipe is a better delimiter than a comma, but since we're writing out the csv and pandas is doing the quoting of special characters I'm not sure it matters. I might just use the default value, but no harm in using a pipe.

civis/io/_tables.py
```diff
@@ -694,21 +689,26 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
...
+
     with TemporaryDirectory() as tmp_dir:
         tmp_path = os.path.join(tmp_dir, 'dataframe_to_civis.csv')
-        to_csv_kwargs = {'encoding': 'utf-8', 'index': False}
+        to_csv_kwargs = {
+            'encoding': 'utf-8', 'index': False, 'header': True,
+            'sep': delimiter, 'quoting': csv.QUOTE_ALL, 'escapechar': '\\',
```
jkulzick: Is there a reason we need `csv.QUOTE_ALL` instead of the default `csv.QUOTE_MINIMAL`? This adds more bytes to files and increases there size. I would opt for the default unless there is some use case we are aware of.

Quick note on `escapechar`, this is not used unless `quoting=csv.QUOTE_NONE` and `doublequote=False`, so someone would have to explicitly change the defaults on both of those for this to take effect. The default value is `None` which seems appropriate here.

civis/io/_tables.py
ffaristocrat: I'd prefer to be explicit about this since it's to demonstrate best practices.

jkulzick:
@stephen-hoover I added some comments. As a rule of thumb, I would rely on the defaults provided by `to_csv` unless there is an explicit reason to set them.

civis/io/_tables.py
ffaristocrat: I'd have to go back and look but anecdotally, I've only had consistent success importing using a pipe. Some of that may be tied in with the backslash quote issue I mentioned above so I'd have to write up some test cases to see if there's a real difference.

In any case, it's the preferred delimiter in our [import documentation](https://civis.zendesk.com/hc/en-us/articles/213756143-File-Import).

civis/io/_tables.py
jkulzick: By that logic should we list all of the defaults here? IMHO we should only list values we are overriding and probably have a comment that explains why we are changing the default. That makes it more clear to contributors why we are not using pandas defaults.

We can also add a `See Also` section that mentions `pandas.to_csv` so that users/contributors reference their documentation.

civis/io/_tables.py
ffaristocrat: Done

civis/io/_tables.py
ffaristocrat: Quote all is needed because of the back slash quote issue breaking platform import. Everything needs to be wrapped in quotes to avoid having as few backslashes in the file as possible.

The escapechar thing I can change.

ffaristocrat: There's an issue with [platform import](https://civisanalytics.atlassian.net/browse/CIVP-12730) where if there's a backslash and a doublequote next to each other anywhere in the file for any reason, the import will fail. The JIRA has my code demonstrating the problem and the fix I've been using for the MobileCommons & Classy intergrations.

I'm doing a PR for these settings because they're the settings that work with importing any kind of extremely messy text data (aside from the two edge cases in the JIRA) into platform. Right now, it's extremely easy to have imports that fail with no easy way to determine the issue.

jkulzick: @ffaristocrat I ran through this example in detail for the last 2 hours and updated linked JIRA. I still do not agree that quote all is needed or helping resolve the issues presented in the example on the JIRA. If anything, for your example you should set `doublequote=False`.

jkulzick:
@stephen-hoover @ffaristocrat 

I won't be able to provide approval from my side on this PR. My outstanding comments are as follows:

1. I don't agree with including default values for `pandas.to_csv` in the code in `to_csv_kwargs`
2. We are providing default settings which are inconsistent with [CSV RFC4180](https://tools.ietf.org/html/rfc4180#page-2), which `pandas.to_csv` appears to follow.
3. It seems like we are trying to address concerns with how an internal process (Civis file transform) is parsing the file. If there are issues with that process, we should address them there. But once again, its not clear to me how these new settings provide a workaround to the reported issues.

keithing:
Closing this per @jkulzick's objections.

PR #258, opened 2018-06-11
MAINT Configure Windows CI using AppVeyor

### LINKED ISSUE #10: Add Tests for Windows Platforms
The issue in #8 might have been caught if we were testing Windows platforms.  We might able to use something like https://www.appveyor.com/ for this.

jacksonllee:
IT support has turned on AppVeyor for CI in Windows. This PR configures that, based on what we already have for Circle CI.

Resolves #10.

### DIFF
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -5,7 +5,6 @@ from six import BytesIO
 import json
 import os
 import pickle
-import tempfile
 import uuid
 
 import joblib
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -6,6 +6,7 @@ import json
 import os
 import pickle
 import tempfile
+import uuid
 
 import joblib
 try:
@@ -143,8 +144,8 @@ def test_block_and_handle_missing(mock_fut):
 
 @mock.patch.object(_model.cio, 'file_to_civis', return_value=-11)
 def test_stash_local_data_from_file(mock_file):
-    with tempfile.NamedTemporaryFile() as tempfname:
-        fname = tempfname.name
+    with tempfile.TemporaryDirectory() as temp_dir:
+        fname = os.path.join(temp_dir, str(uuid.uuid4()))
         with open(fname, 'wt') as _fout:
             _fout.write("a,b,c\n1,2,3\n")
 
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -5,7 +5,6 @@ from six import BytesIO
 import json
 import os
 import pickle
-import uuid
 
 import joblib
 try:
@@ -144,7 +143,7 @@ def test_block_and_handle_missing(mock_fut):
 @mock.patch.object(_model.cio, 'file_to_civis', return_value=-11)
 def test_stash_local_data_from_file(mock_file):
     with TemporaryDirectory() as temp_dir:
-        fname = os.path.join(temp_dir, str(uuid.uuid4()))
+        fname = os.path.join(temp_dir, 'tempfile')
         with open(fname, 'wt') as _fout:
             _fout.write("a,b,c\n1,2,3\n")
 
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -33,7 +33,7 @@ except ImportError:
 from civis import APIClient
 from civis._utils import camel_to_snake
 from civis.base import CivisAPIError, CivisJobFailure
-from civis.compat import mock, FileNotFoundError
+from civis.compat import mock, FileNotFoundError, TemporaryDirectory
 from civis.response import Response
 import pytest
 
@@ -144,7 +144,7 @@ def test_block_and_handle_missing(mock_fut):
 
 @mock.patch.object(_model.cio, 'file_to_civis', return_value=-11)
 def test_stash_local_data_from_file(mock_file):
-    with tempfile.TemporaryDirectory() as temp_dir:
+    with TemporaryDirectory() as temp_dir:
         fname = os.path.join(temp_dir, str(uuid.uuid4()))
         with open(fname, 'wt') as _fout:
             _fout.write("a,b,c\n1,2,3\n")
--- a/civis/tests/test_io.py
+++ b/civis/tests/test_io.py
@@ -98,8 +98,6 @@ class ImportTests(CivisVCRTestCase):
 
             cls.export_job_id = result.sql_id
 
-    @pytest.mark.skipif(sys.platform.startswith('win'),
-                        reason='does not work yet for Windows')  # TODO
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_zip_member_to_civis(self, *mocks):
         with TemporaryDirectory() as temp_dir:
--- a/civis/tests/test_io.py
+++ b/civis/tests/test_io.py
@@ -99,7 +99,7 @@ class ImportTests(CivisVCRTestCase):
 
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_zip_member_to_civis(self, *mocks):
-        with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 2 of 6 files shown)


civis/tests/test_io.py
```diff
@@ -85,22 +86,25 @@ def setup_class(cls, *mocks):
...
-                result = civis.io.civis_to_csv(tmp.name, sql, database,
+                result = civis.io.civis_to_csv(fname, sql, database,
                                                polling_interval=POLL_INTERVAL)
                 result = result.result()
                 cls.export_url = result['output'][0]['path']
                 assert result.state == 'succeeded'
 
             cls.export_job_id = result.sql_id
 
+    @pytest.mark.skipif(sys.platform.startswith('win'))  # TODO
```
keithing: Did you fix this test? Does it still need to be skipped in Windows?

jacksonllee: The test isn't fixed yet, and I plan to. I threw in this skipif thing to suppress for something else. We're _very_ close to getting the whole thing working. I'll tag you when the PR is ready fo review.

keithing: Great work on this so far!

civis/tests/test_io.py
```diff
@@ -98,9 +99,11 @@ def setup_class(cls, *mocks):
...
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_zip_member_to_civis(self, *mocks):
-        with tempfile.NamedTemporaryFile() as tmp:
-            with zipfile.ZipFile(tmp, 'w', zipfile.ZIP_DEFLATED) as zip_file:
-                zip_file.writestr(tmp.name, 'a,b,c\n1,2,3')
+        with TemporaryDirectory() as temp_dir:
+            fname = os.path.join(temp_dir, str(uuid.uuid4()))
+            with zipfile.ZipFile(fname, 'w', zipfile.ZIP_DEFLATED) as zip_file:
+                archive_name = str(uuid.uuid4())
+                zip_file.writestr(archive_name, 'a,b,c\n1,2,3')
```
jacksonllee: `archive_name` should be its own distinct filename rather than being the same as the zipf file's filename. Also, using the zip file's filename failed this test because apparently the `ZipFile` object normalizes backslashes into forward slashes in filenames internally, which caused test failures like [this](https://ci.appveyor.com/project/eabraun/civis-python/build/1.0.8/job/198d2m1dbp38fpy5).

civis/tests/test_polling.py
```diff
@@ -71,8 +71,8 @@ def test_poll_on_creation(self):
                                   poll_on_creation=False)
         pollable.done()  # Check status once to start the polling thread
         assert poller.call_count == 0
-        time.sleep(0.015)
-        assert poller.call_count == 1
+        time.sleep(0.02)
+        assert poller.call_count > 0
```
jacksonllee: Because now we have different execution environments for builds (Circle CI versus AppVeyor), this test has become a little unstable (see this [test failure](https://ci.appveyor.com/project/eabraun/civis-python/build/1.0.9/job/9nn5y16e1ctvjqrk) due to such indeterminancy). Given `polling_internal=0.01` (set just a few lines above, via `PollableResult`),  sleeping 0.02s gives a little more time for at least one poller call count (if not two).

dev-requirements.txt
```diff
@@ -7,7 +7,7 @@ vcrpy>=1.11.0,<=1.11.99
 vcrpy-unittest==0.1.6
 sphinx_rtd_theme>=0.2.4,<1.0.0
 numpydoc>=0.7.0,<1.0.0
-feather-format
+feather-format; sys_platform != 'win32' or python_version >= '3.5'
```
jacksonllee: `feather-format` depends on `pyarrow`, which doesn't have a Windows installer for Python < 3.5; see [here](https://arrow.apache.org/docs/python/install.html). Not having `feather-format` installed only [skips one test](https://github.com/civisanalytics/civis-python/blob/403f318294ab4e204fd1f0c4fb0678105fdbe1d7/civis/ml/tests/test_model.py#L180), which I think is fine (and we _are_ running this test for Python >= 3.5 on Windows anyway).

jacksonllee:
@keithing I've defeated AppVeyor! Good news is that only changes in tests are needed for Windows compatibility. It's mostly about not using `tempfile.NamedTemporaryFile`. I've left comments for other changes. This PR is ready for review.

civis/tests/test_io.py
keithing: So this means that you can't use `archive_name = os.path.join(str(temp_dir, uuid.uiid4()))`?  If you run the tests locally, won't this litter your folder?

jacksonllee: I've just run the test suite locally to verify that no random temp files are created. `zip_file.writestr(archive_name, ...)` writes a new member file named `archive_name` to `zip_file` which is the already instantiated temp zip file; it doesn't mean creating a file called `archive_name` at the current directory.

civis/ml/tests/test_model.py
```diff
@@ -143,8 +143,8 @@ def foo(arg):
 
 @mock.patch.object(_model.cio, 'file_to_civis', return_value=-11)
 def test_stash_local_data_from_file(mock_file):
-    with tempfile.NamedTemporaryFile() as tempfname:
-        fname = tempfname.name
+    with TemporaryDirectory() as temp_dir:
+        fname = os.path.join(temp_dir, str(uuid.uuid4()))
```
stephen-hoover: You don't need a uuid for anything created in a temporary directory -- the file will vanish as soon as you exit the context, and I don't see a danger of a name collision inside this test.

civis/tests/test_io.py
```diff
@@ -129,11 +132,13 @@ def test_bytes_file_to_civis(self, *mocks):
...
-            tmp.write(b'a,b,c\n1,2,3')
-            tmp.flush()
-            tmp.seek(0)
-            result = civis.io.file_to_civis(tmp, tmp.name)
+        with TemporaryDirectory() as temp_dir:
+            fname = os.path.join(temp_dir, str(uuid.uuid4()))
+            with open(fname, 'w+b') as tmp:
+                tmp.write(b'a,b,c\n1,2,3')
+                tmp.flush()
+                tmp.seek(0)
```
stephen-hoover: I would exit the write context and re-open the file in a new `with open(...)` context here. The `seek(0)` was so that we could use a temporary file which disappeared when closed.

civis/tests/test_io.py
```diff
@@ -148,13 +153,15 @@ def test_civis_to_file(self, *mocks):
...
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_csv_to_civis(self, *mocks):
-        with tempfile.NamedTemporaryFile() as tmp:
-            tmp.write(b'a,b,c\n1,2,3')
-            tmp.flush()
+        with TemporaryDirectory() as temp_dir:
+            fname = os.path.join(temp_dir, str(uuid.uuid4()))
+            with open(fname, 'w+b') as tmp:
+                tmp.write(b'a,b,c\n1,2,3')
+                tmp.flush()
```
stephen-hoover: This should be unnecessary now that we're using a write context manager, but it also won't hurt anything.

civis/tests/test_polling.py
stephen-hoover: It was a little unstable in the Linux tests, to be honest. Hopefully this change makes life better for everyone. Thank you!

stephen-hoover:
This test failure was caused by https://github.com/pandas-dev/pandas/pull/21300 . I'm not sure if it represents a bug on our end or just a change in the parameter options available for `to_csv`. I think we should fix `_stash_dataframe_as_csv` in a different PR so that it works with the latest version of `pandas`.

jacksonllee:
I'm sorting out the repo / AppVeyor settings with IT support. Sit tight, everyone.

jacksonllee:
We're all set. All builds for both Linux and Windows have passed. This PR is ready for review again.

PR #259, opened 2018-06-13
BUG A file-like object as arg to df.to_csv must have the method getvalue

jacksonllee:
Since `pandas` v0.23.1 (specifically from https://github.com/pandas-dev/pandas/pull/21300), if a file-like object is passed into `df.to_csv`, the object must have the method `getvalue`, or else an `AttributeError` is raised (which is what we have observed from #258). An `io.TextIOWrapper` object doesn't have the method `getvalue`, which caused the latest build failures in #258. This PR switches to `io.StringIO` instead.

### DIFF
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -1,12 +1,12 @@
 from builtins import super
 from collections import namedtuple
 from concurrent.futures import CancelledError
+from six import StringIO
 import json
 import os
 import pickle
 import tempfile
 
-import six
 import joblib
 try:
     import pandas as pd
@@ -173,7 +173,7 @@ def test_stash_local_data_from_dataframe_csv(mock_file):
     assert _model._stash_dataframe_as_csv(df, mock.Mock()) == -11
     mock_file.assert_called_once_with(mock.ANY, name='modelpipeline_data.csv',
                                       client=mock.ANY)
-    assert isinstance(mock_file.call_args[0][0], six.StringIO)
+    assert isinstance(mock_file.call_args[0][0], StringIO)
 
 
 @pytest.mark.skipif(not HAS_PANDAS, reason="pandas not installed")
--- a/civis/ml/tests/test_model.py
+++ b/civis/ml/tests/test_model.py
@@ -5,8 +5,8 @@ import json
 import os
 import pickle
 import tempfile
-import io
 
+import six
 import joblib
 try:
     import pandas as pd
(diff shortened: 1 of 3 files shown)


jacksonllee:
AppVeyor has been turned on, but obviously there's no AppVeyor configuration yet (upcoming from #258) and hence the relevant build "failure", which should be ignored for this PR.

civis/ml/tests/test_model.py
```diff
@@ -173,7 +173,7 @@ def test_stash_local_data_from_dataframe_csv(mock_file):
     assert _model._stash_dataframe_as_csv(df, mock.Mock()) == -11
     mock_file.assert_called_once_with(mock.ANY, name='modelpipeline_data.csv',
                                       client=mock.ANY)
-    assert isinstance(mock_file.call_args[0][0], BytesIO)
+    assert isinstance(mock_file.call_args[0][0], StringIO)
```
keithing: Shouldn't this fail in the Python27 tests?

jacksonllee: Good question! Behold --

Python 2.7, six 1.11.0:

```python
In [2]: from six import StringIO, BytesIO

In [3]: s = StringIO()

In [4]: b = BytesIO()

In [5]: isinstance(s, StringIO)
Out[5]: True

In [6]: isinstance(s, BytesIO)  # This is why we didn't get the expected failure?
Out[6]: True

In [7]: isinstance(b, StringIO)
Out[7]: True

In [8]: isinstance(b, BytesIO)
Out[8]: True
```

Python 3.6, six 1.11.0 (where everything makes sense to me):

```python
In [1]: from six import StringIO, BytesIO

In [2]: s = StringIO()

In [3]: b = BytesIO()

In [4]: isinstance(s, StringIO)
Out[4]: True

In [5]: isinstance(s, BytesIO)
Out[5]: False

In [6]: isinstance(b, StringIO)
Out[6]: False

In [7]: isinstance(b, BytesIO)
Out[7]: True
```

keithing: Huh.  Thanks for looking into that.

stephen-hoover: I think the test passes because of https://github.com/benjaminp/six/blob/1.11.0/six.py#L664 . Given that equality, do we need the `if six.PY3` conditional?

keithing: Good point! Although I don't have a problem merging as is, either.

PR #260, opened 2018-06-13
MAINT Configure Windows CI using AppVeyor

### LINKED ISSUE #10: Add Tests for Windows Platforms

jacksonllee:
IT support has turned on AppVeyor for CI in Windows. This PR configures that, based on what we already have for Circle CI.

Resolves #10.


jacksonllee:
This is the same as #258. Creating this PR just to see if AppVeyor builds are triggered (which apparently are not...)

PR #264, opened 2018-09-24
make smoke tests not fail if row order varies

mheilman:
The smoke tests were failing because rows would come back in a different order.  This change makes it so that the order shouldn't matter.


mheilman:
The test failures (only in Windows) appear unrelated to this PR.  It seems like we ought to file an issue that links to the CI logs and merge this.  @keithing, does that sound OK?

PR #273, opened 2018-11-30
Add a retry for downloading files via civis.io.file_to_civis

### LINKED ISSUE #272: Retry on Intermittent S3 Errors
Recently hit an error here:
https://github.com/civisanalytics/civis-python/blob/7a08b8c238ba1a6f447ddee68e4ddfc78a4ea560/civis/io/_tables.py#L346
where S3 returned an intermittent 500 on our GET to retrieve an object, causing the read_civis_sql call to raise an exception.
It'd be nice if we had a retry loop for S3 interactions to avoid these intermittent failures.

mheilman:
It also fixes a related bug in file_to_civis where the buffer was reset to 0 rather than the original position when retrying.

Fixes #272.

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -97,9 +97,14 @@ def _single_upload(buf, name, client, **kwargs):
     form_key = OrderedDict(key=form.pop('key'))
     form_key.update(form)
 
+    # Store the current buffer position in case we need to retry below.
+    buf_orig_position = buf.tell()
+
     @retry(RETRY_EXCEPTIONS)
     def _post():
-        buf.seek(0)
+        # Reset the buffer in case we had to retry.
+        buf.seek(buf_orig_position)
+
         form_key['file'] = buf
         # requests will not stream multipart/form-data, but _single_upload
         # is only used for small file objects or non-seekable file objects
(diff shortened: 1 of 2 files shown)


mheilman:
@waltaskew? @jkulzick?

civis/io/_files.py
```diff
@@ -315,11 +320,22 @@ def _civis_to_file(file_id, buf, api_key=None, client=None):
...
-    response.raise_for_status()
-    chunked = response.iter_content(CHUNK_SIZE)
-    for lines in chunked:
-        buf.write(lines)
+
+    # Store the current buffer position in case we need to retry below.
+    buf_orig_position = buf.tell()
+
+    @retry(RETRY_EXCEPTIONS)
+    def _download_url_to_buf():
```
jkulzick: Another way to do this without re-downloading the bytes you already retrieved is to use range headers in the HTTP Get request like [so](https://github.com/civisanalytics/bowerbird/blob/master/bowerbird/filesystem/_http.py#L88). You can probably do something similar on Put but I haven't done that before.

Up to you...

mheilman: That's a good idea, but I'm going to punt on it for now.  I filed #274.

PR #277, opened 2019-01-29
BUG check for primary key to determine how to read index

elsander:
This PR fixes a bug to ensure compatibility with CivisML v2.2.2. In 2.2.2, CivisML does not create an index column when the user does not specify a primary key. This fix forces `civis.ml.ModelFuture.table` to check for a primary key before determining if the first column is an index or not. Unfortunately, the metadata structure is different for training and prediction, so this fix includes two complicated hardcoded paths through the metadata, but I don't see any way around this. I tested the fix on three actual model jobs (training with pkey, predict with pkey, predict without pkey) and it behaved as expected for all three.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -489,20 +489,23 @@ class ModelFuture(ContainerFuture):
 
         return self._train_data
 
+    def _table_primary_key(self):
+        # metadata path to input parameters is different
+        # for training and prediction
+        if self.is_training:
+            pkey = self.metadata[
+                'run']['configuration']['data']['primary_key']
+        else:
+            pkey = self.metadata[
+                'jobs'][0]['run']['configuration']['data']['primary_key']
+        return pkey
+
     @property
     def table(self):
         self.result()  # Block and raise errors if any
         if self._table is None:
             # An index column will only be present if primary key is
-            # metadata path to input parameters is different
-            # for training and prediction
-            if self.is_training:
-                pkey = self.metadata[
-                    'run']['configuration']['data']['primary_key']
-            else:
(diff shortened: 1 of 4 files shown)


elsander:
Looks like there's a Travis issue with installing pytest; it's building on 3.5 but nothing else. Have you run into this before, @stephen-hoover ?

stephen-hoover:
I have not. Looks like there's version incompatibilities. The 3.6 build has the error

> pytest-cov 2.6.1 has requirement pytest>=3.6, but you'll have pytest 3.3.0 which is incompatible.
> pyarrow 0.12.0 has requirement numpy>=1.14, but you'll have numpy 1.13.3 which is incompatible.

My guess is that you could solve this by adjusting some of the version requirements in `dev-requirements.txt`.

civis/ml/_model.py
```diff
@@ -493,13 +493,27 @@ def train_data(self):
...
         if self._table is None:
+            # An index column will only be present if primary key is
+            # metadata path to input parameters is different
+            # for training and prediction
+            if self.is_training:
+                pkey = self.metadata[
+                    'run']['configuration']['data']['primary_key']
+            else:
+                pkey = self.metadata[
+                    'jobs'][0]['run']['configuration']['data']['primary_key']
```
stephen-hoover: Do you think that it's worth pulling this logic block out into a private `_table_primary_key` (or something like that) method? Parsing out the primary key's column name feels logically distinct from this method, which is downloading the table produced by the CivisML job.

elsander: Yeah, I was on the fence about that myself. I'll pull it into a helper method.

PR #283, opened 2019-02-15
FIX get_table_id should correctly handle quoted schema.tablename

### LINKED ISSUE #276: get_table_id does not correctly handle quoted schemas and tablenames
The `get_table_id` function throws a `ValueError` if the schema and tablename are quoted, which doesn't seem to be the correct behavior.

## Civis library version

1.9.2

## How to reproduce the bug

```python
In [1]: import civis

In [2]: client = civis.APIClient()

In [3]: client.get_table_id(table='foo.bar', database='baz')
Out[3]: 123

In [4]: client.get_table_id(table='"foo"."bar"', database='baz')
ValueError: Given table "foo"."bar" is not an exact match for returned table foo.bar.
```

## Desired behavior

```python
In [1]: import civis

In [2]: client = civis.APIClient()

In [3]: client.get_table_id(table='foo.bar', database='baz')
Out[3]: 123

In [4]: client.get_table_id(table='"foo"."bar"', database='baz')
Out[4]: 123
```

## Likely cause/fix

While `get_table_id` derives the schema and tablename via `civis.io.split_schema_tablename`:

https://github.com/civisanalytics/civis-python/blob/eb5d46c8ce14f1bb87b889b69c7fd9e1c751d777/civis/civis.py#L271

which (correctly) gives us `'foo'` as the schema and `'bar'` as the tablename (both without quotes) with `'"foo"."bar"'` (with quotes) as the input, it looks like the following bit (whose purpose is unclear to me) can be removed for the desired behavior:

https://github.com/civisanalytics/civis-python/blob/eb5d46c8ce14f1bb87b889b69c7fd9e1c751d777/civis/civis.py#L277-L280

That being said, perhaps there's more to it, and the current maintainers would probably know more.

jacksonllee:
Resolves #276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-
+### Fixed
+- `get_table_id` will correctly handle quoted schema.tablename. (#283)
 
 ## 1.9.3 - 2019-02-05
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 ### Fixed
-- `get_table_id` will correctly handle quoted schema.tablename. (#283)
+- `get_table_id` will correctly handle a double-quoted schema.tablename. (#283)
 
 ## 1.9.3 - 2019-02-05
 ### Fixed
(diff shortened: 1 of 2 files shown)


jacksonllee:
1. It looks like none of the methods under the `MetaMixin` class have tests. To document the desired behavior fixed by this current PR, I've put down examples in the docstring of `get_table_id` (following a similar pattern from other methods in the class, such as [`get_database_credential_id`](https://github.com/civisanalytics/civis-python/blob/v1.9.3/civis/civis.py#L156-L164)).
2. The Travis CI builds didn't kick off -- I'm emailing IT-Support. Marking this PR as work in progress for now.

jacksonllee:
According to IT-support (who emailed Travis CI's support), Travis CI doesn't yet support builds for PRs marked as "draft" (a new feature on GitHub). Closing this PR in favor of #285.

PR #285, opened 2019-02-18
FIX get_table_id should correctly handle quoted schema.tablename

### LINKED ISSUE #276: get_table_id does not correctly handle quoted schemas and tablenames

jacksonllee:
Resolves #276

### DIFF
--- a/civis/tests/test_civis.py
+++ b/civis/tests/test_civis.py
@@ -1,6 +1,5 @@
-from unittest import mock
-
 import civis
+from civis.compat import mock
 
 import pytest
 
--- a/civis/tests/test_civis.py
+++ b/civis/tests/test_civis.py
@@ -1,0 +1,28 @@
+from unittest import mock
+
+import civis
+
+import pytest
+
+
+@pytest.mark.parametrize('schema_tablename', [
+    'foo.bar', '"foo".bar', 'foo."bar"', '"foo"."bar"'
+])
+def test_get_table_id(schema_tablename):
+    """Check that get_table_id handles quoted schema.tablename correctly."""
+    mock_client = mock.Mock()
+    mock_client.get_database_id.return_value = 123
+
+    mock_tables = mock.MagicMock()
+    mock_tables.__getitem__.side_effect = {0: mock.Mock()}.__getitem__
+
+    mock_client.tables.list.return_value = [mock_tables]
+
+    civis.APIClient.get_table_id(mock_client,
+                                 table=schema_tablename, database=123)
+
+    mock_client.tables.list.assert_called_once_with(
+        database_id=123,
+        schema='foo',
+        name='bar'
+    )
--- a/civis/tests/test_civis.py
+++ b/civis/tests/test_civis.py
@@ -1,5 +1,6 @@
 import civis
 from civis.compat import mock
+from civis.tests.mocks import TEST_SPEC
 
 import pytest
 
@@ -9,18 +10,17 @@ import pytest
 ])
 def test_get_table_id(schema_tablename):
     """Check that get_table_id handles quoted schema.tablename correctly."""
-    mock_client = mock.Mock()
-    mock_client.get_database_id.return_value = 123
+    client = civis.APIClient(local_api_spec=TEST_SPEC, api_key='none')
+    client.get_database_id = mock.Mock(return_value=123)
 
     mock_tables = mock.MagicMock()
     mock_tables.__getitem__.side_effect = {0: mock.Mock()}.__getitem__
 
-    mock_client.tables.list.return_value = mock_tables
+    client.tables.list = mock.Mock(return_value=mock_tables)
 
-    civis.APIClient.get_table_id(mock_client,
-                                 table=schema_tablename, database=123)
+    client.get_table_id(table=schema_tablename, database=123)
 
-    mock_client.tables.list.assert_called_once_with(
+    client.tables.list.assert_called_once_with(
         database_id=123,
         schema='foo',
         name='bar'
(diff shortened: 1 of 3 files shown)


jacksonllee:
All Travis builds have passed. Waiting for AppVeyor builds to finish: https://ci.appveyor.com/project/civisanalytics/civis-python/builds/22468450

jacksonllee:
The AppVeyor builds have finished:

screen shot 2019-02-19 at 11 18 34 am

The Python 3.6 build failed, but the issue was the sort of non-deterministic test behavior we've seen from time to time that I think we can safely ignore for this PR:

```
================================== FAILURES ===================================
_____________________________ test_custom_scripts _____________________________
    def test_custom_scripts():
        with mock.patch.dict('os.environ', {'CIVIS_JOB_ID': '12',
                                            'CIVIS_RUN_ID': '40'}):
>           c = _check_executor(133)
civis\tests\test_futures.py:292: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
from_template_id = 133
    def _check_executor(from_template_id=None):
        job_id, run_id = 42, 43
        c = _setup_client_mock(job_id, run_id, n_failures=0)
        mock_run = c.scripts.post_containers_runs()
        if from_template_id:
            bpe = CustomScriptExecutor(from_template_id=from_template_id,
                                       client=c, polling_interval=0.01)
            future = bpe.submit(my_param='spam')
        else:
            bpe = _ContainerShellExecutor(client=c, polling_interval=0.01)
            future = bpe.submit("foo")
    
        # Mock and test running, future.job_id, and done()
        mock_run.state = "running"
>       assert future.running(), "future is incorrectly marked as not running"
E       AssertionError: future is incorrectly marked as not running
E       assert False
E        +  where False = <bound method Future.running of <ContainerFuture at 0x23e50ed518 state=succeeded returned Response>>()
E        +    where <bound method Future.running of <ContainerFuture at 0x23e50ed518 state=succeeded returned Response>> = <ContainerFuture at 0x23e50ed518 state=succeeded returned Response>.running
civis\tests\test_futures.py:257: AssertionError
```

This PR is ready for review.

civis/tests/test_civis.py
```diff
@@ -0,0 +1,27 @@
...
+    mock_client = mock.Mock()
+    mock_client.get_database_id.return_value = 123
+
+    mock_tables = mock.MagicMock()
+    mock_tables.__getitem__.side_effect = {0: mock.Mock()}.__getitem__
+
+    mock_client.tables.list.return_value = mock_tables
+
+    civis.APIClient.get_table_id(mock_client,
+                                 table=schema_tablename, database=123)
```
jacksonllee: @stephen-hoover The Python 3 builds have passed, but the Python 2.7 one failed on this line (with `TypeError: unbound method get_table_id() must be called with APIClient instance as first argument (got Mock instance instead)`). Would it fine to just `@pytest.mark.skip` this test for Python 2.7? If not, I don't see how we can have a mock client object without also mocking out `get_table_id`. Or am I missing something?

stephen-hoover: If you want a mock client object, it's better to use the `civis.tests.mocks.create_client_mock` function to get one (that will handle autospeccing for you). In this case, you might actually want to create a real client object and then replace the functions you want to mock with `mock.Mock` objects. What do you think of trying:
```python
from civis.tests.mocks import TEST_SPEC
client = APIClient(local_api_spec=TEST_SPEC, api_key='none')
client.get_database_id = mock.Mock(return_value=123)
```
and so forth. If you want to argue that there's so much mocking going on that the test isn't useful, I'm open to that. This test seems nice to have, but maybe a separate test on `civis.io.split_schema_tablename` is all that's needed here.

jacksonllee: Ah, `TEST_SPEC` is what I didn't know about! The test has been updated, and the builds have passed.

stephen-hoover:
It looks like there's still something going on with Travis. Why does it need to run tests on both your branch and the PR? Aren't those the same code? Maybe I'm misunderstanding what "Branch" and "Pull Request" mean there.

jacksonllee:
I've contacted IT-Support to help sort out this repo's config (hopefully once and for all) after I'd noticed merging is still blocked. Will report back as soon as I have updates.

jacksonllee:
According to IT-Support, apparently there are indeed both "branch" and "pull request" options for Travis CI checks, and somehow "branch" was checked -- it is not anymore, and hence everything green now.

Issue #149, opened 2017-09-26
Model polling fails intermittently


DerrickHiggins:
For some reason, when one of the models submitted for parallel training using the `civis.futures` module has `cross_validation_parameters` specified, polling with `futures.wait` does not work.

I get the following error message:
```
Exception in thread Thread-2:
Traceback (most recent call last):
  File "/usr/local/Cellar/python3/3.5.2_1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/threading.py", line 914, in _bootstrap_inner
    self.run()
  File "/Users/dch024/virtualenvs/scipy3/lib/python3.5/site-packages/civis/polling.py", line 41, in run
    if self.poller(*self.poller_args).state in DONE:
AttributeError: 'NoneType' object has no attribute 'state'
```

Can be replicated by running the [parallel training example](https://github.com/civisanalytics/civis-python/blob/master/examples/CivisML_parallel_training.ipynb)

stephen-hoover:
Thank you for the bug report, @DerrickHiggins ! I see you're using Python v3.5.2. What version of `civis` are you using?

When you say that you get this failure in the parallel training example, do you mean cell 8,
```python
%%time 
future_jobs = [model.train(df=df) for model in models]
futures.wait(future_jobs)
```
?

Do you reliably get this error with every call? I've been unable to reproduce in either Python v3.5.2 or v3.6.2, using various versions of `civis`.

DerrickHiggins:
Hey, @stephen-hoover -- turns out the issue is intermittent, and not related to the presence of the `cross_validation_parameters` parameter to `ModelPipeline`. (I thought it was, as it only happened with that parameter set in my limited testing.)

[Here](https://gist.github.com/DerrickHiggins/dd666e5c0c79b91415b6696a3e12c774) is a script to replicate the error.  When I run it, I get the following output:

```
3 / 30 jobs threw exceptions
27 / 30 jobs succeeded
```

This is under python 3.5.2, and with `civis-python` @ 09a92f0fea0893dfdc221f3dac2cd022c97484b6

stephen-hoover:
@DerrickHiggins , an update on my debugging (lack of) progress -- I'm still unable to replicate. I ran the script you provided 11 times using a Python 3.5.2 conda environment and civis-python 09a92f0 , with no exceptions (OS X 10.11.6). If I get a chance this weekend, I'll try from my home network, in case a poorer network connection makes any difference.

[...]

PR #289, opened 2019-03-11
[CIVP-17567] Check if polling result is None before checking its state

hinnefe2:
This PR adds a check to `civis.polling._ResultPollingThread.run()` to make sure the result returned by the polling function is not `None` before attempting to access the result's `state` attribute. 

According to the discussion in #149 the `None` return value is the result of intermittent connectivity errors. The suggested workaround was to rerun the script, but that's annoying when the script takes a long time to run. With this change a `None` response gets treated the same as a response which has a non-DONE state.


civis/polling.py
```diff
@@ -38,7 +38,8 @@ def run(self):
         """Poll until done.
         """
         while not self.finished.wait(self.polling_interval):
-            if self.poller(*self.poller_args).state in DONE:
+            poller_result = self.poller(*self.poller_args)
+            if poller_result is not None and poller_result.state in DONE:
```
mheilman: This merits a comment, I think.

stephen-hoover:
Thank you for adding this! Please also add a line to the changelog describing the change. This would go under "Fixed".

hinnefe2:
I looked into what could be returning `None`s a little bit.

The path I followed was `resources._resources.create_method() -> base.Endpoint._call_api -> response.convert_response_data_type` which lands [here](https://github.com/civisanalytics/civis-python/blob/6b9ccf7e55097e22c1bb5aa744270722ce81e464/civis/response.py#L66).

The only I thing I can thing of is maybe the `assert` is getting removed during compilation (which happens when you [optimize the compilation](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement)) and the request is returning a response type not in `snake, raw, pandas` in which case none of the `if` blocks trigger. I'm not at all confident I understand the flow of how the endpoint functions get generated though so this could be a red herring.

Anyway, I'll merge this PR for now.

PR #290, opened 2019-03-15
ENH Add job_id and run_id property attributes to CivisFuture

jacksonllee:
From time to time, it is necessary to pull out the job ID and run ID from a `CivisFuture` object (e.g., for debugging). Based on internal user experience from both co-workers and myself, it wasn't immediately clear how the IDs can be retrieved, until one realizes that (as of civis v1.9.4) `CivisFuture` subclasses `PollableResult`, which in turn subclasses `CivisAsyncResultBase` that has the `poller_args` attribute which is a tuple of (job_id, run_id). It would be nice to expose the IDs at `CivisFuture` and do so without the language of "polling" or similar.

In terms of implementation, it looks like `ContainerFuture`, which subclasses `CivisFuture`, already has the property attributes `job_id` and `run_id`. This PR moves the relevant code from `ContainerFuture` to `CivisFuture`; class inheritance preserves the `job_id` and `run_id` behavior at `ContainerFuture`. (The only other class that inherits from `CivisFuture` is `SubscribableResult`, which is just `CivisFuture` and has been marked to be deprecated.)


jacksonllee:
> Could you mention these in an "Attributes" section of the docstring?

Done!

jacksonllee:
> What do you think of changing the example to get_runs?

Sounds good -- done at bf7be00

civis/futures.py
```diff
@@ -91,29 +91,27 @@ class CivisFuture(PollableResult):
...
         First element of the tuple given to `poller_args`
-    run_id : int
-        Second element of the tuple given to `poller_args`, if any
+    run_id : int or None
+        Second element of the tuple given to `poller_args`
+        (`None` if the poller function does not require a run ID)
 
     Examples
     --------
-    This example is provided as a function at :func:`~civis.io.query_civis`.
```
stephen-hoover: I think this is still true, isn't it? The mechanics of how we're polling for output in that function may be different, but I think an important purpose of this comment is to prevent people from copy-pasting the below code when they want to run a query.

jacksonllee: I took it away as I found it somewhat confusing, though I don't have a strong preference. Comment added back at 1d53193

PR #295, opened 2019-04-30
Upgrade dependency of jsonref

aripollak:
* Fixes DeprecationWarning in python 3.7
* Also fix pubnub ~= version that is no longer recognized in pip 19.0

This happens when jsonref 0.1 is used, but the warning is now fixed in jsonref 0.2:
```python
% python3 -W default
Python 3.7.3 (default, Mar 27 2019, 09:23:15)
[Clang 10.0.1 (clang-1001.0.46.3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import civis
/usr/local/lib/python3.7/site-packages/jsonref.py:8: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,11 +3,11 @@ click>=6.0,<=6.99
 funcsigs==1.0.2 ; python_version == '2.7'
 future>=0.16,<=0.99 ; python_version == '2.7'
 futures==3.1.1 ; python_version == '2.7'
-jsonref>=0.1.0,<=0.1.99
+jsonref>=0.1,<=0.2
 requests>=2.12.0,==2.*
 jsonschema>=2.5.1,==2.*
 functools32>=3.2,<=3.99 ; python_version == '2.7'
 six>=1.10,<=1.99
 joblib~=0.11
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -3,11 +3,11 @@ click>=6.0,<=6.99
...
 futures==3.1.1 ; python_version == '2.7'
-jsonref>=0.1.0,<=0.1.99
+jsonref>=0.1,<=0.2
 requests>=2.12.0,==2.*
 jsonschema>=2.5.1,==2.*
 functools32>=3.2,<=3.99 ; python_version == '2.7'
 six>=1.10,<=1.99
 joblib~=0.11
-pubnub~=4
+pubnub~=4.0
```
stephen-hoover: While you're fixing this, I think you should change it to match the requirements in the `setup.py` file. Could you include a mention of this in the changelog?
```suggestion
pubnub>=4.0,<=4.99
```

PR #297, opened 2019-05-07
Clarified exception behavior of futures

aripollak:
It was a bit surprising to me that the documentation treated `concurrent.futures.wait` the same as `futures.result()`, but the former doesn't raise an exception and the latter does. Maybe there's a better solution than this, but I thought I'd at least add some clarification.


aripollak:
Thanks, updated with your suggestions. Would it also be useful to show a more concrete example of why someone would want to use `concurrent.futures.wait`, like using it with a `timeout` or custom `return_when` with multiple futures? Or would that be more confusing?

aripollak:
> There's already documentation about it in the Python standard library, so it might just help to be more explicit that our `CivisFuture` is not an idiosyncratic Civis API client thing.

Okay, I think that's already pretty explicit in this file, so I won't add anything else.

PR #299, opened 2019-06-03
Compatibility with joblib > 0.11

### LINKED ISSUE #296: Compatibility with joblib >0.11
Our custom joblib backend relies on some internal joblib features which changed in the v0.12 release. The most recent joblib release is now 0.13. We should update our code so that we're compatible with v0.11 through v0.13.

stephen-hoover:
It turns out our incompatibility with newer versions of `joblib` were due to mocking we were doing in the tests.

In v0.12, a `joblib` change meant that it relies on callback from the result object to decide when to terminate batched consumpution of inputs. Two of our tests mock the result object, which means that joblib's supplied callback doesn't get called. Set `n_jobs=1` so that `joblib` consumes the inputs all at once and doesn't need the callbacks.

This PR also relaxs requirement on the `joblib` and `cloudpickle` version numbers, and modifies our custom backend to include some new backend features.

Closes #296

### DIFF
--- a/civis/parallel.py
+++ b/civis/parallel.py
@@ -9,6 +9,7 @@ import logging
 import os
 import pickle
 import time
+import warnings
 
 import cloudpickle
 from joblib._parallel_backends import ParallelBackendBase
@@ -23,9 +24,17 @@ from civis.compat import TemporaryDirectory
 from civis.futures import _ContainerShellExecutor, CustomScriptExecutor
 
 try:
-    from sklearn.externals.joblib import (
-        register_parallel_backend as _sklearn_reg_para_backend)
-    NO_SKLEARN = False
+    with warnings.catch_warnings():
+        # Ignore the warning: "DeprecationWarning: sklearn.externals.joblib is
+        # deprecated in 0.21 and will be removed in 0.23. Please import this
+        # functionality directly from joblib, which can be installed with:
+        # pip install joblib. If this warning is raised when loading pickled
+        # models, you may need to re-serialize those models with
+        # scikit-learn 0.21+."
+        warnings.simplefilter('ignore', DeprecationWarning)
+        from sklearn.externals.joblib import (
+            register_parallel_backend as _sklearn_reg_para_backend)
+        NO_SKLEARN = False
 except ImportError:
     NO_SKLEARN = True
 
--- a/civis/parallel.py
+++ b/civis/parallel.py
@@ -716,8 +716,13 @@ class _CivisBackendResult:
 
         return _fetch_result
 
-    def get(self):
+    def get(self, timeout=None):
         """Block and return the result of the job
+
+        Parameters
+        ----------
+        timeout: float, optional
+            If provided, wait this many seconds before issuing a TimeoutError
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -57,9 +57,9 @@ def main():
             'requests>=2.12.0,==2.*',
             'jsonschema>=2.5.1,==2.*',
             'six>=1.10,<=1.99',
-            'joblib>=0.11,<=0.11.99',
+            'joblib>=0.11,<=0.13.99',
```
mheilman: why not 0.14, out of curiosity?

civis/parallel.py
```diff
@@ -761,18 +775,24 @@ class _CivisBackend(ParallelBackendBase):
 
     Users should interact with this through ``make_backend_factory``.
     """
+    uses_threads = False
```
mheilman: Is setting this to `False` unnecessary?  I'm fine with keeping it, to be explicit.

(It looks like the other backends [don't always set all three of these](https://github.com/joblib/joblib/blob/acca6b95ed9e28498441d630ed2ddf674b66c8b8/joblib/_parallel_backends.py#L171-L172))

civis/parallel.py
```diff
@@ -761,18 +775,24 @@ class _CivisBackend(ParallelBackendBase):
...
                  remote_backend='sequential',
+                 nesting_level=0,
                  **executor_kwargs):
         self.setup_cmd = setup_cmd
         self.from_template_id = from_template_id
         self.max_submit_retries = max_submit_retries
         self.client = client
         self.remote_backend = remote_backend
         self.executor_kwargs = executor_kwargs
+        self.nesting_level = nesting_level
```
mheilman: What does it mean for this to be `>0`?

setup.py
stephen-hoover: It's not released yet, so I can't test on it.

mheilman: oh, sorry, i'd only looked at https://joblib.readthedocs.io/en/latest/developing.html#latest-changes, which makes it look like it's released.

civis/parallel.py
stephen-hoover: I think it's unnecessary, yes. I added it to be explicit. The places in the joblib code which look for it access it via a `getattr` that defaults to `False`.

civis/parallel.py
stephen-hoover: The `nesting_level` attribute is to handle the case where you get a `Parallel` call inside a `Parallel` call, e.g. a multiprocessing Estimator called inside a cross-validation loop. I added it because it's now part of the `ParallelBackendBase` signature (https://github.com/joblib/joblib/blob/acca6b95ed9e28498441d630ed2ddf674b66c8b8/joblib/_parallel_backends.py#L36). This function looks at it: https://github.com/joblib/joblib/blob/acca6b95ed9e28498441d630ed2ddf674b66c8b8/joblib/_parallel_backends.py#L120 . We're not using it, so for now the purpose is just making sure we accept all of the parameters that the parent class does.

setup.py
stephen-hoover: That confused me too, but there's no tag for it yet.

PR #301, opened 2019-06-03
ENH Include Python version in User-Agent string

### LINKED ISSUE #255: Add Python version to User-Agent header
The API client is currently recording its name ("civis-python"), its version number, and the version of `requests` which it's using. We should add the Python version to this. It will be useful to know how many people are using different language versions, so we can get an idea of which versions are most important to support.

stephen-hoover:
Collect the user's Python version along with the API client version. This can be useful in evaluating maintenance requirements going forward.

Closes #255 .


stephen-hoover:
@keithing , I had a merge conflict with the changelog. Could you check once more, please?

PR #302, opened 2019-06-04
ENH last_response attribute for the APIClient

### LINKED ISSUE #153: cache last response
It might be nice to save the last response in the `APIClient` class so that users could retrieve it if they forget to assign it when using the client in a REPL or notebook.

e.g., 

```
> client = civis.APIClient()
> client.scripts.post_containers(...)
> client.last_response
{"id": ..., }
```

This idea came from a comment by @skomanduri.  I can't think of drawbacks to this, but perhaps there are.

stephen-hoover:
As a convenience, keep track of the last response received from the Civis API.

Closes #153 .


keithing:
Since this is touch the root `Endpoint` class, would you mind running `tools/smoke_test.py`?

stephen-hoover:
@keithing , the smoke test completes successfully (and I merged in master and fixed the CHANGELOG conflicts).

PR #304, opened 2019-06-04
ENH dataframe and json to-file conveniences

### LINKED ISSUE #262: Convenience function for persisting dataframe to files endpoint
`civis.io` has a function `file_to_dataframe`, but no function going in the opposite direction. I find myself rewriting this simple function for myself a lot, and I thought it would make a nice little convenience for others. @keithing, would you be interested in including something like this in the `civis.io` function suite (I would make a PR)?

stephen-hoover:
We have functions to download Civis File objects and convert them to a DataFrame or a JSON object. Add functions to go the other way. In particular, users often expect the `dataframe_to_file` function and are surprised that it doesn't exist.

Closes #262 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Added
+- `civis.io.dataframe_to_file` and `civis.io.json_to_file` convenience functions.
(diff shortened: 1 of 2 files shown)


keithing:
cc @channiemills

channiemills:
@keithing and I were just talking about the need for this method today. I too was surprised it didn't already exist!

PR #305, opened 2019-06-04
ENH utils module available from civis namespace

stephen-hoover:
Import `utils` into `civis` so that users can access `civis.utils.run_job` after an `import civis` import. Many users expect this and are surprised when it's not available.


keithing:
Anything else you're planning to add to this PR? I'm happy to review.

stephen-hoover:
No, that was it. I was trying to avoid asking for review on too many PRs at once.

PR #308, opened 2019-06-07
TST Test in Xenial environment and include Python 3.8-dev

stephen-hoover:
Python versions >=3.7 require OpenSSL 1.0.2+, which is not available in Trusty (the default environment). Switch to using Xenial (Ubuntu 16.04). Add the Python 3.8 dev branch so that we can make sure we're compatible when it releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,12 @@
+dist: xenial
 language: python
 python:
     - "2.7"
     - "3.4"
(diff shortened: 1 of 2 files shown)


stephen-hoover:
Of course wheels aren't available for Python 3.8-dev, and we hit an error compiling numpy from source. It's not worth trying to make it compile.

PR #311, opened 2019-06-10
COMPAT Deprecation decorator takes multiple parameter names

stephen-hoover:
In Python v3, multiple layers of the deprecation decorator work as expected. But we run into problems with Python v2.7 (apparently due to missing features in the `Signature` class). If the decorator takes multiple parameter names, it's more complex, but will be compatible with more Python versions.


stephen-hoover:
@keithing , I ran into this problem when doing #307 . I could also do the deprecation there without using this decorator, in which case I don't need this PR. Let me know if you prefer that.

stephen-hoover:
Yes, that's right.

PR #312, opened 2019-06-13
Allow changing log level & base URL for CLI with environment variables 

aripollak:


civis/cli/__main__.py
```diff
@@ -33,8 +34,7 @@
 
 
 _REPLACEABLE_COMMAND_CHARS = re.compile(r'[^A-Za-z0-9]+')
-_API_URL = "https://api.civisanalytics.com"
-_OPENAPI_SPEC_URL = "https://api.civisanalytics.com/endpoints"
```
keithing: Is this removed because it is not used anywhere?

aripollak: Yup, I couldn't find it used anywhere.

stephen-hoover: PR #117 refactored how the CLI gets the API spec, and `_OPENAPI_SPEC_URL`'s use was removed there. It looks like I neglected to remove the unused parameter when I did that. Thanks for the cleanup!

PR #314, opened 2019-08-02
ENH helper function to list models

elsander:
This PR adds a helper function to list a user's models (training, prediction, or both).

### DIFF
--- a/civis/ml/_helper.py
+++ b/civis/ml/_helper.py
@@ -14,7 +14,7 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
 
     Parameters
     ----------
-    job_type : {None, "train", "predict"}
+    job_type : {"train", "predict", None}
         The type of model job to list. If "train", list training jobs
         only (including registered models trained outside of CivisML).
         If "predict", list prediction jobs only. If None, list both.
--- a/civis/ml/_helper.py
+++ b/civis/ml/_helper.py
@@ -51,6 +51,9 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
     if author is SENTINEL:
         author = client.users.list_me().id
 
+    # default to showing most recent models first
+    kwargs.setdefault('order_dir', 'desc')
+
     models = client.scripts.list_custom(from_template_id=template_id_str,
                                         author=author,
                                         **kwargs)
--- a/civis/ml/_helper.py
+++ b/civis/ml/_helper.py
@@ -1,9 +1,16 @@
+from collections import namedtuple
+
 from civis import APIClient
 from civis.ml._model import _PRED_TEMPLATES
 
+__all__ = ['list_models']
 
-def list_models(job_type=None, client=None, **kwargs):
-    """List the current user's CivisML models.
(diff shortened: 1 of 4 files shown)


civis/ml/_helper.py
```diff
@@ -0,0 +1,40 @@
...
+    else:
+        raise ValueError("Parameter 'job_type' must be None, 'train', "
+                         "or 'predict'.")
+    template_id_str = ', '.join([str(tmp) for tmp in template_id_list])
+
+    if client is None:
+        client = APIClient()
+
+    models = client.scripts.list_custom(from_template_id=template_id_str,
+                                        author=client.users.list_me().id,
```
stephen-hoover: I suggest making `author` a parameter and defaulting to the current user. People may wish to list someone else's models.

civis/ml/_helper.py
```diff
@@ -0,0 +1,40 @@
...
+    job_type : {None, "train", "predict"}
+        The type of model job to list. If "train", list training jobs
+        only (including registered models trained outside of CivisML).
+        If "predict", list prediction jobs only. If None, list both.
+    client : :class:`civis.APIClient`, optional
+        If not provided, an :class:`civis.APIClient` object will be
+        created from the :envvar:`CIVIS_API_KEY`.
+    **kwargs : kwargs
+        Extra keyword arguments passed to `client.scripts.list_custom()`
+    """
```
stephen-hoover: Maybe a 
```
See Also
--------
APIClient.scripts.list_custom
```
?

civis/ml/_helper.py
```diff
@@ -0,0 +1,40 @@
+from civis import APIClient
+from civis.ml._model import _PRED_TEMPLATES
+
+
+def list_models(job_type=None, client=None, **kwargs):
```
stephen-hoover: What would you think of defaulting to `"train"`? When I think of a "model", usually it's the training job rather than the scoring job.

civis/ml/tests/test_helper.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_list_models_bad_job_type():
+    with pytest.raises(ValueError):
+        list_models(job_type="fake")
+
+
+def test_list_models():
+    resp = [Response({'id': 2834, 'name': 'RFC model'})]
+    m_client = mock.Mock()
```
stephen-hoover: This is a good place to use the (poorly documented) `create_client_mock` function, https://github.com/civisanalytics/civis-python/blob/v1.10.0/civis/tests/mocks.py#L15 .

civis/ml/_helper.py
```diff
@@ -0,0 +1,40 @@
+from civis import APIClient
+from civis.ml._model import _PRED_TEMPLATES
+
+
```
stephen-hoover: If the `__init__` is doing an `import *`, then we should have this module control what gets imported.
```suggestion
__all__ = ['list_models']
```

civis/ml/_helper.py
elsander: Yeah, I went back and forth between defaulting to "train" vs None. Happy to change it.

PR #315, opened 2019-08-09
ENH Model sharing helpers

stephen-hoover:
A CivisML model has File and Project run outputs, each of which need to be shared to effectively share the "model". This helper will transparently handle sharing the run outputs. Include JSONValue in case a future version of CivisML uses that. The functions are patterned after the autogenerated API sharing endpoints.

### DIFF
--- a/civis/ml/_helper.py
+++ b/civis/ml/_helper.py
@@ -1,9 +1,12 @@
 from collections import namedtuple
+import logging
 
 from civis import APIClient
 from civis.ml._model import _PRED_TEMPLATES
 
-__all__ = ['list_models']
+__all__ = ['list_models', 'put_models_shares_groups',
+           'put_models_shares_users']
+log = logging.getLogger(__name__)
 
 # sentinel value for default author value
 SENTINEL = namedtuple('Sentinel', [])()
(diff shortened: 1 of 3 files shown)


stephen-hoover:
@elsander , see what you think of the naming and other design choices here. It's not obvious what the right interface would be. I considered adding a `list_models_shares` function as well, but doing that right would be really complex, and I wasn't sure it was worth it.

civis/ml/_helper.py
```diff
@@ -58,3 +62,261 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
...
+                _func = getattr(client.files, "put_shares_" + entity_type)
+                if _output['name'] == 'log.txt' and permission_level == 'read':
+                    # Require "write" level permission or higher to view
+                    # the debug logs.
+                    continue
+                obj_permission = permission_level
+            elif _output['object_type'] == 'Project':
+                _func = getattr(client.projects, "put_shares_" + entity_type)
+                if permission_level == 'read':
+                    # Users must be able to add to projects to use the model
```
elsander: Just for my own understanding, why is this?

civis/ml/_helper.py
```diff
@@ -58,3 +62,261 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
...
+                    obj_permission = 'write'
+                else:
+                    obj_permission = permission_level
+            elif _output['object_type'] == 'JSONValue':
+                _func = getattr(client.json_values,
+                                "put_shares_" + entity_type)
+                obj_permission = permission_level
+            else:
+                continue
```
elsander: Would it make sense to issue a message to the debug log if you hit this condition? 

Also, are there any outputs where we'd expect to hit this condition? If you write OOS scores to a table, does that show up as a run output?

civis/ml/_helper.py
```diff
@@ -58,3 +62,261 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
...
+        log.debug("Unsharing outputs on %d, run %s.", job_id, run.id)
+        outputs = client.scripts.list_containers_runs_outputs(job_id, run.id)
+        for _output in outputs:
+            if _output['object_type'] == 'File':
+                _func = getattr(client.files, endpoint_name)
+            elif _output['object_type'] == 'Project':
+                _func = getattr(client.projects, endpoint_name)
+            elif _output['object_type'] == 'JSONValue':
+                _func = getattr(client.json_values, endpoint_name)
+            else:
```
elsander: Same note here, would it be useful to log skipped outputs?

civis/ml/_helper.py
```diff
@@ -58,3 +62,261 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
...
+        If not provided, an :class:`civis.APIClient` object will be
+        created from the :envvar:`CIVIS_API_KEY`.
+    share_email_body : string, optional
+        Custom body text for e-mail sent on a share.
+    send_shared_email : boolean, optional
+        Send email to the recipients of a share.
+
+    Returns
+    -------
+    readers : dict::
```
elsander: I'm confused by the type annotation here. It looks like "users" and "groups" are lists of dicts, right? Should this say `list[dict]` to specify that explicitly?

elsander:
I think the function names make a lot of sense. I could see the list function being useful, but if it's complex to implement it's probably not worth doing right now. If we hear requests for it we could reevaluate.

civis/ml/tests/test_helper.py
```diff
@@ -22,3 +26,164 @@ def test_list_models():
...
+    m_client = _create_share_model_client_mock([11])
+
+    helper._share_model(3, [7, 8], 'read', 'groups', client=m_client)
+
+    m_client.projects.put_shares_groups.assert_called_once_with(
+        31, [7, 8], 'write', send_shared_email=False)
+
+
+def test_share_model_debug_log():
+    # Debug logs need "write" permission to be shared
```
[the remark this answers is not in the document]
elsander: I think you're right in most cases, a person who only needs read permissions won't need access to the debug logs. But I don't think it would ever be a problem to expose the debug logs, and there are some cases where you might expect read permissions to provide log access. Probably the most common is CS sharing one of us on a client job so that we can examine the logs and debug, but I could also imagine clients sharing jobs to debug internally, before they escalate to CS. My preference is to share all run outputs for consistency, but I'm open to discussion.

civis/ml/_helper.py
elsander: Hmm, I didn't think about whether a "read" user should be able to score based off the original model rather than a clone. That reasoning makes sense to me, thanks.

Follow-up question: what permissions does the original user have on a scoring job created by the new user in this way?

stephen-hoover: The original user wouldn't have any special permissions on the scoring job by default. If you don't have "read" permission on the scoring job, it wouldn't be visible to you in the project. I think this is expected behavior -- I could imagine the "scoring" project being filled with scoring jobs, and no one user being able to see all of them.

civis/ml/tests/test_helper.py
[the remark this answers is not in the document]
stephen-hoover: Okay, this makes the code simpler, so I'm happy to make the change. I'm not thinking of anything sensitive in the logs which wouldn't also be in the other model artifacts which would be shared at read level.

civis/ml/_helper.py
elsander: Sounds good. I agree that this is expected behavior.

civis/ml/_helper.py
stephen-hoover: The `readers` is a single dictionary. It contains keys `users` and `groups`, each of which is a list of dictionaries. In Python notation, I think you're right that the types of `users` and `groups` would be `list[dict]`, but I think that what's show here is also correct, and consistent with the type notation used in the other API endpoints.

stephen-hoover:
@elsander , I addressed your comments.

Issue #228, opened 2018-02-15
Make the Response object immutable


stephen-hoover:
The `civis.response.Response` object subclasses `dict`, but it also copies its dictionary keys into attributes. I've seen code alternate between accessing its contents as attributes and as items. This is mostly okay, but as soon as you try to alter the contents, the two ways of accessing data get out of sync. If we want to keep the dual object / dictionary identity (we may not), then we should make the object, or at least the attributes which are part of the dictionary, immutable.

See this confusing exchange:
```python
In [25]: resp = civis.response.Response({'spam': 'eggs'})

In [26]: resp.spam
Out[26]: 'eggs'

In [27]: resp['spam']
Out[27]: 'eggs'

In [28]: resp['spam'] = 'foo'

In [29]: resp['spam']
Out[29]: 'foo'

In [30]: resp.spam
Out[30]: 'eggs'

In [31]: resp.spam = 'bar'

In [32]: resp['spam']
Out[32]: 'foo'

In [33]: resp
Out[33]: {'spam': 'foo'}
```

PR #318, opened 2019-08-16
[SERD-1446] ENH Added run_template function

arthir:
Mimics the civis-R run_template function.


civis/io/_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+    fut.result()
+    outputs = client.scripts.list_containers_runs_outputs(job.id, run.id)
+    if JSONValue:
+        json_output = [
+            o.value for o in outputs if o.object_type == "JSONValue"
+        ]
+        if len(json_output) > 1:
+            raise RuntimeError(
+                'Error in returning JSON outputs for template {}'
+                'run -- too many JSON outputs'.format(id)
```
stephen-hoover: The first string needs an extra space, and I suspect that the `format` will fail. I think the strings are concatenated after the function call.

civis/io/_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+            )
+            return ()
```
stephen-hoover: You'll never get to this line because of the exception above.
```suggestion
```

civis/io/_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+        if len(json_output) == 0:
+            log.info('No JSON output for template {}'.format(id))
```
stephen-hoover: I think that `debug` is more appropriate here.
```suggestion
            log.debug('No JSON output for template {}'.format(id))
```

civis/io/_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+            return ()
```
stephen-hoover: If you want to return `None`, then a bare `return` will do it. An empty tuple would be `(,)`, but I think that `None` is appropriate here.
```suggestion
            return
```

civis/io/_scripts.py
```diff
@@ -0,0 +1,63 @@
...
+        JSON object containing the results of the template run.
+        Expects only a single JSON result. Will return nothing if
+        either there is no JSON result or there is more than 1 JSON result.
+
+    Examples
+    --------
+    >>> # Run template to return file_ids
+    >>> run_template(my_template_id, arguments=my_dict_of_args)
+    >>> # Run template to return JSON output
+    >>> run_template(my_template_id, arguments=my_dict_of_args, JSONValue=True)
```
stephen-hoover: I like the examples. Could you also show examples of what the output might look like? That would be helpful for me as a user trying to understand what this function does.

stephen-hoover:
Almost forgot -- this would also need an entry in the changelog.

civis/io/_scripts.py
arthir: I wasn't actually sure what makes sense in this case - `None`, empty object but your comment clarifies that.

civis/io/_scripts.py
arthir: Added a space. 
I was assuming that this would work since a similar error message exists in _files.py (line 321)
`
    if not url:
        raise EmptyResultError('Unable to locate file {}. If it previously '
                               'existed, it may have '
                               'expired.'.format(file_id))
`

stephen-hoover: I just tried it out. You're correct, this does work.

civis/io/_scripts.py
```diff
@@ -0,0 +1,62 @@
...
+        json_output = [
+            o.value for o in outputs if o.object_type == "JSONValue"
+        ]
+        if len(json_output) > 1:
+            raise RuntimeError(
+                'Error in returning JSON outputs for template {}'
+                ' -- too many JSON outputs'.format(id)
+            )
+        if len(json_output) == 0:
+            log.debug('No JSON output for template {}'.format(id))
```
keithing: What do you think about a warning here instead of a debug message?

civis/io/_scripts.py
```diff
@@ -0,0 +1,62 @@
...
+    outputs = client.scripts.list_containers_runs_outputs(job.id, run.id)
+    if JSONValue:
+        json_output = [
+            o.value for o in outputs if o.object_type == "JSONValue"
+        ]
+        if len(json_output) > 1:
+            raise RuntimeError(
+                'Error in returning JSON outputs for template {}'
+                ' -- too many JSON outputs'.format(id)
+            )
```
keithing: Similar comment to what I noted in the companion R client PR: What do you think about returning the first element and warning rather than failing?

civis/io/_scripts.py
```diff
@@ -0,0 +1,62 @@
...
+        If False, will return the file ids associated with the
+        output results.
+
+    Returns
+    -------
+    If JSONValue = False:
+      file_ids : dict
+        Dictionary of file ids with the keys being their output names.
+    If JSONValue = True:
+      json_output : civis.response.Response
```
keithing: does this return a `civis.response.Response` or a `dict`?  I think we would prefer the latter.

arthir: as far as I can tell, `civis.response.Response` is a dict? At least it seems to behave as one. The reason I wrote `civis.response.Response` here is that's the class of the returned object from `client.scripts.list_containers_runs_outputs`

civis/io/_scripts.py
arthir: Changed to warning.

civis/io/_scripts.py
arthir: Sounds good; changed to match civis-r behavior.

civis/io/_scripts.py
keithing: Isn't there a call to subset that output to `value` so it's no longer a `civis.response.Response`?

arthir: Oddly, `type(outputs[0].value)` is still `<class 'civis.response.Response'>`

stephen-hoover: `Response` is a subclass of `dict` (see e.g. #228 ). I suggest casting the result as a `dict` before returning it.

civis/io/_scripts.py
```diff
@@ -0,0 +1,62 @@
...
+    fut = CivisFuture(client.scripts.get_custom_runs, (job.id, run.id))
+    fut.result()
+    outputs = client.scripts.list_containers_runs_outputs(job.id, run.id)
+    if JSONValue:
+        json_output = [
+            o.value for o in outputs if o.object_type == "JSONValue"
+        ]
+        if len(json_output) == 0:
+            log.warning('No JSON output for template {}'.format(id))
+            return
```
keithing: I think most users would expect an empty dict here instead of `None`.  That would allow them to call things like

```python
o = run_template(...)
x = o.get("my_key", "default")
```
instead of

```python
o = run_template(...)
x = o.get("my_key", "default") if x else "default"
```

stephen-hoover: Counterpoint -- "output which lacks my data" is different from "no output", and I'd like to be able to distinguish them.

civis/io/_scripts.py
arthir: Interesting. 
Changed it to be a dict.

civis/io/_scripts.py
keithing: Yeah, that's absolutely right. What sorts of use cases could you imagine for needing to know that distinction?  I suppose from a user perspective I'd be happy to lose that information for the convenience of always expecting a `dict`.

arthir:
> Thanks for adding this! High level comments -- Keith may have other opinions, but I think that `civis.utils._jobs` is the most natural place to put this new function. It's not IO related. It should also take an `APIClient` object as its last parameter, defaulting to `None`. The other helper functions use this pattern, and it does make testing much easier.

Moved the function; also added the client as the last parameter.

keithing:
@arthir and I talked about this online, but we agree with @stephen-hoover that `civis.utils._jobs` is the right place for this function.  As a reminder, once we move that we'll need to add this function to `__all__` in `civis.utils.__init__` in order to import it with `civis.utils.run_template`.

civis/utils/_jobs.py
```diff
@@ -1,7 +1,11 @@
+import logging
```
keithing: Great, you've already moved this over to `_jobs.py`.  Will you expose this in `civis.utils.__init__` so it can be imported with `from civis.utils import run_template` ?

arthir: Done

arthir:
> Almost forgot -- this would also need an entry in the changelog.

Added to the changelog (I'm not sure if there's a specific style to follow).

civis/utils/_jobs.py
```diff
@@ -30,3 +34,63 @@ def run_job(job_id, api_key=None, client=None):
...
+    fut = CivisFuture(client.scripts.get_custom_runs, (job.id, run.id))
+    fut.result()
+    outputs = client.scripts.list_containers_runs_outputs(job.id, run.id)
+    if JSONValue:
+        json_output = [
+            o.value for o in outputs if o.object_type == "JSONValue"
+        ]
+        if len(json_output) == 0:
+            log.warning('No JSON output for template {}'.format(id))
+            return
```
keithing: I don't too strongly about `None` vs `dict()`, let's move forward with `None` for now per @stephen-hoover's point.

arthir: Sounds good.

jacksonllee:
How is the user going to discover this awesome new function? Would it be documented in the Sphinx docs? Relatedly, I could also see the `run_job` function is arguably undocumented (though [subtly mentioned](https://civis-python.readthedocs.io/en/v1.10.0/user_guide.html?highlight=run_job)). Would both `run_X` functions deserve a small section on the [User Guide](https://civis-python.readthedocs.io/en/v1.10.0/user_guide.html) page, for example?

arthir:
> How is the user going to discover this awesome new function? Would it be documented in the Sphinx docs? Relatedly, I could also see the `run_job` function is arguably undocumented (though [subtly mentioned](https://civis-python.readthedocs.io/en/v1.10.0/user_guide.html?highlight=run_job)). Would both `run_X` functions deserve a small section on the [User Guide](https://civis-python.readthedocs.io/en/v1.10.0/user_guide.html) page, for example?

I can add (both) to the docs. Thanks for pointing it out.

arthir:
> Code looks good to me. Before merging would you mind
> 
> 1. Adding a test
> 2. Addressing @stephen-hoover's comment about examples in the docstring.

- Added some tests for run_template (currently there are none for run_job())
- Also added the outputs to the examples in the docstring

arthir:
@keithing Tests added (and work!) Also docs.

civis/utils/_jobs.py
```diff
@@ -1,32 +1,102 @@
...
+        ]
+        if len(json_output) == 0:
+            log.warning("No JSON output for template {}".format(id))
+            return
+        if len(json_output) > 1:
+            log.warning(
+                "More than 1 JSON output for template {}"
+                " -- returning only the first one.".format(id)
+            )
+        return dict(json_output[0])
```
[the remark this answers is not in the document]
arthir: Add a comment for this (in the code)

arthir:
@keithing @stephen-hoover I think I need to re-request reviews after that last comment change.

PR #324, opened 2019-09-10
ENH Optional `name` arg in civis.io.file_to_civis

jacksonllee:
This main issue that this PR aims to address is to avoid having to write client code like `civis.io.file_to_civis('foobar.txt', 'foobar.txt')`, where the second arg `name` is obligatory currently (as of v1.11.0) but seems unnecessary. I think it'd be perfectly reasonable to allow code like `civis.io.file_to_civis('foobar.txt')` where `name` can internally be inferred from the basename of the given file path -- this is what this PR implements.

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -183,8 +183,8 @@ def file_to_civis(buf, name=None, api_key=None, client=None, **kwargs):
         treated as paths to local files to open.
     name : str, optional
         The name you wish to give the file. If not given, it will be inferred
-        from the basename of `buf` (if `buf` is a string for a file path) or
-        `buf.name` (if `buf` is a file-like object).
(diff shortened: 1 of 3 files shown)


civis/io/_files.py
```diff
@@ -198,10 +200,17 @@ def file_to_civis(buf, name, api_key=None, client=None, **kwargs):
     file_id : int
         The new Civis file ID.
 
+    Raises
+    ------
+    ValueError
```
stephen-hoover: I think this should be a `TypeError` -- it looks like you get it when the type of the input is incorrect (i.e., a file-like object with no `name`), yes?
```suggestion
    TypeError
```

jacksonllee: Agreed that `TypeError` makes more sense here, as it's more like a "missing an argument" error. Code updated at c6fc7a3.

PR #327, opened 2019-10-22
ADD: CLI command for civis username

channiemills:
I recently became aware of the CLI and the command `civis users list-me` which I now regularly use to see which CIVIS_API_KEY I have active. `list-me` returns a lot of information about my current user, but for my purposes the username would be sufficient. This PR adds a command to the CLI to print out the response from `civis.APIClient().username`

### DIFF
--- a/civis/cli/__main__.py
+++ b/civis/cli/__main__.py
@@ -25,8 +25,8 @@ import click
 from jsonref import JsonRef
 import yaml
 from civis.cli._cli_commands import (
-    civis_ascii_art, files_download_cmd, files_upload_cmd,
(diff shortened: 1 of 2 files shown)


channiemills:
Hi, I'm new to this repo. @stephen-hoover or @mheilman would either of you be available to review a small PR?

stephen-hoover:
Thank you for the suggestion! Instead of adding a new special command, could you handle this use case with `civis users list-me|grep username`?

channiemills:
I definitely could, thanks for the recommendation!

PR #328, opened 2019-10-23
[CIVP-19296] Update Python Client to use CSVImports

zacharydfriedlander:
In this PR:

* We migrate the convenience functions for imports in `civis.io` from using `AutoImport`s via the `/imports` and `/imports/syncs` endpoints to using the newly created `/imports/files/csv` and `/files/preprocess/csv` endpoints. This opens up some new opportunities for users via these endpoints, such as upserts, multi-file imports, and importing with a subset of columns.

* As part of developing these new capabilities, a number of the tests for the `civis.io` module are also refactored from using VCR cassettes to mocking API client calls directly.

* We add a new `get_storage_host_id` method to the `APIClient`. This method makes easier interacting with the new Storage hosts abstraction, used by the new CSV imports to import directly from S3.


elsander:
@zacharydfriedlander, can you pull in @leeacto for review as well?

Also tagging @patr1ckm for viz.

civis/civis.py
```diff
@@ -287,6 +287,44 @@ def get_table_id(self, table, database):
...
+        1234
+
+        >>> client.get_storage_host_id(1111)
+        1111
+        """
+        if isinstance(storage_host, int):
+            return storage_host
+        sh = find_one(self.storage_hosts.list(), name=storage_host)
+        if not sh:
+            raise ValueError("Storage Host {} not found.".format(storage_host))
```
leeacto: f-string?

civis/civis.py
```diff
@@ -287,6 +287,44 @@ def get_table_id(self, table, database):
...
+
+        return sh["id"]
```
leeacto: I'm guessing it's fine, but any risk of a KeyError here?

civis/civis.py
zacharydfriedlander: As much as I would love to, we can't in order to support Python 2.7, 3.4, or 3.5. When we drop support for those, f-strings are definitely of interest.

leeacto: oh snap. yeah makes sense then

civis/io/_tables.py
```diff
@@ -819,24 +858,29 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
                         max_errors=None, existing_table_rows="fail",
                         diststyle=None, distkey=None,
                         sortkey1=None, sortkey2=None,
-                        delimiter=",", headers=None,
+                        primary_keys=None, last_modified_keys=None,
+                        escaped=False, execution="immediate",
+                        delimiter=None, headers=None,
                         credential_id=None, polling_interval=None,
                         hidden=True):
     """Upload the contents of a Civis file to a Civis table.
```
patr1ckm: maybe `one or more Civis files`

civis/civis.py
elsander: I believe the API client avoids f-strings in general to allow python <3.6 compatibility (although it's possible this isn't strictly enforced throughout the repo anymore)

civis/tests/test_civis.py
```diff
@@ -25,3 +25,26 @@ def test_get_table_id(schema_tablename):
...
+
+    storage_hosts = [StorageHost(1234, 'test'), StorageHost(5678, 'othertest')]
+    client.storage_hosts.list = mock.Mock(return_value=storage_hosts)
+
+    assert client.get_storage_host_id('test') == 1234
+
+    client.storage_hosts.list.assert_called_once_with()
+
+    assert client.get_storage_host_id(4732) == 4732
+    with pytest.raises(ValueError, match="Storage Host invalidname not found"):
```
elsander: I didn't know about this "match" argument to `pytest.raises`. TIL!

civis/io/_tables.py
```diff
@@ -618,7 +615,9 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
                        headers=None, credential_id=None,
-                       polling_interval=None,
+                       primary_keys=None, last_modified_keys=None,
+                       execution="immediate",
+                       delimiter=None, polling_interval=None,
                        archive=False, hidden=True, **kwargs):
     """Upload a `pandas` `DataFrame` into a Civis table.
```
patr1ckm: I can't add a comment to a hidden part of the PR for some reason, but the docs for `existing_table_rows` should have `upsert` as an option as well

civis/io/_tables.py
```diff
@@ -774,6 +796,23 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+        A boolean value indicating whether or not the source file has quotes
+        escaped with a backslash. Defaults to false.
+    execution: string, optional, default "immediate"
+        One of "delayed" or "immediate". If "immediate", refresh column
+        statistics as part of the run. If "delayed", flag the table for a
+        deferred statistics update; column statistics may not be available
+        for up to 24 hours. In addition, if existing_table_rows is "upsert",
+        delayed executions move data from staging table to final table after a
+        brief delay, in order to accommodate multiple concurrent imports to the
+        same destination table.
```
patr1ckm: Just to check my understanding:

Should `execution` always be `delayed` if passing multiple file ids when `existing_table_rows = "upsert"`? 
Could there ever be a situation when the delay is not long enough when upserting? What happens from the user perspective in that case?

civis/io/_tables.py
patr1ckm: Would it be worth documenting `CivisImportError`? And maybe documenting under what circumstances errors are thrown when attempting to upload multiple file ids?

civis/io/_tables.py
```diff
@@ -819,24 +858,29 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+def civis_file_to_table(file_ids, database, table, client=None,
                         max_errors=None, existing_table_rows="fail",
                         diststyle=None, distkey=None,
                         sortkey1=None, sortkey2=None,
-                        delimiter=",", headers=None,
+                        primary_keys=None, last_modified_keys=None,
+                        escaped=False, execution="immediate",
+                        delimiter=None, headers=None,
                         credential_id=None, polling_interval=None,
                         hidden=True):
```
patr1ckm: Is there any case when it would be helpful to expose options for `files.post_process_csv` such as `force_character_set_conversion`?

civis/io/_tables.py
```diff
@@ -618,7 +615,9 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
                        headers=None, credential_id=None,
-                       polling_interval=None,
+                       primary_keys=None, last_modified_keys=None,
+                       execution="immediate",
```
leeacto: `"immediate"` seems like it should be a constant somewhere

civis/io/_tables.py
zacharydfriedlander: I don't think so. These options are set the way they are in order to achieve parity with what `AutoImport`s were/are doing under the hood. Exposing them to users would add to the complexity of an already-massive function signature for little benefit. If there's a specific use case in mind, then we should discuss further, but if it's just for the sake of having another parameter, then in my opinion we don't gain enough to justify the cost.

civis/civis.py
zacharydfriedlander: I don't believe so. This code is modeled off of that used for `get_database_id` and `get_database_credential_id` above, and the API spec should take care of guaranteeing that ID is a valid key.

civis/io/_tables.py
patr1ckm: cleared this up offline

civis/io/_tables.py
patr1ckm: Awesome, thanks!

civis/io/_tables.py
zacharydfriedlander: Talked offline, but leaving comments for posterity/general understanding:

1. `Execution` can be both `delayed` and `immediate` with multiple files - the parameter is more to control the interplay of multiple jobs writing to the same table concurrently.
2. In `delayed` mode, each subsequent attempt to upsert pushes back the final transfer of data - data is moved from staging to final table 15 minutes after the last requested upsert.

zacharydfriedlander:
If y'all have any questions about the overall functionality of the `imports/files/csv` endpoint, [these Zendesk docs](https://civis.zendesk.com/hc/en-us/articles/360027016851-Advanced-CSV-Imports-via-the-Civis-API) might be helpful - should probably have included these from the outset!

civis/io/_tables.py
zacharydfriedlander: Good catch. I will update this.

civis/io/_tables.py
zacharydfriedlander: Good catch - I will update. And yes, we should for sure - I can add a `Raises` section to the docstring here?

civis/io/_tables.py
```diff
@@ -1094,3 +1168,122 @@ def split_schema_tablename(table):
...
+                                   delimiter,
+                                   output_file.id,
+                                   detected_info["columnDelimiter"])
+                               )
+
+    compression = detected_info['compression']
+
+    cleaned_file_ids.append(output_file.object_id)
+
+    for result in concurrent.futures.as_completed(done | still_going):
```
elsander: Not sure I understand what this line is doing. Is the goal to iterate over completed futures first to give the ongoing ones more time? A comment would be helpful.

civis/io/_tables.py
```diff
@@ -1094,3 +1168,122 @@ def split_schema_tablename(table):
...
+            file_columns = detected_info['tableColumns']
+            if table_columns != file_columns:
+                raise CivisImportError('All files should have the same '
+                                       'schema. Expected {}'
+                                       'but file {} has {}'.format(
+                                           table_columns,
+                                           output_file.object_id,
+                                           file_columns)
+                                       )
+        if headers != detected_info['includeHeader']:
```
elsander: This seems like a lot of repeated code from the checks on the first file. You could check for `None` separately in the first set of conditionals, then put this set of checks into a helper function that gets called here and above.

civis/tests/test_io.py
```diff
@@ -144,42 +149,353 @@ def test_large_file_to_civis(self, *mocks):
...
+    ):
         table = "scratch.api_client_test_fixture"
         database = 'redshift-general'
-        result = civis.io.civis_file_to_table(self.file_id, database, table,
-                                              existing_table_rows='truncate',
-                                              polling_interval=POLL_INTERVAL)
-        result = result.result()  # block until done
+        mock_file_id = 1234
+        mock_cleaned_file_id = 1235
+        mock_import_id = 8675309
```
elsander: Ha, glad to see someone else uses this as a mock number (although you might be copying it from my code in the client, I suppose)

civis/tests/test_io.py
```diff
@@ -206,29 +524,56 @@ def test_read_civis_sql(self, *mocks):
                                        polling_interval=POLL_INTERVAL)
         assert data == expected
 
+    @pytest.mark.dataframe_to_civis
     @pytest.mark.skipif(not has_pandas, reason="pandas not installed")
-    @mock.patch(api_import_str, return_value=civis_api_spec)
-    def test_dataframe_to_civis(self, *mocks):
```
elsander: I found this older test easier to read, and it checks the result output rather than implementation/call arguments. Can you talk about why these tests were updated?

civis/tests/test_io.py
```diff
@@ -245,6 +590,7 @@ def test_civis_to_multifile_csv(self, *mocks):
                                                              'console.s3.'
                                                              'amazonaws.com/')
 
+    @pytest.mark.civis_to_multifile_csv
```
elsander: Thanks for adding these marks, great idea.

civis/io/_tables.py
elsander: Oops, not all of these comments showed up while I was reviewing. Adding a `Raises` section is a good idea.

civis/io/_tables.py
```diff
@@ -671,6 +670,23 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
     credential_id : str or int, optional
         The ID of the database credential.  If ``None``, the default
         credential will be used.
+    primary_keys: list[str], optional
```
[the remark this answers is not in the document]
zacharydfriedlander: Actually, this is relevant regardless of databases for imports in `upsert` mode. Our Platform-side upsert logic requires primary keys (in order to identify common records between staging and final tables) and last-modified columns (to determine which record is newer). Happy to talk more offline about this.

civis/io/_tables.py
```diff
@@ -956,31 +1042,19 @@ def _get_sql_select(table, columns=None):
 
 
 def _get_headers(client, sql, database, credential_id, polling_interval=None):
-    # use 'begin read only;' to ensure we can't change state
```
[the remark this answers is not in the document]
zacharydfriedlander: The latter; specifically, the assumption was made that Redshift would return results as an escaped-delimited file rather than a true CSV. This was, as I understand, only true when a a specific Platform feature flag was active. For more info please see the [JIRA for this change](https://civisanalytics.atlassian.net/browse/CIVP-19104).

civis/io/_tables.py
```diff
@@ -819,24 +858,29 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
 
     Parameters
     ----------
-    file_id : int
-        Civis file ID.
+    file_ids : int or list[int]
+        Civis file ID or a list of Civis file IDs.
     database : str or int
         Upload data into this database. Can be the database name or ID.
     table : str
```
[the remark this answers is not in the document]
zacharydfriedlander: `max_errors` is applied to the load as a whole, so across all files.

CHANGELOG.md
```diff
@@ -5,12 +5,18 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
+- Added method `get_storage_host_id` to the APIClient
 
 ### Fixed
 - Fixed requirements.txt listing for `cloudpickle` -- `>=0.2`, not `<=0.2`. (#323)
+- Fixed issue in `civis.io.read_civis_sql` when returning data that contains 
+  double quotes
 
 ### Changed
 - Allow the `name` arg to be optional in `civis.io.file_to_civis`. (#324)
+- Refactor `civis.io.civis_file_to_table` to use a new set of Civis API endpoints.
```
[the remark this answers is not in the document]
zacharydfriedlander: Sounds good - updated.

civis/io/_tables.py
zacharydfriedlander: The idea is to make sure that we have accounted for all files - since `concurrent.futures.wait` returns a pair of sets, we may end up with multiple cleaners that have finished "at the same time" (i.e. when their futures poll again). By combining the set of `done`s with the set of `still_running`s, we make sure we account for all files.

I will add a comment to this effect clarifying the intent.

civis/tests/test_io.py
zacharydfriedlander: The previous cassette-based tests were difficult to update to account for the use of new endpoints; in particular, re-recording the cassettes proved a challenge, and attempts to learn more about the process of doing so yielded little result. 

This test does seem unwieldy; however, I think part of that is making explicit the stubbing behavior that was previously encapsulated by the cassettes. I am happy to take another pass at simplifying this test to recover some of its prior readability, perhaps by removing some of the mocking around the `DataFrame` and `TemporaryDirectory` in favor of live objects.

civis/io/_tables.py
zacharydfriedlander: > Do the files get joined together (presumably not)?

You are correct; the files do not get concatenated, but rather [entered into a manifest](https://docs.aws.amazon.com/redshift/latest/dg/loading-data-files-using-manifest.html) which is passed to the underlying redshift `COPY` statement. 

> Are they assumed to have exactly the same set of columns in exactly the same order?

Yes - as laid out in the [Helpdesk docs for the new endpoint](https://civis.zendesk.com/hc/en-us/articles/360027016851-Advanced-CSV-Imports-via-the-Civis-API), files should share the same column schema and file format (e.g. same delimiter, encoding, compression, etc.). Some of this can be taken care of by the cleaning step, but users should still take care.

> Are they guaranteed to get appended in the order they're passed in?

I do not believe this is the case, as the load into Redshift happens in parallel.

> If the very last file id fails, do you get an error and no table or an error and a table with all of the files but the last one?

This is addressed below, but the COPY statement is atomic - if a single file fails, the entire load fails, and therefore no table is created. 


The fact that there are so many questions here indicates that this docstring is not nearly sufficient. I will add additional detail to the overall summary as well as the relevant parameters.

civis/io/_tables.py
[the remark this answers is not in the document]
elsander: In that case, I would call that out explicitly in the documentation, because it won't necessarily be intuitive to a user who's used to redshift.

civis/tests/test_io.py
elsander: Ah, got it, this is where we were relying on the cassettes. That makes sense. Where it's possible to test on live objects like dataframes and temp directories, I'd favor that.

civis/io/_tables.py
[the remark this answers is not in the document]
zacharydfriedlander: Both of those parameters have a note in their section that says exactly that. It is also listed in our [API Documentation for the `imports/files/csv` endpoint](https://platform.civisanalytics.com/spa/#/api#post_imports_files_csv) and [Our Helpdesk docs for new imports under "Upsert Mode"](https://civis.zendesk.com/hc/en-us/articles/360027016851-Advanced-CSV-Imports-via-the-Civis-API). I can add a note to the `existing_table_rows` documentation as well, or add some RST emphasis to draw extra attention to it - let me know what you think.

civis/io/_tables.py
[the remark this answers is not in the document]
elsander: Good to know-- if you can mention that in the docs that would be great (maybe you already did, but the review comments aren't loading when I look at the code, so I can't tell right now)

civis/io/_tables.py
```diff
@@ -861,8 +905,26 @@ def civis_file_to_table(file_id, database, table, client=None,
...
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
+    primary_keys: list[str], optional
+        A list of the primary key column(s) of the destination table. If
+        existing_table_rows is "upsert", this field is required.
+    last_modified_keys: list[str], optional
+        A list of the columns indicating a record has been updated. If
+        existing_table_rows is "upsert", this field is required.
+    escaped: bool, optional
+        A boolean value indicating whether or not the source file has quotes
```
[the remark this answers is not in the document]
zacharydfriedlander: Good catch - fixed.

civis/io/_tables.py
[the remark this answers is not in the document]
zacharydfriedlander: Added!

civis/io/_tables.py
```diff
@@ -1094,3 +1168,122 @@ def split_schema_tablename(table):
...
+    for i, col in enumerate(column_list):
+        # Avoid mutating input arguments
+        new_col = dict(col)
+        if new_col.get('name') is None:
+            new_col['name'] = 'column_{}'.format(i)
+        new_cols.append(new_col)
+    return new_cols
+
+
+def _run_cleaning(file_ids, client, need_table_columns, hidden):
```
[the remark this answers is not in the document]
zacharydfriedlander: So this file cleaning was already happening under the hood of `AutoImport`s; Platform just hid it from users. This cleaning is necessary to achieve parity with what these functions used to do.

As far as slowdowns, the overall time to import a file should be about the same. However, users may see the time spent before the Python functions return increase, as the cleaning time has been shifted from being part of the import to being part of the calling function. 

Happy to discuss more about this offline if this explanation doesn't make sense.

civis/io/_tables.py
zacharydfriedlander: It looks like other parameters default to strings in this and other functions in this module, which is why I had opted for this approach. I'm happy to make the change for this parameter; however, a broader refactor is outside the scope of this PR.

As an aside, what I think we really ought to be doing here is leveraging [the `enum` module](https://docs.python.org/3/library/enum.html#functional-api) to create named constants for these values, rather than magic strings. However, that is decidedly a larger refactor and would (I believe) require dropping Python 2 support.

leeacto: sure I'm on board w/ this reasoning

civis/tests/test_io.py
zacharydfriedlander: Refactored - let me know if this helps with readability?

civis/io/_tables.py
zacharydfriedlander: John had suggested a refactor to this end in our discussion, but I pushed back then as the repetition didn't bother me as much. However, now that both you and he have raised it, I will move forward with this refactor.

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: For broader context, the new CSV import API endpoint does not perform file preprocessing. That function was decoupled into its own API endpoint. IMHO that is good for two reasons: 1) if you don't need to preprocess a file (which is true is many cases), you can skip that step (that step accounts for a majority of the execution time) 2) it allows us to add functionality that would not have been feasible without creating some truly horrendous code (e.g., importing files from S3).

IMHO the key point for consideration here is that we now block on file cleaning before returning a future.

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: Two additional points here and for full disclosure I wrote the original code so I was shaking my head at myself:

1. The existing code is written based on implementation details in the Console Ruby code to attempt to optimize the performance. That is bad. In the future we should consider removing this optimization all together.

2. There was a lot of unnecessary complexity injected because of the use of the `include_header` parameter even though a user cannot pass that parameter to `civis_file_to_table`. IMHO the new implementation is much more succinct.

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: Open to suggestions here but the challenge appears to be can we educate users on what an "upsert" is in a docstring? The primary keys are relevant regardless of the destination database because they are required to generate the SQL for an upsert operation.

One change we could consider to this description is: "A list of the primary key column(s) in the destination table _**that uniquely identify a record**_."

zacharydfriedlander:
I think this is ready for another round of review!

civis/io/_tables.py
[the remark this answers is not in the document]
elsander: Building off of that, I might say "A list of the primary key column(s) of the destination table that uniquely identify a record. If existing_table_rows is "upsert", this field is required. Note that this is true regardless of whether the destination database itself requires a primary key." This might seem like overkill, but because "primary key" is somewhat overloaded, I think it's useful to be as explicit as possible.

civis/io/_tables.py
```diff
@@ -819,24 +858,32 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+    """Upload the contents of one or more Civis files to a Civis table.
+       All provided files will be loaded as an atomic unit in parallel, and
+       should share the same columns in the same order, and be in the same
+       format.
 
     Parameters
     ----------
-    file_id : int
-        Civis file ID.
+    file_ids : int or list[int]
```
elsander: Same issue here, keep this `file_id`.

civis/io/_tables.py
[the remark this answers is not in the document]
elsander: That's helpful, thanks.

civis/io/_tables.py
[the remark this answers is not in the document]
elsander: Thanks for the additional context. I wasn't aware of the work that was happening on the Platform side, so I assumed this was additional checking being added. I agree, this seems like a good change.

civis/tests/test_io.py
```diff
@@ -144,42 +149,396 @@ def test_large_file_to_civis(self, *mocks):
...
 
+    @pytest.mark.civis_to_file
     @mock.patch(api_import_str, return_value=civis_api_spec)
     def test_civis_to_file(self, *mocks):
         buf = BytesIO()
         civis.io.civis_to_file(self.file_id, buf)
         buf.seek(0)
         assert buf.read() == b'a,b,c\n1,2,3'
 
-    @mock.patch(api_import_str, return_value=civis_api_spec)
```
elsander: I know you were having issues with the cassettes and I totally support removing them. But is there any way to do a manual check that the old tests still pass with the new code? Since we're changing the code and the tests at the same time, I'd like to check that we're not breaking the API.

zacharydfriedlander: I'm not sure exactly what you mean by "that the old tests still pass" ? They would fail, and be right to do so. The API calls that the cassettes were recording would no longer be hit, and VCR would complain about unexpected calls being made. I am happy to run live tests against Platform with the new code if that would help put worries to rest?

elsander: Ah, you're completely right. I'm basically trying to think of a way to check that existing code that calls the convenience functions won't break, and that the user will get fundamentally the same behavior. But the old tests won't do that. If you could run a couple live tests with the new code, that would be helpful. Mostly because it's a fairly large/complex PR, so it's very possible for me to miss something.

CHANGELOG.md
```diff
@@ -5,13 +5,19 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+- Added method `get_storage_host_id` to the APIClient
```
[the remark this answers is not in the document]
zacharydfriedlander: Oops - Added!

civis/io/_tables.py
```diff
@@ -819,24 +858,32 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
...
+                                  primary_keys=primary_keys,
+                                  last_modified_keys=last_modified_keys,
+                                  escaped=escaped, execution=execution,
                                   polling_interval=polling_interval,
                                   hidden=hidden)
     return fut
 
 
-def civis_file_to_table(file_id, database, table, client=None,
+def civis_file_to_table(file_ids, database, table, client=None,
```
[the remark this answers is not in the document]
zacharydfriedlander: I would like to officially register my displeasure that `file_id` can, in fact, be a list of file ids. However, not breaking the API is a bigger concern than any semantic quibbles. Changed back.

civis/io/_tables.py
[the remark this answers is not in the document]
zacharydfriedlander: Added more description to primary keys where relevant.

civis/tests/test_io.py
zacharydfriedlander: In reviewing this code, I actually realized that we _had_ introduced a new issue: in changing from directly constructing the`CivisFuture` to using `run_job`, `polling_interval` seemed to get incorrectly passed. This was because `run_job` didn't take a `polling_interval` argument. I have added an optional one, and used in the appropriate methods.

Other than that, I have created a [platform notebook](https://platform.civisanalytics.com/spa/#/notebooks/13325) to run the convenience methods. I replicated this notebook locally using the new implementation, (I would have loved to do so in platform, but ran into some issues with installing `civis` from my fork in a notebook) and verified that the results are the same as they were under the old implementation.

Please let me know if there are other tests besides these that you would like to see run!

zacharydfriedlander:
In manual testing, we discovered an issue with SQL types of different precisions being incorrectly rejected, e.g. `VARCHAR(42)` and `VARCHAR(32)` are rejected even though both are VARCHARs. The fix will be two-fold:

1. Add an option to loosen types as we do in AutoImports to CSVImports - that will require Platform side changes.
2. Loosen the strictness with which types are checked client-side. Rather than comparing types for string equality, we can look at just the base type.

Just wanted to give everyone a heads up here - I have [opened a PR in my fork](https://github.com/zacharydfriedlander/civis-python/pull/6) to address point 2, and will update as work progresses.

zacharydfriedlander:
After a number of other things arose, I've finally been able to get back to this - the necessary Platform side changes have been made, and I have updated `civis_file_to_table` to leverage the new parameter added.

You can see a sample import produced by the new code [here](https://api.civisanalytics.com/imports/files/csv/46859764).

civis/io/_tables.py
```diff
@@ -1335,3 +1336,58 @@ def _process_cleaning_results(cleaning_futures, client, headers,
...
+                               'has {} columns'.format(
+                                   len(table_columns),
+                                   output_obj_id,
+                                   len(file_columns))
+                               )
+
+    error_msgs = []
+    for idx, (tcol, fcol) in enumerate(zip(table_columns, file_columns)):
+        # for the purposes of type checking, we care only that the types
+        # share a base type (e.g. INT, VARCHAR, DECIMAl) rather than that
```
elsander: It looks like this check will cause an error for different integer types. Would you want this to fail if one is `INTEGER` and one is `BIGINT`? Also, is there any possibility you'd get different aliases passed in here? e.g. one is `INT` and one is `INT4`, and it throws an error?

elsander: Something else that occurred to me... is this interoperable with Postgres?

zacharydfriedlander: 1. We wanted to keep our scope as limited here as possible, which means that we are not processing types behind those with explicit lengths like `VARCHAR`. We are accepting that the type comparisons are not going to be perfect as the price we pay for avoiding rewriting the database's type inheritance system. The easiest way to get around the issue of type inference is to pre-create a table with the types that a user knows are correct. 

As far as your comment about aliases like `INT` vs `INT4`, the underlying system that performs the inference Platform-side will only return a single type from a given alias family, so that is not a concern.

2. This should be interoperable with Postgres - the preprocessing job is unaware of any types specific to a database. The types returned here are generic across the databases we support.

PR #333, opened 2019-12-18
DEP deprecation warning in readme and on import

### LINKED ISSUE #332: Add deprecation warning for Python 2
We should add a deprecation warning for Python 2 and/or update the `README.md`.

@elsander

elsander:
This PR adds a note in the README that Python 2 will no longer be supported as of April 1, 2020. The Python client will also issue a warning to Python 2 users when imported. I've attached an image below of what this looks like. One wrinkle here is that interpreters ignore deprecation warnings by default, so this doesn't show up unless the user disables that default. I don't know that there's a good way around this problem.

I initially tried writing a unit test for this, but I couldn't get it to pass, I think because pytest imports the library as it's setting up the tests. The warning did show up in the pytest warning summary, but was not captured in the unit test. I confirmed manually that the warning issues for Python 2 and does not issue for Python 3.

Closes #332 .

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   double quotes. (#328)
 
 ### Changed
-- Issue a `DeprecationWarning` on import for Python 2 users. (#333)
+- Issue a `FutureWarning` on import for Python 2 users. (#333)
 - Update the Civis logo in the Sphinx documentation. (#330)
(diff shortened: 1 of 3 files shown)


stephen-hoover:
Other projects in the scientific Python ecosystem have used `FutureWarning` to make their deprecation warnings visible, and we are currently doing that for deprecated parameters in this library: https://github.com/civisanalytics/civis-python/blob/v1.11.0/civis/_deprecation.py#L86. You might consider a `FutureWarning` instead of a `DeprecationWarning`.

elsander:
Great idea! I'll fix that.

PR #334, opened 2019-12-23
ENH Pass headers and delimiter to post_preprocess_csv

jkulzick:
In `civis.io.civis_file_to_table`, pass the `headers` and `delimiter` parameters to the call to `post_process_csv`. This allows the file cleaning process to use this information when users provide it. Since header row detection in CSVs can be difficult depending on the columns and data types represented in the file its good to give users the ability to specify these values.

I performed two manual tests of `civis_file_to_table` from my local branch:
1. One without headers/delimiters
`civis.io.civis_file_to_table(74975425, 'redshift-general', 'jkulzick.py_hdrs', existing_table_rows='append')`
2. One with headers/delimiters
`civis.io.civis_file_to_table(74975425, 'redshift-general', 'jkulzick.py_hdrs', headers=True, delimiter=',', existing_table_rows='append')`


jkulzick:
@zacharydfriedlander fyi

PR #337, opened 2019-12-31
ENH more informative error for expired files

### LINKED ISSUE #336: Better error handling for expired files in `civis.io.file_to_dataframe`
When reading from an expired file using `civis.io.file_to_dataframe`, the user gets a long traceback and the following error message: `ValueError: Invalid file path or buffer object type: <class 'NoneType'>`. We should catch this error and return something more informative.

elsander:
Raise a more informative exception when calling `civis.io.file_to_dataframe` on a file with no url (which likely means the file is expired). I went with a `ValueError` here but could change it if a different exception would be easier to understand. This error check is pre-empting a `ValueError`, since the error comes from reading a csv from `None`.

Closes #336.

### DIFF
--- a/civis/io/_files.py
+++ b/civis/io/_files.py
@@ -470,6 +470,10 @@ def file_to_dataframe(file_id, compression='infer', client=None,
     client = APIClient() if client is None else client
     file_info = client.files.get(file_id)
     file_url = file_info.file_url
+    if not file_url:
+        raise ValueError(
+            "File url does not exist for file {}. File may be expired.".format(
+                file_id))
     file_name = file_info.name
     if compression == 'infer':
         comp_exts = {'.gz': 'gzip', '.xz': 'xz', '.bz2': 'bz2', '.zip': 'zip'}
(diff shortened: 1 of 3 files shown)


civis/io/_files.py
```diff
@@ -470,6 +470,10 @@ def file_to_dataframe(file_id, compression='infer', client=None,
     client = APIClient() if client is None else client
     file_info = client.files.get(file_id)
     file_url = file_info.file_url
+    if not file_url:
+        raise ValueError(
+            "File url does not exist for file {}. File may be expired.".format(
+                file_id))
```
arthir: Do we want to make this a general check for the other file functions as well (like `file_to_json`)? We're not running into issues with it right now but it might make sense to have them throw consistent errors

elsander: Good point! We definitely should.

elsander: As it turns out, the other file functions do have a check like this, and `file_to_dataframe` was the exception because it doesn't call `civis_to_file`. I updated the exception to match the pre-existing one in the other functions.

arthir: Ah got it.

PR #340, opened 2019-12-31
339 3.4 deprecation

### LINKED ISSUE #339: Deprecation warning for Python 3.4
Python 3.4 reached EOL in March 2019. We should deprecate API client support for 3.4 at the same time that we do for Python 2. We can add a `FutureWarning` for the next release, then officially remove testing and version-specific support in a future release.

elsander:
Adds a `FutureWarning` that Python 3.4 will be deprecated along with Python 2. Closes #339. I considered adding 3.5 to the list, since it's reaching EOL in 2020, but I figured it's better to be conservative about dropping support.

I don't know of a way to test that this code works within pytest (see discussion in #333 ), but I've attached screenshots confirming that the warning is issued in 3.4 and not in 3.6.
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   double quotes. (#328)
 
 ### Changed
-- Issue a `FutureWarning` on import for Python 2 users. (#333)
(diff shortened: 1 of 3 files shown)


elsander:
@patr1ckm flagging this in case it got lost in the shuffle.

PR #341, opened 2020-01-06
[PARO-685] [PARO-744] Decouple CivisML from civis-python: Use template ID aliases and move docs to Zendesk

jacksonllee:
This PR decouples CivisML from civis-python with respect to the following:
* PARO-685: Use template ID aliases, so that a new CivisML release doesn't require a new civis-python release anymore.
* PARO-744: Document CivisML dependency versions on Zendesk rather than in civis-python's Sphinx docs, so that we can update Zendesk independently when CivisML is updated for the modeling dependencies (civis-python's Sphinx docs don't update unless a new civis-python release is made).

@stephen-hoover You are the appointed reviewer, since I think you're the person who has thought about the use of aliases for this scenario the most.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -22,7 +22,7 @@ try:
 except ImportError:
     HAS_SKLEARN = False
 
-from civis import APIClient, find, find_one, __version__
+from civis import APIClient, find, find_one
 from civis._utils import camel_to_snake
 from civis.base import CivisAPIError, CivisJobFailure
 from civis.compat import FileNotFoundError, TemporaryDirectory
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -25,7 +25,7 @@ except ImportError:
 from civis import APIClient, find, find_one
 from civis._utils import camel_to_snake
 from civis.base import CivisAPIError, CivisJobFailure
-from civis.compat import FileNotFoundError, TemporaryDirectory
+from civis.compat import FileNotFoundError, TemporaryDirectory, lru_cache
 import civis.io as cio
 from civis.futures import ContainerFuture
 from civis.response import Response
@@ -36,8 +36,6 @@ log = logging.getLogger(__name__)
 
 # sentinel value for default primary key value
 SENTINEL = collections.namedtuple('Sentinel', [])()
-
-_TEMPLATE_IDS = None  # To be updated by _get_template_ids_all_versions
 
 
 class ModelError(RuntimeError):
@@ -297,6 +295,7 @@ def _get_job_type_version(alias):
     return job_type, version
 
 
+@lru_cache()
 def _get_template_ids_all_versions(client):
     """Get templates IDs for all accessible CivisML versions.
 
@@ -351,11 +350,9 @@ def _get_template_ids(civisml_version, client):
     int
         Template ID for pre-trained model registration
     """
-    global _TEMPLATE_IDS
-    if _TEMPLATE_IDS is None:
-        _TEMPLATE_IDS = _get_template_ids_all_versions(client)
+    template_ids_all_versions = _get_template_ids_all_versions(client)
     try:
-        ids = _TEMPLATE_IDS[civisml_version]
+        ids = template_ids_all_versions[civisml_version]
     except KeyError:
         msg = (
             '"{civisml_version}" is an invalid CivisML version. '
@@ -368,7 +365,9 @@ def _get_template_ids(civisml_version, client):
             accessible_versions=', '.join(
                 '"%s"' % v
                 # Don't include None, or else it would crash sorted()
-                for v in sorted(v for v in _TEMPLATE_IDS.keys() if v)
+                for v in sorted(
+                    v for v in template_ids_all_versions.keys() if v
+                )
             )
         )
         raise ValueError(msg)
@@ -1073,13 +1072,9 @@ class ModelPipeline:
         >>> model.train_result_.metrics['roc_auc']
         0.843
         """
-        global _TEMPLATE_IDS
-
         train_job_id = int(train_job_id)  # Convert np.int to int
         if client is None:
             client = APIClient()
-        if _TEMPLATE_IDS is None:
-            _TEMPLATE_IDS = _get_template_ids_all_versions(client)
         train_run_id = _decode_train_run(train_job_id, train_run_id, client)
         try:
             fut = ModelFuture(train_job_id, train_run_id, client=client)
@@ -1141,7 +1136,7 @@ class ModelPipeline:
         # Set prediction template corresponding to training template
         template_id = int(container['from_template_id'])
         ids = find_one(
-            _TEMPLATE_IDS.values(),
+            _get_template_ids_all_versions(client).values(),
             lambda ids: ids['training'] == template_id
         )
         p_id = ids['prediction']
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -289,7 +289,8 @@ def _get_job_type_version(alias):
         job_type = match_special.group(1)
         version = match_special.group(2)
     else:
-        msg = '"%r" does not look like a CivisML alias'
+        msg = ('Unable to parse the job type and version '
+               'from the CivisML alias "%r"')
         raise ValueError(msg % alias)
 
     return job_type, version
@@ -321,7 +322,15 @@ def _get_template_ids_all_versions(client):
         lambda: {'training': None, 'prediction': None, 'registration': None}
     )
     for alias_obj in civisml_template_alias_objects:
-        job_type, version = _get_job_type_version(alias_obj.alias)
+        try:
+            job_type, version = _get_job_type_version(alias_obj.alias)
+        except ValueError:
+            msg = (
+                '%r looks like a CivisML alias for the prefix "civis-civisml-"'
+                ', but it is impossible to parse its job type and version'
+            )
+            log.debug(msg % alias_obj)
+            continue
         ids[version][job_type] = alias_obj.object_id
     if not ids:
         r = Response({'status_code': 404,
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -308,11 +308,20 @@ def _get_template_ids(civisml_version, client):
         ids = _TEMPLATE_IDS[civisml_version]
     except KeyError:
         msg = (
-            '%r is an invalid CivisML version. Valid versions are in the '
-            'form of "v2.3", "v2.2", etc., '
-            'or simply None for the latest production version.'
+            '"{civisml_version}" is an invalid CivisML version. '
+            'Either this version does not exist, or you do not have access '
+            'to this version. '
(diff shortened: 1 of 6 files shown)


civis/ml/_model.py
```diff
@@ -21,32 +22,22 @@
...
-                   9112: 9113,    # v1.1
-                   8387: 8388,    # v1.0
-                   7020: 7021,    # v0.5
-                   }
-_CIVISML_TEMPLATE = None  # CivisML training template to use
-REGISTRATION_TEMPLATES = [11221,  # v2.2
-                          ]
+SENTINEL = collections.namedtuple('Sentinel', [])()
+
+_TEMPLATE_IDS = None  # To be updated by _get_template_ids_all_versions
```
jacksonllee: I don't want to trigger API calls when the module is loaded, and hence this approach. The drawback is that the following pattern is found here and there:

```python
    global _TEMPLATE_IDS
    if _TEMPLATE_IDS is None:
        _TEMPLATE_IDS = _get_template_ids_all_versions(client)
```

"Drawback" might not be the right word, because it's highly readable and maintainable to me? Feel free to let me know if there's a better way to populate `_TEMPLATE_IDS`.

civis/ml/_model.py
```diff
@@ -268,6 +259,122 @@ def _show_civisml_warnings(warn_list):
...
+    """
+    template_alias_objects = client.aliases.list(
+        object_type='template_script', iterator=True
+    )
+    civisml_template_alias_objects = find(
+        template_alias_objects,
+        alias=lambda alias: alias.startswith('civis-civisml-')
+    )
+    ids = collections.defaultdict(
+        lambda: {'training': None, 'prediction': None, 'registration': None}
```
jacksonllee: I considered some sort of `collections.namedtuple` to represent `{'training': None, 'prediction': None, 'registration': None}`, but decided that the dictionary syntax is clearer for readable and maintainability.

civis/ml/_model.py
```diff
@@ -1042,17 +1140,11 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
...
-        if p_id is None:
-            warnings.warn('Model %s was trained with a newer version of '
-                          'CivisML than is available in the API client '
-                          'version %s. Please update your API client version. '
-                          'Attempting to use an older version of the '
-                          'prediction code. Prediction will either fail '
-                          'immediately or succeed.'
-                          % (train_job_id, __version__), RuntimeWarning)
-            p_id = max([v for k, v in _PRED_TEMPLATES.items()
-                        if k not in REGISTRATION_TEMPLATES])
```
jacksonllee: Removing this warning since this situation should be impossible to arise, given that we get the registration template ID that corresponds to the given training template ID.

civis/ml/_model.py
```diff
@@ -801,16 +885,14 @@ def __getstate__(self):
     def __setstate__(self, state):
         self.__dict__ = state
         self._client = APIClient()
-        template_ids = self._get_template_ids(self._client)
-        self.train_template_id, self.predict_template_id = template_ids
```
jacksonllee: Before this PR: The behavior was that the code always tried to get the latest template IDs, which might be different from before the pickled `ModelPipeline` was created (due to a civis-python/CivisML new version).

After this PR: When a `ModelPipeline` instance is created, the training template ID is already set, possibly by the user-provided `civisml_version` parameter. This means the unpickling a `ModelPipeline` object shouldn't attempt to update the training and prediction template IDs anymore. Hence deleting these two lines. I've added a test to make sure that template IDs are preserved before vs after pickling.

jacksonllee:
There's a couple very minor things for test failures that I'm going to fix, but this PR is still ready for review as-is.

jacksonllee:
I've fixed the test failures related to this PR. The overall status for TravisCI is still red, because of all the following facts together:
* _Two hours ago_, PyYAML released v5.3 ([which bans Python 3.4](https://github.com/yaml/pyyaml/blob/5.3/setup.py#L314)).
* civis-python [allows PyYAML v5.3](https://github.com/civisanalytics/civis-python/blob/d3d695330ffd4cdb771edea5d02cc7a02f10d242/setup.py#L54).
* [civis-python still runs tests on Python 3.4](https://github.com/civisanalytics/civis-python/blob/d3d695330ffd4cdb771edea5d02cc7a02f10d242/.travis.yml#L5).

Only the build for Python 3.4 failed, and I'm going to ignore this failure in this PR. The solution would be to either restrict civis-python to use PyYAML < 5.3 (which doesn't seem desirable, since we're deprecating Python 3.4 support and eventually removing testing later anyway, cf. #339), or remove Python 3.4 from the Travis CI build yaml config when the civis-python maintainers officially do so.

stephen-hoover:
It's really exciting to see this finally happening! I will get to code review sometime tomorrow. On the test failure, it looks like @uttercm is addressing the pyyaml dependency as part of #342 . That will need to merge before this can (or at the very least, we'll need the dependency updates from there).

uttercm:
@jacksonllee my PR was merged so if you merge master you should be good to go

jacksonllee:
Thanks for the help, @uttercm! All builds are green now.

docs/source/ml.rst
```diff
@@ -130,13 +130,8 @@ the pre-defined ones. Create the object and pass it as the ``model`` parameter
...
 :ref:`custom-dependencies` if they are not already installed in
-CivisML. Preinstalled libraries available for your use include:
-
-- `scikit-learn <http://scikit-learn.org>`_ v0.19.1
-- `glmnet <https://github.com/civisanalytics/python-glmnet>`_ v2.0.0
-- `xgboost <http://xgboost.readthedocs.io>`_ v0.6a2
-- `muffnn <https://github.com/civisanalytics/muffnn>`_ v1.2.0
-- `civisml-extensions <https://github.com/civisanalytics/civisml-extensions>`_ v.0.1.6
+CivisML. Please check `here <https://civis.zendesk.com/hc/en-us/articles/360000260011-CivisML>`_
+for the available pre-installed libraries and their versions.
```
stephen-hoover: Comment: This does have the downside of requiring that people have Platform access before they see what libraries we have available. It seems worth it to avoid the dependency between CivisML updates and API client updates, though.

docs/source/ml.rst
```diff
@@ -414,3 +423,9 @@ Object reference
 .. autoclass:: civis.ml.ModelFuture
 	       :members:
 	       :inherited-members:
+
+.. autofunction:: civis.ml.put_models_shares_users
+.. autofunction:: civis.ml.put_models_shares_groups
+.. autofunction:: civis.ml.delete_models_shares_users
+.. autofunction:: civis.ml.delete_models_shares_groups
+.. autofunction:: civis.ml.list_models
```
stephen-hoover: Thank you for adding these in! Question -- would it work to put something like `.. automodule:: civis.ml._helper`? If so, that would mean we wouldn't forget to document new functions in that module. The displayed path might be wrong that way, though. Maybe `.. automodule:: civis.ml`? I can't remember if we've tried that before.

jacksonllee: Yes, I did actually try versions of `.. automodule:: civis.ml._helper`, but no luck. I am 99% sure we had the exact same conversation in the past when we added functions under `civis.io` from time to time but kept forgetting to add the new ones to Sphinx docs. I'm okay with just trying to be more mindful of documentation. :-)

civis/ml/_model.py
```diff
@@ -268,6 +257,132 @@ def _show_civisml_warnings(warn_list):
...
+    elif match_v:
+        job_type = match_v.group(1)
+        version = 'v%s.%s' % match_v.group(2, 3)
+    elif match_special:
+        job_type = match_special.group(1)
+        version = match_special.group(2)
+    else:
+        msg = ('Unable to parse the job type and version '
+               'from the CivisML alias "%r"')
+        raise ValueError(msg % alias)
```
jacksonllee: Previously this ValueError message said "it doesn't look like a CivisML alias". This message was misleading, because the alias wouldn't end up being here if it didn't start with `"civis-civisml-"` from the upstream code.

civis/ml/_model.py
```diff
@@ -268,6 +257,132 @@ def _show_civisml_warnings(warn_list):
...
+    for alias_obj in civisml_template_alias_objects:
+        try:
+            job_type, version = _get_job_type_version(alias_obj.alias)
+        except ValueError:
+            msg = (
+                '%r looks like a CivisML alias for the prefix "civis-civisml-"'
+                ', but it is impossible to parse its job type and version'
+            )
+            log.debug(msg % alias_obj)
+            continue
```
jacksonllee: I've chosen to `log.debug` this if this unlikely situation arises. A warning would be visible to the user and it'd be at least confusing, if not incomprehensible, to them.

civis/ml/tests/test_model.py
```diff
@@ -666,50 +690,91 @@ def test_metrics_prediction(mock_file_id_from_run_output):
...
+def test__get_template_ids_all_versions():
+    m_client = create_client_mock()
+    m_client.aliases.list.return_value = TEST_TEMPLATE_ID_ALIAS_OBJECTS
+    actual_template_ids = _model._get_template_ids_all_versions(m_client)
+    expected_template_ids = TEST_TEMPLATE_IDS
+    assert actual_template_ids == expected_template_ids
+
+
+@mock.patch.object(_model, '_get_template_ids_all_versions',
+                   mock.Mock(return_value=TEST_TEMPLATE_IDS))
```
stephen-hoover: Comment: You'll get the same effect by supplying `return_value` directly. Extra keyword arguments get passed directly to the `Mock` object's `__init__` (which can cause subtle bugs if you misspell a keyword parameter in your patch).
```suggestion
                   return_value=TEST_TEMPLATE_IDS)
```

civis/ml/_model.py
```diff
@@ -864,6 +946,9 @@ def register_pretrained_model(cls, model, dependent_variable=None,
         client : :class:`~civis.APIClient`, optional
             If not provided, an :class:`~civis.APIClient` object will be
             created from the :envvar:`CIVIS_API_KEY`.
+        civisml_version : str, optional
```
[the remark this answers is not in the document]
stephen-hoover: Okay, that makes sense. And as I think about it, I can see situations where someone might want to specify the version number -- for example, if you were putting a training step in a production pipeline, you would probably want to pin the CivisML version.

docs/source/ml.rst
stephen-hoover: Okay, thanks. I thought we must have tried automodule in the past, but I thought I'd double-check.

civis/ml/tests/test_model.py
jacksonllee: I had tried the kwarg `return_value=TEST_TEMPLATE_IDS` but the tests didn't pass, which is why I'm using the supposedly equivalent `mock.Mock(return_value=TEST_TEMPLATE_IDS)`. I was puzzled why `return_value=TEST_TEMPLATE_IDS` didn't work, but wasn't interested enough to spend more time on this.

PR #343, opened 2020-01-08
[Serd-1694] service client

ggarcia-civis:
Adding ServiceClient() method which takes a service id, optional base path (default is None), and optional swagger endpoint (default is '/endpoints'), then creates a client from the api spec and endpoints.

A lot of the work is being done in the helper functions and Endpoint class for the original APIClient, but small changes were needed in a few cases (mostly in regards to the url and making requests). These changes are overridden by the ServiceClient and ServiceEndpoint class.

### DIFF
--- a/civis/service_client.py
+++ b/civis/service_client.py
@@ -6,10 +6,10 @@ from jsonref import JsonRef
 import requests
 import six
 
-from civis import APIClient
 from civis.base import Endpoint, CivisAPIError, tostr_urljoin
 from civis.resources._resources import parse_method
 from civis._utils import to_camelcase
+from civis import APIClient
 
 
 def auth_service_session(session, service_id):
--- a/civis/service_client.py
+++ b/civis/service_client.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 from collections import OrderedDict
+import json
 import re
 
 from jsonref import JsonRef
@@ -100,12 +101,13 @@ class ServiceClient():
             list-like responses and a :class:`pandas:pandas.Series` for
             single a json response.
         local_api_spec : collections.OrderedDict or string, optional
-            The methods on this class are dynamically built from the Service API
-            specification, which can be retrieved from the /endpoints endpoint.
-            When local_api_spec is None, the default, this specification is
-            downloaded the first time APIClient is instantiated. Alternatively,
-            a local cache of the specification may be passed as either an
-            OrderedDict or a filename which points to a json file.
+            The methods on this class are dynamically built from the Service
+            API specification, which can be retrieved from the /endpoints
+            endpoint. When local_api_spec is None, the default, this
+            specification is downloaded the first time APIClient is
+            instantiated. Alternatively, a local cache of the specification
+            may be passed as either an OrderedDict or a filename which
+            points to a json file.
         """
         if return_type not in ['snake', 'raw', 'pandas']:
             raise ValueError("Return type must be one of 'snake', 'raw', "
--- a/civis/service_client.py
+++ b/civis/service_client.py
@@ -134,6 +134,6 @@ class ServiceClient():
             service = client.services.get(self._service_id)
         except CivisAPIError as err:
             msg = ('There was an issue '
-                   'finding service with ID {}.').format(service_id)
+                   'finding service with ID {}.').format(self._service_id)
             six.raise_from(ValueError(msg), err)
         return service['current_url']
--- a/civis/service_client.py
+++ b/civis/service_client.py
@@ -31,19 +31,18 @@ def auth_service_session(session, service_id):
 class ServiceEndpoint(Endpoint):
 
     def __init__(self, session_kwargs, client,
-                 return_type='civis', root_path=None):
+                 return_type='civis'):
         self._session_kwargs = session_kwargs
         self._return_type = return_type
         self._client = client
-        self._root_path = root_path
 
     def _build_path(self, path):
         if not path:
             return self._client._base_url
-        if not self._root_path:
+        if not self._client._root_path:
             return tostr_urljoin(self._client._base_url, path.strip("/"))
         return tostr_urljoin(self._client._base_url,
-                             self._root_path.strip("/"),
+                             self._client._root_path.strip("/"),
                              path.strip("/"))
 
     def _make_request(self, method, path=None, params=None, data=None,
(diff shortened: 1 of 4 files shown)


ggarcia-civis:
I have a few fixes to make, one moment

ggarcia-civis:
Okay, ready for review when you all are @patr1ckm @elsander

elsander:
Also, don't forget to update the changelog!

civis/service_client.py
```diff
@@ -0,0 +1,140 @@
...
+    # Make request for adding Authentication Cookie to session
+    session.get(auth_url)
+
+
+class ServiceEndpoint(Endpoint):
+
+    def __init__(self, session_kwargs, client,
+                 return_type='civis'):
+        self._session_kwargs = session_kwargs
+        self._return_type = return_type
```
patr1ckm: Is this used?

civis/service_client.py
```diff
@@ -0,0 +1,140 @@
...
+        return tostr_urljoin(self._client._base_url,
+                             self._client._root_path.strip("/"),
+                             path.strip("/"))
+
+    def _make_request(self, method, path=None, params=None, data=None,
+                      **kwargs):
+        url = self._build_path(path)
+
+        with requests.Session() as sess:
+            auth_service_session(sess, self._client._service_id)
```
patr1ckm: Does this mean that every request from the generated client needs to be re-authorized? Or does `requests` automatically cache for a session?

civis/service_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def __init__(self, service_id, root_path=None, swagger_path="/endpoints"):
+        return_type = 'snake'
+        self._session_kwargs = {}
+        self._service_id = service_id
+        self._base_url = self.get_base_url()
+        self._root_path = root_path
+        self._swagger_path = swagger_path
+        # Catch deprecation warnings from generate_classes_maybe_cached and
+        # the functions it calls until the `resources` argument is removed.
```
patr1ckm: I don't understand this, can you say more?

civis/service_client.py
```diff
@@ -0,0 +1,140 @@
...
+                                        params=params, **kwargs)
+
+        if not response.ok:
+            six.raise_from(ValueError(response.text),
+                           ValueError)
+
+        return response
+
+
+class ServiceClient():
```
patr1ckm: Somewhere in the docs we should mention that the endpoint must be swagger compatible, and briefly mention the additional flexibility from the spec that we provide.

civis/service_client.py
ggarcia-civis: no, unfortunately it has to be reauthorized. when you retrieve a service from the api client, it has a url that will generate a session cookie, but the cookie expires after a few minutes. So for every request we are refreshing that url and hitting it first before hitting the actual request endpoint.

civis/service_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def auth_service_session(session, service_id):
+    try:
+        client = APIClient()
+        service = client.services.get(service_id)
+    except CivisAPIError as err:
+        msg = ('There was an issue '
+               'finding service with ID {}.').format(service_id)
+        six.raise_from(ValueError(msg), err)
```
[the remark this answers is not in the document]
ggarcia-civis: that's fair, the APIClient can provide a better error description

civis/service_client.py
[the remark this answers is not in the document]
ggarcia-civis: sure, I'll look into it. I wasn't aware of the help() method.

civis/service_client.py
ggarcia-civis: This was copied from APIClient(), generate_classes_maybe_cached() is a function with from depreciated parameter so this was here to catch warnings. But we are not using that function so I can take it out.
At the time I wasn't sure if I should keep generate_classes_maybe_cached() or not. It allows a user to pass in an api spec as a file or dict, instead of downloading one from an endpoint.
Do you have strong feelings about using it or not?

patr1ckm: I do not

elsander: Seems like [YAGNI](https://martinfowler.com/bliki/Yagni.html) to me. I'd vote to take it out until we hear requests for it. I don't feel strongly on that though.

stephen-hoover: I can comment on the motivation for allowing API client generation without having to hit a remote endpoint. It's been very useful for testing. Building an API client from a stored spec lets us test code that depends on the client without needing to hard-code structure (example: https://github.com/civisanalytics/civis-python/blob/v1.11.0/civis/tests/mocks.py#L30). It's also used for between-session caching in the CLI. Downloading the spec and regenerating endpoints can take a non-trivial amount of time. I don't know if those are relevant here. You could consider adding an `@lru_cache` decorator to `ServiceClient.generate_classes` to speed up ServiceClient instantiation within a session.

ggarcia-civis: Thanks for your Insight @stephen-hoover! I'll take a better look at our test cases and the decorator. It does take a bit of time to generate the client.

elsander: Ah, that makes sense. In that case, it seems reasonable to keep (although we don't have anything to test it on quite yet).

ggarcia-civis:
I think I covered all the previous comments. Let me know if I missed anything.

civis/service_client.py
```diff
@@ -2,41 +2,37 @@
...
-def _get_service(service_id):
-    try:
-        client = APIClient()
-        service = client.services.get(service_id)
-    except CivisAPIError as err:
-        msg = ('There was an issue '
-               'finding service with ID {}.').format(service_id)
-        six.raise_from(ValueError(msg), err)
+def _get_service(client):
+    if client._api_key:
```
patr1ckm: I don't think having the api key as an argument is a wise interface to have. When using `ServiceClient`, a novice user could e.g. accidentally commit their API key to a public repo.

civis/service_client.py
```diff
@@ -68,23 +64,52 @@ def _make_request(self, method, path=None, params=None, data=None,
 
 class ServiceClient():
 
-    def __init__(self, service_id, root_path=None, swagger_path="/endpoints"):
-        return_type = 'snake'
-        self._session_kwargs = {}
+    def __init__(self, service_id, root_path=None,
+                 swagger_path="/endpoints", api_key=None,
```
patr1ckm: I don't think we want to have an argument for the `api_key` here, accessing only via an environment variable is a more secure interface.

civis/service_client.py
ggarcia-civis: @patr1ckm, thats a good point. I brought this up to @elsander before adding it, and they thought we should include it so we could support different work styles like APIClient does. Your concern is valid though, but I see ServiceClient as being an internal tool, with novice users mainly using the clients generated from it instead, like we plan to do with ToolboxClient(). Thoughts?

patr1ckm: Ah, thanks for the additional clarity!

patr1ckm:
LGTM!

PR #347, opened 2020-01-13
[PARO-685] Smoke tests for different CivisML alias objects

jacksonllee:
This PR is a follow-up of #341, which makes strong assumptions about what CivisML alias objects exist on Civis Platform.

Before this PR, the smoke test for CivisML modeling used the default scenario of not specifying `civisml_version` for `civis.ml.ModelPipeline` (which would use the latest CivisML version in production). This PR makes the CivisML modeling smoke test also run on an explicitly specified `civisml_version` as well, which would access different CivisML alias objects than the unspecified `civisml_version` does.

While working on this PR, I uncovered a bug where `civis.ml.ModelPipeline`'s `civisml_version` param with a non-existent CivisML version did not raise the expected error of "the specified CivisML version doesn't exist". Bug fixed at 0eb8dd4.

### DIFF
--- a/tools/smoke_tests.py
+++ b/tools/smoke_tests.py
@@ -82,15 +82,18 @@ def main():
 
     # Test modeling.
     logger.info('Testing Civis-ML...')
-    mp = civis.ml.ModelPipeline(
-        model="sparse_logistic",
-        dependent_variable="type",
-        primary_key="index",
-        client=client
-    )
-    result = mp.train(
-        table_name="datascience.iris", database_name=database).result()
-    assert result['state'] == 'succeeded'
+    for civisml_version in (None, 'v2.2'):  # None = latest production version
+        logger.info('CivisML version: %r', civisml_version)
+        mp = civis.ml.ModelPipeline(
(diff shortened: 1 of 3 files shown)


tools/smoke_tests.py
```diff
@@ -82,15 +82,18 @@ def main():
...
-    mp = civis.ml.ModelPipeline(
-        model="sparse_logistic",
-        dependent_variable="type",
-        primary_key="index",
-        client=client
-    )
-    result = mp.train(
-        table_name="datascience.iris", database_name=database).result()
-    assert result['state'] == 'succeeded'
+    for civisml_version in (None, 'v2.2'):  # None = latest production version
```
elsander: Just to clarify, specifying `v.2.2` is so that you'll be testing both the production and versioned alias? This test won't need to be updated if a new template is released, right?

jacksonllee: > specifying v.2.2 is so that you'll be testing both the production and versioned alias?

Correct.

> This test won't need to be updated if a new template is released, right?

Correct.

PR #353, opened 2020-01-24
handle deprecated endpoints

### LINKED ISSUE #345: Add `ignore_deprecated` flag to `APIClient`
The API client currently ignores any endpoints with a deprecation warning. This means that endpoints effectively cannot be deprecated-- once they are tagged deprecated, they are inaccessible from the client.

One option would be to automatically pull in these endpoints, which would allow for deprecation, but would make deprecated endpoints visible to new users, which may not be desirable. 

The alternative I'd propose is to add an `ignore_deprecated` flag when instantiating the client, which defaults to `True`. This means that existing code based on newly deprecated endpoints would still break, but could be updated to work until the endpoint is removed entirely.

salilgupta1:
One thing to note:
- We'll have code that asserts that an endpoint must have both `deprecated` and `warning` set, not one or the other.

This PR will close #345.

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -421,7 +421,6 @@ def parse_method(verb, operation, path):
     summary = operation["summary"]
     params = operation.get("parameters", [])
     responses = operation["responses"]
-    deprecated = operation.get('deprecated', False)
     deprecation_warning = operation.get("x-deprecation-warning", None)
     if 'deprecated' in summary.lower():
         return None
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -132,6 +132,16 @@ def docs_from_properties(properties, level=0):
     return docs
 
 
+def deprecated_notice(deprecation_warning):
+    """ Return a doc string element for the deprecation notice. The
+    doc string can be an empty string if the warning is None
+    """
+    if deprecation_warning is None:
+        return ""
+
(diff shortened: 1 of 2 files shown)


elsander:
@stephen-hoover Wanted to flag this for you, since you might have more context on the original decision here.

@salilgupta1 , can you note in the description that this closes #345 ?

civis/resources/_resources.py
```diff
@@ -411,13 +421,14 @@ def parse_method(verb, operation, path):
     summary = operation["summary"]
     params = operation.get("parameters", [])
     responses = operation["responses"]
-    deprecated = operation.get('deprecated', False)
-    if 'deprecated' in summary.lower() or deprecated:
+    deprecation_warning = operation.get("x-deprecation-warning", None)
+    if 'deprecated' in summary.lower():
```
salilgupta1: I have to leave this in because there seem to be a handful of endpoints that use this paradigm. We can get rid of this once we remove those endpoints.

stephen-hoover:
> @stephen-hoover Wanted to flag this for you, since you might have more context on the original decision here.

Thanks for the flag, but I don't recall being involved in any discussions around the initial implementation of this part of the code.

civis/resources/_resources.py
elsander: Strange that we have endpoints which explicitly do not work but are marked as deprecated. This seems like a reasonable workaround in the meantime.

PR #359, opened 2020-02-10
ENH CLI live log tail

aripollak:
* Adds `civis jobs follow-log` and `civis jobs follow-run-log`
* Like `tail -f` for job run logs, closely mirroring the output of run logs on the web UI

<details><summary>Demo GIFs</summary>
<p>

demo1

demo2

</p>
</details>

### DIFF
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -5,8 +5,10 @@ Additional commands to add to the CLI beyond the OpenAPI spec.
 """
 from __future__ import print_function
 import functools
+import operator
 import os
 import sys
+import time
 
 import click
 import requests
@@ -32,6 +34,7 @@ _CIVIS_ASCII_ART = r"""
  \   \ .'  '---'         '---" '---'
   `---`
 """
+_LOG_POLL_INTERVAL_SEC = 3
 
 
 @click.command('upload')
@@ -154,6 +157,41 @@ def _str_table_result(cols, rows):
         tb_strs.append(" | ".join("{0:>{width}}".format(_v, width=_l)
                                   for _l, _v in zip(max_len, row)))
     return '\n'.join(tb_strs)
+
+
+@click.command('follow-runs-logs')
+@click.argument('id', type=int)
+@click.argument('run_id', type=int)
+def jobs_follow_run_logs(id, run_id):
+    """Output live run logs.
+
+    NOTE: This command could miss some log entries from a currently-running
+    job. It does not re-fetch logs that might have been saved out of order, to
+    preserve the chronological order of the logs and without duplication.
+    """
+    client = civis.APIClient(return_type='raw')
+    seen_max_log_id = None
+    continue_polling = True
+
+    while continue_polling:
+        response = client.jobs.list_runs_logs(id, run_id,
+                                              last_id=seen_max_log_id)
+        if 'civis-max-id' in response.headers:
+            remote_max_log_id = int(response.headers['civis-max-id'])
+        log_finished = response.headers['civis-cache-control'] != 'no-store'
+        logs = response.json()
+        if logs:
+            seen_max_log_id = max(log['id'] for log in logs)
+            logs.sort(key=operator.itemgetter('createdAt'))
+        for log in logs:
+            print(' '.join((log['createdAt'], log['message'])))
+        # if output is a pipe, write the buffered output immediately:
+        sys.stdout.flush()
+        if seen_max_log_id == remote_max_log_id:
+            if log_finished:
+                continue_polling = False
+            else:
+                time.sleep(_LOG_POLL_INTERVAL_SEC)
 
 
 @click.command('download')
--- a/civis/cli/_cli_commands.py
+++ b/civis/cli/_cli_commands.py
@@ -196,19 +196,19 @@ def jobs_follow_run_log(id, run_id):
 
 def _jobs_follow_run_log(id, run_id):
     client = civis.APIClient(return_type='raw')
-    seen_max_log_id = 0
+    local_max_log_id = 0
     continue_polling = True
 
     while continue_polling:
         response = client.jobs.list_runs_logs(id, run_id,
-                                              last_id=seen_max_log_id)
+                                              last_id=local_max_log_id)
         if 'civis-max-id' in response.headers:
             remote_max_log_id = int(response.headers['civis-max-id'])
         else:
             remote_max_log_id = None
         logs = response.json()
         if logs:
-            seen_max_log_id = max(log['id'] for log in logs)
+            local_max_log_id = max(log['id'] for log in logs)
             logs.sort(key=operator.itemgetter('createdAt', 'id'))
         for log in logs:
             print(' '.join((log['createdAt'], log['message'].rstrip())))
@@ -217,7 +217,7 @@ def _jobs_follow_run_log(id, run_id):
         log_finished = response.headers['civis-cache-control'] != 'no-store'
         if remote_max_log_id is None:
             time.sleep(_FOLLOW_POLL_INTERVAL_SEC)
-        elif seen_max_log_id == remote_max_log_id:
+        elif local_max_log_id == remote_max_log_id:
             if log_finished:
                 continue_polling = False
             else:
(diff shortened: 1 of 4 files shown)


civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
...
+def _jobs_follow_run_log(id, run_id):
+    client = civis.APIClient(return_type='raw')
+    seen_max_log_id = 0
+    continue_polling = True
+
+    while continue_polling:
+        response = client.jobs.list_runs_logs(id, run_id,
+                                              last_id=seen_max_log_id)
+        if 'civis-max-id' in response.headers:
+            remote_max_log_id = int(response.headers['civis-max-id'])
```
elsander: This variable is only set if the conditional is met. If the conditional is not met, the user will get a confusing error on line 210. What should happen if the header is not present? Maybe raise an informative error rather than continuing on in the function?

civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
...
+        log_finished = response.headers['civis-cache-control'] != 'no-store'
```
elsander: A lot of these items being pulled out of headers are mysterious to me. It would be helpful to have a comment saying what is being pulled out of the header and what the value you're expecting means.

civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
...
+        logs = response.json()
+        if logs:
+            seen_max_log_id = max(log['id'] for log in logs)
+            logs.sort(key=operator.itemgetter('createdAt', 'id'))
+        for log in logs:
+            print(' '.join((log['createdAt'], log['message'].rstrip())))
+        # if output is a pipe, write the buffered output immediately:
+        sys.stdout.flush()
+        if seen_max_log_id == remote_max_log_id:
```
elsander: What does it mean if this condition is met vs not? If it's not met, won't you be stuck in an infinite loop?

civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
...
+    continue_polling = True
+
+    while continue_polling:
+        response = client.jobs.list_runs_logs(id, run_id,
+                                              last_id=seen_max_log_id)
+        if 'civis-max-id' in response.headers:
+            remote_max_log_id = int(response.headers['civis-max-id'])
+        log_finished = response.headers['civis-cache-control'] != 'no-store'
+        logs = response.json()
+        if logs:
```
elsander: What if `if logs` is not met? Will you be stuck in an infinite loop?

civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
...
+
+def _jobs_follow_run_log(id, run_id):
+    client = civis.APIClient(return_type='raw')
+    seen_max_log_id = 0
+    continue_polling = True
+
+    while continue_polling:
+        response = client.jobs.list_runs_logs(id, run_id,
+                                              last_id=seen_max_log_id)
+        if 'civis-max-id' in response.headers:
```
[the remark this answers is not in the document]
aripollak: It's the maximum log id that has been seen on Platform. It's used to decide if we need to keep retrieving logs instead of waiting for the next 3 second poll interval.
Maybe I could rename the variable here to `max_log_id_seen_on_platform`?

civis/cli/_cli_commands.py
aripollak: That's okay and normal if there are no more log results; we should normally just keep polling.

civis/cli/_cli_commands.py
aripollak: If it's not met, we'll continue polling.

civis/cli/_cli_commands.py
aripollak: I guess the meaning of the value here doesn't really matter as much as the fact that the 'no-store' means Platform is no longer expecting to get any more logs for this run.
Is there a variable name that would make it clearer?

civis/cli/_cli_commands.py
```diff
@@ -156,6 +171,59 @@ def _str_table_result(cols, rows):
...
+        # if output is a pipe, write the buffered output immediately:
+        sys.stdout.flush()
+        log_finished = response.headers['civis-cache-control'] != 'no-store'
+        if remote_max_log_id is None:
+            time.sleep(_FOLLOW_POLL_INTERVAL_SEC)
+        elif seen_max_log_id == remote_max_log_id:
+            if log_finished:
+                continue_polling = False
+            else:
+                time.sleep(_FOLLOW_POLL_INTERVAL_SEC)
```
aripollak: I feel like it should be easier to DRY the sleep logic so we only have to do it in one place, but I wasn't able to come up with anything. Suggestions welcome.

aripollak: Maybe something like this to both document and DRY? It's more code, but might be clearer?

```py
        if remote_max_log_id is None:
            platform_has_more_logs_to_retrieve_now = False
        elif local_max_log_id == remote_max_log_id:
            platform_has_more_logs_to_retrieve_now = False
            if log_finished:
                continue_polling = False
        else:
            platform_has_more_logs_to_retrieve_now = True

        if continue_polling and not platform_has_more_logs_to_retrieve_now:
            time.sleep(_FOLLOW_POLL_INTERVAL_SEC)

```

civis/cli/_cli_commands.py
```diff
@@ -156,6 +165,55 @@ def _str_table_result(cols, rows):
     return '\n'.join(tb_strs)
 
 
+@click.command(
+    'follow-log',
+    help='Output live log from the most recent job run.' + _FOLLOW_LOG_NOTE)
+@click.argument('id', type=int)
+def jobs_follow_log(id):
```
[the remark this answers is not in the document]
elsander: Yeah, the intended behavior and output format would be helpful. It could also be helpful to mention how often it polls for new logs.

civis/cli/_cli_commands.py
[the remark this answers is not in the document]
elsander: Part of the issue is that I don't know how platform stores log information. Are you saying that each log line has its own id?

Currently the people who maintain the API clients are not people who work on the API itself, so an inline comment mentioning that we can check for new platform logs by looking for new ids would help clarify.

civis/cli/_cli_commands.py
elsander: The variable name makes sense to me, but I find the names of things in the Platform response confusing (`civis-cache-control`, `no-store`). I can take your word for it that this variable holds a bool of whether the log is done or not, but I wondered if there was any context you could add about what Platform is passing. This could provide extra context if, for example, I needed to modify the code in the future for some reason.

civis/cli/_cli_commands.py
elsander: Yeah, I think this is clearer.

civis/cli/_cli_commands.py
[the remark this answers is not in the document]
elsander: Update: my tab was out of date, so I didn't see the change you made to _FOLLOW_LOG_NOTE. That resolves what I'm looking for here.

elsander:
@aripollak I left a couple notes about places where inline comments would be helpful. Otherwise this looks good.

civis/cli/_cli_commands.py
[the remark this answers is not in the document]
aripollak: Ah, okay. Yeah, each log message (I'm not calling it a line because I think a single message could technically have multiple newlines) has an ID that should be unique within each run and is in chronological order based on when the message was originally sent. 
Here's what one of the log entries from `civis jobs list-runs-logs` looks like in YAML format, since that's the same endpoint gets used here:
```
- createdAt: '2019-11-04T15:56:44.492Z'
  id: 492602116
  level: info
  message: Queued
```

civis/cli/_cli_commands.py
aripollak: Currently it's either `max-age=120` or `no-store`, where the latter means that Platform has decided it might still be getting more logs. The `max-age=120` value is currently not used directly by the web UI either. It's in the same format as a [regular Cache-Control header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) to tell the client how long it can cache the results, but right now it's just used to tell the client when to stop polling for results.

PR #363, opened 2020-02-21
avoid duplicate parent job parameters when inferring joblib backend

mheilman:
This addresses a bug where submitting jobs from a parallel backend from `infer_joblib_backend` failed when used in a child job where `CIVIS_PARENT_JOB_ID` and `CIVIS_PARENT_RUN_ID` were set. Those parameter names would be duplicated, which causes platform to reject the job.

I tweaked `infer_joblib_backend` because the issue pertains to use of that function.  I decided to also replace rather than append parent job parameters in ` _ContainerShellExecutor.__init__` [here](https://github.com/civisanalytics/civis-python/blob/adc87f1a8ec2c7501b611f9213054234cccc14f1/civis/futures.py#L562-L568) because the behavior there also seemed problematic.

To do:
- [x] update changelog
- [ ] should I go ahead and bump the version number?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Added
 - Add `civis jobs follow-log` and `civis jobs follow-run-log` CLI commands (#359)
 ### Fixed
+- Fixed a bug related to duplicating parent job parameters when using `civis.parallel.infer_backend_factory`. (#363)
 ### Changed
 
 ## 1.12.1 - 2020-02-10
 ### Fixed
 - Fixed issue where client did not generate functions for deprecated API endpoints. (#353)
-
-- Fixed a bug related to duplicating parent job parameters when using `civis.parallel.infer_backend_factory`. (#363)
-
 ### Changed
 - Changed `ServiceClient` to raise `CivisAPIError`. (#355)
 - Updated documentation language for CivisML version. (#358)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## 1.12.1 - 2020-02-10
 ### Fixed
 - Fixed issue where client did not generate functions for deprecated API endpoints. (#353)
+
+- Fixed a bug related to duplicating parent job parameters when using `civis.parallel.infer_backend_factory`. (#363)
+
 ### Changed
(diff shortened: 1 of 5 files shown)


civis/tests/test_futures.py
```diff
@@ -325,24 +326,32 @@ def test_custom_scripts():
...
 
-    params = c.scripts.post_containers.call_args[1].get('params')
-    assert params is not None
-    assert {'name': 'CIVIS_PARENT_JOB_ID', 'type': 'integer',
-            'value': job_id} in params, \
-        "The creator's job ID wasn't passed to the child."
-    assert {'name': 'CIVIS_PARENT_RUN_ID', 'type': 'integer',
-            'value': run_id} in params, \
-        "The creator's run ID wasn't passed to the child."
+    params = sorted(c.scripts.post_containers.call_args[1].get('params'),
```
mheilman: The sorting here is maybe a bit excessive since the init function is just gonna tack on the list of params passed in above, but we don't technically care about the ordering that `__init__` adds them in

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## 1.12.1 - 2020-02-10
 ### Fixed
 - Fixed issue where client did not generate functions for deprecated API endpoints. (#353)
+
+- Fixed a bug related to duplicating parent job parameters when using `civis.parallel.infer_backend_factory`. (#363)
```
stephen-hoover: Request: This should go in the "unreleased" section of the changelog. This section is for changes that went out with the most recent release.

civis/tests/test_parallel.py
```diff
@@ -327,6 +343,22 @@ def _get_container(job_id):
     mock_make_factory.assert_called_once_with(**expected_kwargs)
 
 
+@mock.patch.object(civis.parallel, 'make_backend_factory')
+def test_infer_in_child_job(mock_make_factory, mock_child_job):
+    # Verify that infer_backend_factory doesn't include CIVIS_PARENT_JOB_ID and
+    # CIVIS_PARENT_RUN_ID since those will be automatically added later.
+    mock_client = mock.MagicMock()
```
stephen-hoover: Suggestion: Using `create_client_mock` will give you an autospecced APIClient mock, which should error on things like invalid / misspelled endpoint names.
```suggestion
    mock_client = create_client_mock()
```

CHANGELOG.md
mheilman: oops!

PR #365, opened 2020-02-28
account for when n_jobs is none

salilgupta1:
`sklearn==0.22.0` sets the default for `n_jobs` to `None` for `LogisticRegression` models https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression

This is a breaking change from `0.19.x` which set it to 1. The `joblib` library accounts for this already in a recent bug fix but we don't in our custom backend. Here's the fix.


jacksonllee:
The fix looks good. We need an entry in the change log.

mheilman:
For posterity, I think [this](https://github.com/joblib/joblib/blob/6afc9c3fc34d463842f1842110cfce6593cfbfdf/joblib/_parallel_backends.py#L232-L235) is the analogous part of joblib that checks for None for the multiprocessing backend.

salilgupta1:
> Are there already tests running through this method? If so we might want to add this test case.

I didn't find any tests for this function.

PR #366, opened 2020-03-02
Print NULL for `civis sql` output with Nones instead of crashing

aripollak:
Before this change, running `civis sql` with some cells being NULL results in this error:
```py
  File "/Users/apollak/dev/civis-python/civis/cli/_cli_commands.py", line 166, in _str_table_result
    [[len(_v) for _v in _r] for _r in [cols] + str_rows])
  File "/Users/apollak/dev/civis-python/civis/cli/_cli_commands.py", line 166, in <listcomp>
    [[len(_v) for _v in _r] for _r in [cols] + str_rows])
  File "/Users/apollak/dev/civis-python/civis/cli/_cli_commands.py", line 166, in <listcomp>
    [[len(_v) for _v in _r] for _r in [cols] + str_rows])
TypeError: object of type 'NoneType' has no len()
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Add `civis jobs follow-log` and `civis jobs follow-run-log` CLI commands (#359)
 ### Fixed
 - Fixed a bug related to duplicating parent job parameters when using `civis.parallel.infer_backend_factory`. (#363)
+- Fixed crashing on NULL fields in `civis sql` CLI command (#366)
 ### Changed
 
 ## 1.12.1 - 2020-02-10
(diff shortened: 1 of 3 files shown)


mheilman:
Could you please add a test and a changelog entry?

Should nulls be blanks instead? (From doing `SELECT NULL` in psql, it looks like psql uses blanks for nulls.)

aripollak:
> Could you please add a test and a changelog entry?
👍 to changelog entry. Do we want to start adding tests for extra CLI commands?

> Should nulls be blanks instead? (From doing `SELECT NULL` in psql, it looks like psql uses blanks for nulls.)

🤷‍♂ Maybe it should be configurable?

mheilman:
It looks like there's already [a relevant test](https://github.com/civisanalytics/civis-python/blob/bbd6c0f0b05383a6663223499400350f0e7c4869/civis/tests/test_cli.py#L124-L129) that would be pretty easy to copy/extend to check the `None` case.

PR #369, opened 2020-03-03
bug with from_existing

salilgupta1:
@jacksonllee take a look at what I found and let me know if this makes sense.

### DIFF
--- a/civis/ml/_model.py
+++ b/civis/ml/_model.py
@@ -1145,11 +1145,12 @@ class ModelPipeline:
                     verbose=args.get('DEBUG', False))
         klass.train_result_ = fut
 
-        # Set prediction template corresponding to training template
+        # Set prediction template corresponding to training
+        # or registration template
         template_id = int(container['from_template_id'])
         ids = find_one(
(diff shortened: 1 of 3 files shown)


civis/ml/_model.py
```diff
@@ -1146,6 +1148,8 @@ def from_existing(cls, train_job_id, train_run_id='latest', client=None):
         klass.train_result_ = fut
 
         # Set prediction template corresponding to training template
+        # The bug is here: We expect a job that is based off of a Training template
+        # but it is really based off of a Registration template
```
jacksonllee: `from_existing` looks correct to me, because this class method _can_ be used on its own (so not necessarily called by `register_pretrained_model`) and it does expect a training job ID from its function signature in order to infer the training template ID here. If there's a bug, it might be at your other comment above at `register_pretrained_model`. I'll chat with you offline to look at this together more closely.

salilgupta1: Yea, you're right. I think the issue is further up. What do you recommend as the fix to the problem?

jacksonllee:
The change log needs an entry for this fix.

PR #378, opened 2020-03-30
remove support for py 2.7 and 3.4

mheilman:


setup.py
```diff
@@ -6,10 +6,7 @@
 
 CLASSIFIERS = [
     'Programming Language :: Python',
-    'Programming Language :: Python :: 2',
     'Programming Language :: Python :: 3',
-    'Programming Language :: Python :: 2.7',
-    'Programming Language :: Python :: 3.4',
```
jacksonllee: We don't know how these trove classifiers are used by whoever/whatever that tracks the published packages on PyPI. Since the codebase is becoming Python 3 only now, we might as well throw in `Programming Language :: Python :: 3 :: Only` ([reference](https://pypi.org/pypi?%3Aaction=list_classifiers)), though no idea how useful it is for us, if at all. Okay to not do anything here -- your call.

PR #379, opened 2020-04-02
[APIC-273] Add Table Columns Parameter

uttercm:
**JIRA Ticket:**
[APIC-273](https://civisanalytics.atlassian.net/browse/APIC-273)

**Summary:**
Adding the table_columns parameter to io methods.

**Testing:**
- [x] Updated specs
- [x] Performed manual tests

Manual Tests
- [x] import to a pre-existing table with non-drop mode and table columns not specified [Job 70574228](https://platform.civisanalytics.com/spa/jobs/70574228)
- [x] import to pre-existing table with existing_table_rows=drop and table columns not specified [Job 70573579](https://platform.civisanalytics.com/spa/jobs/70573579)
- [x] import to pre-existing table with existing_table_rows=drop and table columns specified [Job 70572994](https://platform.civisanalytics.com/spa/jobs/70572994)
- [x] import to non-existent table with table columns specified [Job 70455477](https://platform.civisanalytics.com/spa/jobs/70455477)
- [x] multiple file import using `civis_file_to_table` [Job 70579238](https://platform.civisanalytics.com/spa/jobs/70579238)

Notes:

**Additional Info:**

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1024,11 +1024,11 @@ def civis_file_to_table(file_id, database, table, client=None,
                                      headers, delimiter, hidden)
 
     (cleaned_file_ids, headers, compression, delimiter,
-     cleaning_table_columns) = _process_cleaning_results(
+     cleaned_table_columns) = _process_cleaning_results(
         cleaning_futures, client, headers, need_table_columns, delimiter
     )
 
-    table_columns = table_columns or cleaning_table_columns
+    table_columns = table_columns or cleaned_table_columns
 
     source = dict(file_ids=cleaned_file_ids)
     destination = dict(schema=schema, table=table, remote_host_id=db_id,
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -617,6 +617,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                        max_errors=None, existing_table_rows="fail",
                        diststyle=None, distkey=None,
                        sortkey1=None, sortkey2=None,
+                       table_columns=None,
                        headers=None, credential_id=None,
                        primary_keys=None, last_modified_keys=None,
                        execution="immediate",
@@ -660,6 +661,9 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
         The column to use as the sortkey for the table.
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
+    table_columns : list[Dict[str, str]], optional
+        An list of dictionaries corresponding to the columns in the source file.
+        Each dictionary should have keys for column "name" and "sqlType"
     headers : bool, optional [DEPRECATED]
         Whether or not the first row of the file should be treated as
         headers. The default, ``None``, attempts to autodetect whether
@@ -740,6 +744,7 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
                               existing_table_rows=existing_table_rows,
                               diststyle=diststyle, distkey=distkey,
                               sortkey1=sortkey1, sortkey2=sortkey2,
+                              table_columns=table_columns,
                               delimiter=delimiter, headers=headers,
                               credential_id=credential_id,
                               primary_keys=primary_keys,
@@ -756,6 +761,7 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
                  max_errors=None, existing_table_rows="fail",
                  diststyle=None, distkey=None,
                  sortkey1=None, sortkey2=None,
+                 table_columns=None,
                  delimiter=",", headers=None,
                  primary_keys=None, last_modified_keys=None,
                  escaped=False, execution="immediate",
@@ -795,6 +801,9 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
         The column to use as the sortkey for the table.
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
+    table_columns : list[Dict[str, str]], optional
+        An list of dictionaries corresponding to the columns in the source file.
+        Each dictionary should have keys for column "name" and "sqlType"
     delimiter : string, optional
         The column delimiter. One of ``','``, ``'\\t'`` or ``'|'``.
     headers : bool, optional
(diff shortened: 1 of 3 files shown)


elsander:
Tagging @salilgupta1 since Team Sparkle owns the API clients now.

uttercm:
> Tagging @salilgupta1 since Team Sparkle owns the API clients now.

Oh sorry, I forgot that switch happened.

civis/io/_tables.py
```diff
@@ -918,6 +931,10 @@ def civis_file_to_table(file_id, database, table, client=None,
         The column to use as the sortkey for the table.
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
+    table_columns : list[Dict[str, str]], optional
```
mheilman, line 934: Is a useful error message going to come back if this is incorrectly specified?

civis/io/_tables.py
```diff
@@ -999,16 +1016,19 @@ def civis_file_to_table(file_id, database, table, client=None,
 
     # Use Preprocess endpoint to get the table columns as needed
     # and perform necessary file cleaning
-    need_table_columns = not table_exists or existing_table_rows == 'drop'
+    need_table_columns = ((not table_exists or existing_table_rows == 'drop')
```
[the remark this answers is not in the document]
zacharydfriedlander: So FWIW `table_columns` actually _won't_ be ignored in this case. If `table_columns` is provided, the import will copy only those columns. This is useful if, for instance, importing a file with columns `a` and `b` into a table with columns `a`, `b`, and `c`.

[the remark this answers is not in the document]
zacharydfriedlander: The API requires `table_columns` if we can't use the pre-existing table columns (i.e. the table doesn't exist, or the table is being dropped).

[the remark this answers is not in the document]
mheilman: oh, interesting. sorry, i misread this

civis/io/_tables.py
zacharydfriedlander: The import will fail if this is the case.

civis/io/_tables.py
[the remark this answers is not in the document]
zacharydfriedlander: No worries - looking at it again, it doesn't look like we explicitly say that anywhere in the client docs. (It is, however, mentioned, in the API docs). We could add it to the docstring description of `table_columns`?

civis/io/_tables.py
```diff
@@ -660,6 +661,10 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
         The column to use as the sortkey for the table.
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
+    table_columns : list[Dict[str, str]], optional
+        An list of dictionaries corresponding to the columns in
```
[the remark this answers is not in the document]
uttercm: ya, I don't think we want to explain sqlTypes here

civis/io/_tables.py
mheilman: Will it fail with an informative error message?  I'd prefer a useful error message from platform, but if that's not available, then it seems like input validation here might be warranted.

uttercm: the import will fail with saying the table could not be created due to the bad sqlType provided example: `Job::UserError: Failed to create table: type "json" does not exist`

mheilman: Hmm, that's not particularly helpful but better than just "your job failed".  I'll follow up internally about maybe filing an internal issue about improving that error message from platform. I'm OK with leaving this as is.

civis/io/_tables.py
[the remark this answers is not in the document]
uttercm: I added `The import will only copy these columns regardless if there are more columns in the table.` to the `table_columns` description

uttercm:
manual tests ran added to description of the PR

civis/io/_tables.py
jkulzick: If i'm reading this error correctly, a "json" data type was provided in `table_columns` but its not a valid data type on Redshift?

jkulzick: This is the error returned by Redshift. The API simply returns the error the database provides. Creating and maintaining some complex mapping of Redshift/Postgres errors to what we consider more user friendly errors is not something we have resourcing for right now.

If there are other scenarios that are providing unhelpful errors due to validation we are doing in our API then I'm all for addressing that.

uttercm: I thought you meant, sqlType was incorrectly specified, I can add a validation on if it's a list.

civis/io/_tables.py
```diff
@@ -981,6 +1001,9 @@ def civis_file_to_table(file_id, database, table, client=None,
         file_id = [file_id]
     if schema is None:
         raise ValueError("Provide a schema as part of the `table` input.")
+    if table_columns is not None and not isinstance(table_columns, list):
```
mheilman, line 1004: Sorry, after thinking about this some more, I think platform would ideally be providing this error message rather than API client.  Since you already added it, though, and it might be helpful in the short term, I'm fine with leaving it in (or whatever you'd prefer).

mheilman:
I'm cool with this.  I defer to @zacharydfriedlander for final approval

mheilman:
also, thanks!

civis/io/_tables.py
uttercm: In looking through it further and talking with @jkulzick, I'm going to remove this error check so we rely on platform's error message for this.

civis/io/_tables.py
```diff
@@ -981,6 +1001,9 @@ def civis_file_to_table(file_id, database, table, client=None,
...
+        raise ValueError("Provide table_columns as a list of dictionaries"
+                         " with keys of `name` and `sqlType`.")
```
jkulzick: We shouldn't check if `table_columns` is a `list`. If we're going to implement type checking in the client then we should do that everywhere.

If we wanted to check if `table_columns` has keys of `name` and `sqlType` then I would be okay with that but I think its unnecessary because the API will tell us thats the problem. Now someone will have to maintain this code in the client and update it when the API is updated versus just using the API.

For reference, the `fut` should return the below error when a `table_columns` is passed without the required keys (e.g., `sqlTypes`).

image

jkulzick: @mheilman @zacharydfriedlander fyi

mheilman: oh, nice.  i'm glad the error is pretty informative

PR #381, opened 2020-04-06
ENH add method to retrieve run outputs from future

elsander:
This PR adds a new method to `CivisFuture` objects, `.outputs()`, which blocks on job completion and returns the run outputs. I kept it as minimal as possible to make it generalizeable and easy to maintain, while still making it easy to maintain.

I tested the method manually on succeeded and failed container scripts using both `CivisFuture` and `ContainerFuture`, and also tested on a successful sql script. It's always challenging to add meaningful unit tests to the API client, so I'm happy for feedback on the unit testing I added.

### DIFF
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -235,7 +235,7 @@ class CivisFuture(PollableResult):
 
         Raises
         ------
-        CivisJobFailure
+        civis.base.CivisJobFailure
             If the job fails.
         """
         self.result()
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -222,7 +222,20 @@ class CivisFuture(PollableResult):
                 self._set_api_exception(exc=e)
 
     def outputs(self):
-        """Block on job completion and return a list of run outputs."""
+        """
+        Block on job completion and return a list of run outputs if the
+        job is successful.
+
+        Returns
+        -------
+        list[dict]
+            List of run outputs from a successfully completed job.
+
+        Raises
+        ------
+        CivisAPIError
+            If the job fails.
+        """
         self.result()
         outputs = self.client.jobs.list_runs_outputs(self.job_id, self.run_id)
         return outputs
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -223,8 +223,10 @@ class CivisFuture(PollableResult):
 
     def outputs(self):
         """
-        Block on job completion and return a list of run outputs if the
-        job is successful.
+        Block on job completion and return a list of run outputs.
+
+        The method will only return run outputs for successful jobs.
+        Failed jobs will raise an exception.
 
         Returns
         -------
@@ -233,7 +235,7 @@ class CivisFuture(PollableResult):
 
         Raises
         ------
-        CivisAPIError
+        CivisJobFailure
             If the job fails.
         """
         self.result()
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -54,8 +54,7 @@ class JobCompleteListener(SubscribeCallback):
 
 
 class CivisFuture(PollableResult):
-    """
-    A class for tracking future results.
+    """A class for tracking future results.
 
     This class will attempt to subscribe to a Pubnub channel to listen for
     job completion events. If you don't have access to Pubnub channels, then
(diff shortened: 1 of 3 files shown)


civis/tests/test_futures.py
```diff
@@ -183,6 +183,20 @@ def test_set_api_result_failed(self, mock_subscribe, mock_api):
         assert result._state == 'FINISHED'
         with pytest.raises(CivisJobFailure):
             result.result()
+        with pytest.raises(CivisJobFailure):
+            result.outputs()
+
+    def test_outputs_succeeded(self):
```
arthir, line 189: Can we add a test for if the job fails? Does `list_run_outputs` return a sensible result?

civis/futures.py
```diff
@@ -221,6 +221,12 @@ def _poll_and_set_api_result(self):
             except Exception as e:
                 self._set_api_exception(exc=e)
 
+    def outputs(self):
+        """Block on job completion and return a list of run outputs."""
+        self.result()
```
stephen-hoover, line 226: Could you use a `concurrent.futures.wait` instead? The `.result()` will raise if the job fails, but you could still have outputs on a failed job that you want to retrieve.

mheilman: Hmm... raising an exception if the job fails seems desirable, by default at least.  I think a user generally  wouldn't want outputs if the job failed. If they have some situation where that's not the case, they could still use `wait` and the runs_outputs endpoint.

🤷‍♂

civis/tests/test_futures.py
elsander: If the job fails, it will fail during the `.result()` call, and the user will see the same error that they would get from `.result()`. My hope is that this would be an error a user would be familiar with. The addition to the test on line 186 there is meant to check for this case.

civis/futures.py
stephen-hoover: Erroring on a failed job is a reasonable thing to do here, but it's not what I personally would have expected from the function name and docstring. I have jobs which do produce useful output on failure (for example, a validation step in a workflow which runs Great Expectations and outputs table details whether it passes or fails, or jobs which write debug logs to run outputs).

I guess the most important thing is to document the behavior on job failure in the docstring. You could imagine a flag to select the behavior on failure, but maybe that's adding too much complexity.

mheilman: How about we continue calling `result()` here and adding a full docstring with a `Raises` section that explains that the particular exception will be raised when the job fails?

stephen-hoover: That would work for me, thanks!

elsander:
@mheilman Okay, ready for another round of review.

civis/futures.py
```diff
@@ -221,6 +221,25 @@ def _poll_and_set_api_result(self):
             except Exception as e:
                 self._set_api_exception(exc=e)
 
+    def outputs(self):
+        """
+        Block on job completion and return a list of run outputs if the
```
mheilman, line 226: Sorry to be nitpicky, but I think it'd be more in line with [numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html#sections) to have this sentence be shorter and fit on the first line like it was initially (`"""Block on job completion and return a list of run outputs.`). I think the "if the job is successful" part could be a separate sentence (see "Extended Summary" in the numpydoc page).

civis/futures.py
```diff
@@ -221,6 +221,25 @@ def _poll_and_set_api_result(self):
...
+        job is successful.
+
+        Returns
+        -------
+        list[dict]
+            List of run outputs from a successfully completed job.
+
+        Raises
+        ------
+        CivisAPIError
```
mheilman, line 236: I think it'd be nice to have this be `civis.base.CivisAPIError` so users don't have to hunt for how to import it.  I don't feel strongly about this since [numpydoc's](https://numpydoc.readthedocs.io/en/latest/format.html#sections) example for "Raises" just says "LinAlgException" rather than the full path.

mheilman: Also, should this be CivisJobFailure instead?  I didn't look into it too deeply, but I'm surprised it'd be CivisAPIError rather than CivisJobFailure. 🤷‍♂

elsander: Good catch! My text editor autocompleted that one and I didn't notice that it was incorrect.

elsander:
@mheilman ready for another round!

elsander:
I used the newline to match the formatting in the `CivisFuture` docstring itself, although it looks like we then don't use it in `ContainerFuture`, so not consistent throughout. I can update it in CivisFuture as well.

elsander:
Ready for re-approval @mheilman

PR #382, opened 2020-04-17
ENH parse_service_api_spec outside of serviceclient object

elsander:
This PR moves API spec parsing for services outside of the `ServiceClient` object. This allows a user to parse a spec without an API key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `table_columns` parameter to `civis.io.civis_file_to_table`, `civis.io.dataframe_to_civis`, and `civis.io.csv_to_civis` (#379)
 
 ### Fixed
-
+- No longer require ServiceClient to be instantiated to parse a
+  service api spec. (#382)
 - Fixed/relaxed version specifications for click, jsonref, and jsonschema. (#377)
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,16 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `table_columns` parameter to `civis.io.civis_file_to_table`, `civis.io.dataframe_to_civis`, and `civis.io.csv_to_civis` (#379)
 
 ### Fixed
-- No longer require ServiceClient to be instantiated to parse a
-  service api spec. (#382)
 - Fixed/relaxed version specifications for click, jsonref, and jsonschema. (#377)
 
 ### Removed
 
 - Removed support for Python 2.7 and 3.4. (#378)
+
+### Changed
+- No longer require ServiceClient to be instantiated to parse a
+  service api spec. (#382)
+
 
 ## 1.13.1 - 2020-03-06
 ### Added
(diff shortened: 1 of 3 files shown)


civis/service_client.py
```diff
@@ -28,6 +28,40 @@ def auth_service_session(session, client):
...
+    methods = []
+    for verb, op in operations.items():
+        method = parse_method(verb, op, path)
+        if method is None:
+            continue
+        methods.append(method)
+    return modified_base_path, methods
+
+
+def parse_service_api_spec(api_spec, service_client=None):
```
mheilman, line 49: I think we should add a docstring here.

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added `table_columns` parameter to `civis.io.civis_file_to_table`, `civis.io.dataframe_to_civis`, and `civis.io.csv_to_civis` (#379)
 
 ### Fixed
-
+- No longer require ServiceClient to be instantiated to parse a
```
mheilman, line 12: I think this should go under "Changed".

Also, I think this is technically a breaking change, unless we want to keep the `parse_path` and `parse_api_spec` functions and have them call the new ones.  I think it'd be a good idea to replace `## Unreleased` here with `## 2.0.0 (Unreleased)` here as a reminder if we're going to keep the breaking change.

civis/service_client.py
```diff
@@ -28,6 +28,40 @@ def auth_service_session(session, client):
     session.get(auth_url)
 
 
+def _parse_service_path(path, operations, service_client=None):
```
mheilman, line 31: How about having this function just take "root_path" or something like?  Passing in the whole service_client object seems unnecessary.

CHANGELOG.md
elsander: Oops, I thought I moved it to "changed", my mistake.

I can add the note for 2.0.0. Would making this breaking change impact when you'll want to cut a release? I know there's some other changes throughout the repo that are set for the next major version.

As an alternative, I could follow your suggestion to make it a non-breaking change, and add a deprecation tag that the duplicate methods will be dropped in 2.0.0.

mheilman: If you think it wouldn't be too bad to make some pass-through calls with deprecation warnings, that seems better.

Regarding the "will be removed in v2.0.0" warnings, I think we could just update those to say "v3.0.0" or "a future release" to avoid delaying release.  Same for the 2.0 milestone in github.

elsander:
This should be ready for another pass @mheilman

civis/service_client.py
```diff
@@ -28,6 +28,54 @@ def auth_service_session(session, client):
...
+    Parse an OpenAPI (Swagger) specification into a dictionary of classes
+    where each class represents an endpoint resource and contains
+    methods to make http requests on that resource.
+
+    Parameters
+    ----------
+    api_spec : OrderedDict
+        The Civis Service API specification to parse.  References should be
+        resolved before passing, typically using jsonref.JsonRef().
+
```
mheilman, line 61: `root_path` is missing here

elsander:
@mheilman fixed

PR #383, opened 2020-04-21
prep for 1.14.0 release

mheilman:


docs/source/ml.rst
```diff
@@ -402,7 +402,7 @@ Sharing Models
 ==============
 
 Models produced by CivisML can't be shared directly through the Civis Platform
-UI or API. The :module:`~civis.ml` namespace provides functions which will
```
mheilman, line 405: I noticed that this wasn't getting parsed correctly on RtD.

https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-mod

PR #384, opened 2020-04-22
[APIC-278] bugfix root path in parse

jsfalk:
This PR fixes a bug in civis-python 1.14.0 that breaks autospec generation. The issue arises because the root path is not passed to the parsing function when generating classes. This PR also replaces an undefined variable with the defined variable in a now deprecated function.


mheilman:
oh, could you also update the changelog?  Please go ahead and do `## 1.14.1 - 2020-04-22`

PR #387, opened 2020-05-01
Update primary key docstrings

zacharydfriedlander:
Add an explicit statement that primary key columns should not contain `NULL` values. We are concurrently updating the API documentation to make this clear there as well.


zacharydfriedlander:
Hm, something seems not right with this fork - it has old changes and such that should have already been merged.

PR #389, opened 2020-05-05
[APIC-238] Make CLI exit code nonzero when response status code != 2xx

hinnefe2:
When using the python API client CLI tool a command which fails can still exit with a 0 exit code. This can lead Platform to mark jobs which didn't successfully complete with a green checkmark or to mistakenly progress workflows when a step didn't work.

MWE:
Running 
```
civis tables post-scan some_db_id some_schema some_table
```
Generates logs:
```
2020-05-05 04:14:37 PM code: 404
2020-05-05 04:14:37 PM error: not_found
2020-05-05 04:14:37 PM errorDescription: The requested resource could not be found.
```
But the job is still marked as successful.

### DIFF
--- a/civis/cli/__main__.py
+++ b/civis/cli/__main__.py
@@ -162,10 +162,12 @@ def invoke(method, path, op, *args, **kwargs):
     with open_session(get_api_key(), user_agent=CLI_USER_AGENT) as sess:
         response = sess.request(**request_info)
 
-    # Print the response to stderr if there was an error.
+    # Print the response to stderr and set exit code to 1 if there was an error
     output_file = sys.stdout
+    exit_code = 0
     if response.status_code != 200:
         output_file = sys.stderr
+        exit_code = 1
 
     # Print the output, if there is any.
     # For commands such as DELETE /scripts/containers/{script_id}/runs/{id},
@@ -184,6 +186,8 @@ def invoke(method, path, op, *args, **kwargs):
         # Otherwise, do nothing.
         if response.text.strip():
             print("Error parsing response: {}".format(e), file=sys.stderr)
+
+    sys.exit(exit_code)
 
 
 def retrieve_spec_dict(api_version="1.0"):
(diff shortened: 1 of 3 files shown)


civis/cli/__main__.py
```diff
@@ -162,10 +162,12 @@ def invoke(method, path, op, *args, **kwargs):
     with open_session(get_api_key(), user_agent=CLI_USER_AGENT) as sess:
         response = sess.request(**request_info)
 
-    # Print the response to stderr if there was an error.
+    # Print the response to stderr and set exit code to 1 if there was an error
     output_file = sys.stdout
+    exit_code = 0
     if response.status_code != 200:
```
mheilman, line 168: hmmm... i think we might need to change this to capture any 2xx statuses rather than just 200.  i dunno if we use 201 errors anywhere, for example, but a 201 response for object creation is reasonable. https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

WillRaphaelson:
nice

civis/tests/test_cli.py
```diff
@@ -81,9 +81,39 @@ def test_blank_output(mock_session):
     # is blank.
     session_context = mock_session.return_value.__enter__.return_value
     session_context.request.return_value.json.side_effect = ValueError()
+    session_context.request.return_value.status_code = 200
 
     op = {"parameters": []}
-    invoke("WIBBLE", "/wobble/wubble", op)
+
+    with pytest.raises(SystemExit) as pytest_wrapped_e:
+        invoke("WIBBLE", "/wobble/wubble", op)
```
mheilman, line 89: For my own understanding, did the `invoke` command raise a `SystemExit` previously that pytest didn't count as an error such that the `pytest.raises` wasn't necessary?  That is, the `pytest.raises` is just needed so that the test can check `.value.code`, right?

hinnefe2: I don't think `invoke` was raising `SystemExit` previously. See for example the [failing tests](https://travis-ci.com/github/civisanalytics/civis-python/jobs/328391483) from the first commit in this PR. Now I'm kind of curious what happens under the hood when a python program ends without a `sys.exit()` but a quick bit of googling didn't turn up anything ...

PR #390, opened 2020-05-08
update pytest to 5.x

mheilman:
this also fixes some tests using the old message kwarg incorrectly.

see https://docs.pytest.org/en/5.0.0/deprecations.html#message-parameter-of-pytest-raises


mheilman:
i don't think it needs a changelog entry since it's dev stuff only

PR #391, opened 2020-05-08
fix futures for python 3.8 and enable python 3.8 builds

### LINKED ISSUE #370: Python 3.8 incompatibility with `read_civis`
Running `civis.io.read_civis` using Python 3.8 raises the following exception:
```
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/miniconda3/envs/survey-toolbox-python/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/miniconda3/envs/survey-toolbox-python/lib/python3.8/site-packages/civis/polling.py", line 44, in run
    poller_result = self.poller(*self.poller_args)
  File "/miniconda3/envs/survey-toolbox-python/lib/python3.8/site-packages/civis/polling.py", line 168, in _check_result
    self._set_api_result(self._last_result)
  File "/miniconda3/envs/survey-toolbox-python/lib/python3.8/site-packages/civis/polling.py", line 182, in _set_api_result
    self.set_result(result)
  File "/miniconda3/envs/survey-toolbox-python/lib/python3.8/concurrent/futures/_base.py", line 524, in set_result
    raise InvalidStateError('{}: {!r}'.format(self._state, self))
concurrent.futures._base.InvalidStateError: FINISHED: <CivisFuture at 0x10fcd1c40 state=succeeded returned NoneType>
```

Note that this occurs regardless of whether a valid table name is passed in to the function.

mheilman:
closes #370

### DIFF
--- a/civis/base.py
+++ b/civis/base.py
@@ -161,6 +161,10 @@ class CivisAsyncResultBase(futures.Future):
         self.poller = poller
         self.poller_args = poller_args
         self.polling_interval = polling_interval
+
+        # TODO explain this. also, maybe rename to _state_override or something
+        self.__state = None
+
         if api_key is not None:
             warnings.warn('The "api_key" parameter is deprecated and will be '
                           'removed in v2. Please use the `client` parameter '
@@ -220,10 +224,27 @@ class CivisAsyncResultBase(futures.Future):
     @property
     def _state(self):
         """State of the CivisAsyncResultBase in `future` language."""
+        if self.__state is not None:
+            return self.__state
         with self._condition:
             return STATE_TRANS[self._civis_state]
 
     @_state.setter
     def _state(self, value):
         # Ignore attempts to set the _state from the `Future` superclass
-        pass
+        # TODO explain
+        self.__state = value
+
+    def set_result(self, result):
+        # TODO explain this
+        self.__state = futures._base.RUNNING
+        super().set_result(result)
+        self.__state = None
+
+    def set_exception(self, exception):
+        # TODO explain this
+        self.__state = futures._base.RUNNING
+        super().set_exception(exception)
+        self.__state = None
+
+    # TODO set docstrings for set_result and set_exception
--- a/civis/base.py
+++ b/civis/base.py
@@ -161,10 +161,6 @@ class CivisAsyncResultBase(futures.Future):
         self.poller = poller
         self.poller_args = poller_args
         self.polling_interval = polling_interval
-
-        # TODO explain this. also, maybe rename to _state_override or something
-        self.__state = None
-
         if api_key is not None:
             warnings.warn('The "api_key" parameter is deprecated and will be '
                           'removed in v2. Please use the `client` parameter '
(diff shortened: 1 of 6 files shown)


mheilman:
I'm still not sure why some of the tests pass in 3.6 but not 3.8 (e.g., [this one](https://github.com/civisanalytics/civis-python/blob/fd7700c80f1757b90f1f63a5728961e1d039c988/civis/tests/test_polling.py#L66-L75)) without explicitly setting `daemon=True` so that polling threads don't keep the Python (or pytest) process from quitting.

From [the threading.Thread docs](https://docs.python.org/3.8/library/threading.html#threading.Thread):

> Its initial value is inherited from the creating thread; the main thread is not a daemon thread and therefore all threads created in the main thread default to daemon = False.
> The entire Python program exits when no alive non-daemon threads are left.

civis/base.py
```diff
@@ -220,10 +224,27 @@ def _civis_state(self):
...
     @_state.setter
     def _state(self, value):
         # Ignore attempts to set the _state from the `Future` superclass
-        pass
+        # TODO explain
+        self.__state = value
+
+    def set_result(self, result):
+        # TODO explain this
+        self.__state = futures._base.RUNNING
```
mheilman, line 240: This fiddling with the job state is pretty unpleasant, but I didn't see a clear alternative.

The challenge is that the state of a CivisFuture (or asyncresultbase, more specifically) is simply a [translation](https://github.com/civisanalytics/civis-python/blob/fd7700c80f1757b90f1f63a5728961e1d039c988/civis/base.py#L220-L224) of the Civis job state, which seems to violate an assumption in python's Futures that the state is only changed within the set_result() or set_exception() functions.

In 3.8, the set_result/set_exception functions [now check](https://github.com/python/cpython/blob/dec367261e7e2bb4dd42feeb58031abed2ade683/Lib/asyncio/futures.py#L236-L249) the state and make sure it's not already done before setting it to done.

civis/base.py
```diff
@@ -161,6 +161,10 @@ def __init__(self, poller, poller_args,
         self.poller = poller
         self.poller_args = poller_args
         self.polling_interval = polling_interval
+
+        # TODO explain this. also, maybe rename to _state_override or something
+        self.__state = None
```
salilgupta1, line 166: I think you should rename this as `_state_override` because I was getting confused with how this was related to `self._state`.

civis/base.py
```diff
@@ -220,10 +224,27 @@ def _civis_state(self):
     @property
     def _state(self):
         """State of the CivisAsyncResultBase in `future` language."""
+        if self.__state is not None:
+            return self.__state
```
salilgupta1, line 228: should we have some check here that is validating that `self._state` and `self.__state` are in sync?

civis/base.py
salilgupta1: Why even have the override? Why not directly update `self._state`

salilgupta1: I'm a little confused. The link in your comment points to a `Futures` class that is [not the same](https://github.com/python/cpython/blob/dec367261e7e2bb4dd42feeb58031abed2ade683/Lib/asyncio/futures.py#L30) as `concurrent.futures.Future`. 

Is this what you meant: https://github.com/python/cpython/blob/3.8/Lib/concurrent/futures/_base.py#L523

mheilman: ah, yup. sorry about the bad link.  i updated my original comment.

civis/base.py
mheilman: The issue is that we don't want them to be in sync while the `concurrent.futures.Future.set_result()` and `set_exception()` calls are happening. At that point, the platform job has finished, so (before this PR) `CivisFuture.state` would return `futures._base.FINISHED` when python expects it to return `futures._base.RUNNING`.

civis/base.py
mheilman: > Why not directly update `self._state`

`civis.base.AsyncResultBase._state` is actually [a function with a property decorator](https://github.com/civisanalytics/civis-python/blob/bb9c6df4023f96897ec728ce4251b2016322ad6d/civis/base.py#L226-L229) that ignores attempts to set it, the idea being that platform provides the actual state of the future object. I think this is probably undesirable, but I didn't see a simple way to unwind that assumption without a substantial refactor, which doesn't seem warranted here since it has worked pretty reliably for a long time.

civis/base.py
```diff
@@ -220,10 +224,27 @@ def _civis_state(self):
...
         with self._condition:
             return STATE_TRANS[self._civis_state]
 
     @_state.setter
     def _state(self, value):
         # Ignore attempts to set the _state from the `Future` superclass
-        pass
+        # TODO explain
+        self.__state = value
```
mheilman, line 236: this part is weird. I'm going to try to find a way to avoid changing the `_state` setter here

mheilman:
After further consideration, I think it probably makes the most sense to just override `set_result` and `set_exception`.  It might be annoying to keep those in sync, but the workarounds seem ugly.

PR #393, opened 2020-05-12
DEP Remove file_id deprecation warning

stephen-hoover:
There was no way for users to avoid this deprecation warning when calling `civis.io.civis_file_to_table`, and it was also triggered by other functions which called `civis_file_to_table` (such as `dataframe_to_civis`). The original thought had been to replace the singular `file_id` parameter with a plural `file_ids` parameter in v2, but per discussion in #360, the singular will be acceptable.

Closes #360.

PR #394, opened 2020-05-12
fix some minor issues that flake8 3.8 finds

mheilman:
related to #393


mheilman:
Ha, yeah, I was wondering the same thing.  It means visually ambiguous (e.g., "l" vs "I"), so it kinda makes sense (though fixed width fonts usually render those differently: `l` vs `I`).

https://www.flake8rules.com/rules/E741.html

stephen-hoover:
> Ha, yeah, I was wondering the same thing. It means visually ambiguous (e.g., "l" vs "I"), so it kinda makes sense (though fixed width fonts usually render those differently: `l` vs `I`).
> 
> https://www.flake8rules.com/rules/E741.html

That makes a lot of sense, now that you point it out. I'm happy to have that check in.

PR #395, opened 2020-05-14
[ASE-146] Fix civis parallel backend bug

evanderh:
When using `civis.parallel.make_backend_factory` with joblib, cancelling a child job fails to cancel all the remaining child jobs and hangs indefinitely

To reproduce:
- Check out the minimal example: `git checkout 38b0e02`
- Run example reproduce script: `./examples/parallel_cancel.py`
- Through the platform UI, cancel the first job

What's going wrong:
- The first job gets cancelled, calling [_CivisBackend.abort_everything](https://github.com/civisanalytics/civis-python/blob/master/civis/parallel.py#L850) which then should cancel all the other jobs
- The second job is cancelled through the api in [ContainerFuture.cancel](https://github.com/civisanalytics/civis-python/blob/master/civis/futures.py#L334)
- In the second job's [callback](https://github.com/civisanalytics/civis-python/blob/master/civis/parallel.py#L733), `future.cancelled() == False`, because the future hasn't resolved yet (I think?) and isn't aware that `is_cancel_requested == True`
- This causes the third job to not get cancelled and the parent script hangs open indefinitely

Fix:
- Disclaimer: This likely isn't the best fix but I just wanted to try an initial fix to see if it worked
- I made the callback aware of `is_cancel_requested`, which correctly causes the the third job to get cancelled
- The parent script correctly raises a CancelledError instead of hanging indefinitely


civis/parallel.py
```diff
@@ -730,7 +730,11 @@ def _fetch_result(fut):
                 fut.remote_func_output = exc
             else:
                 fut.result_fetched = True
-                if not fut.cancelled() and not fut.exception():
+                cancelled = (fut.cancelled()
+                             or (hasattr(fut, 'is_cancel_requested')
```
mheilman, line 734: Looking at the `is_cancel_requested` attribute here seems reasonable to me since platform doesn't report a job as cancelled until well after the cancellation request is made.

I think it'd be more typical to do the following here, though, rather than using `hasattr`

```
cancelled = fut.cancelled()
try:
    cancelled |= fut.is_cancel_requested
except AttributeError:
    pass
```

That said, I think it'd be best not to create a new attribute `is_cancel_requested` on the `Future` object. Could we check the run info here to get that from the API?

mheilman:
I think the CI errors might be resolved by bringing the branch up-to-date with master. See #394

civis/parallel.py
```diff
@@ -730,7 +730,12 @@ def _fetch_result(fut):
                 fut.remote_func_output = exc
             else:
                 fut.result_fetched = True
-                if not fut.cancelled() and not fut.exception():
+                cancelled = fut.cancelled()
+                try:
+                    cancelled |= fut._result.is_cancel_requested
```
evanderh, line 735: @mheilman It's a bit caveman-like but this works, your thoughts?

mheilman: seems reasonable. could you add a comment, though?

e.g., "After the user requests cancellation, a script run will stay in the running state with "is_cancel_requested" set to True for a while to allow for cleanup logic. Here, we want to treat runs in that state the same way we treat runs in the cancelled state."

examples/parallel_cancel.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
mheilman, line 1: Do you intend to add this to the repo, or was it just for testing?

civis/tests/test_parallel.py
```diff
@@ -482,6 +482,22 @@ def test_result_eventual_failure(mock_civis):
...
+def test_result_running_and_cancel_requested(mock_civis):
+    callback = mock.MagicMock()
+    mock_civis.io.civis_to_file.side_effect = make_to_file_mock('spam')
+    response = Response({'is_cancel_requested': True,
+                         'state': 'running'})
+    client = mock.MagicMock()
+    client.scripts.post_cancel.return_value = response
+    fut = ContainerFuture(1, 2, client=client)
+    fut.set_result(response)
+    civis.parallel._CivisBackendResult(fut, callback)
```
mheilman, line 495: Is this to emulate what the joblib backend does? Would it be too difficult to mock out parts of the joblib backend and test with that?

civis/parallel.py
evanderh: Good idea

civis/tests/test_parallel.py
evanderh: More or less yeah. Since the bug was in _CivisBackendResult, I was trying isolate it as closely as possible by forcing the future callback to be called with state=running and is_cancel_requested=True, then checking to make sure the joblib callback hadn't been called. I could also have mocked the joblib but this was the first test that came to mind. Do you think it'd be worth it to test _CivisBackend?

examples/parallel_cancel.py
evanderh: Whoops hadn't checked the diff yet, good catch

civis/tests/test_parallel.py
mheilman: I think this is probably fine as a test of the functionality, though a test using _CivisBackend would be easier to understand, I think.

If you want to keep this approach, I think it's a bit confusing to look at this test code right now because it's not obvious how the `cancel` call would affect the callback, so I think it'd benefit from a comment. How about something like this?

"When a future created by the Civis joblib backend completes normally, a callback function will be executed. If the future is cancelled, we don't want that callback to be run. Instantiating the _CivisBackendResult object emulates what the backend does internally."

evanderh: 👍 Good idea, added some explanation

PR #396, opened 2020-06-03
Show tables of methods for each set of endpoints in the API Resources

mheilman:
This also bumps the version for a new release.


mheilman:
good catch

PR #397, opened 2020-06-25
[APIC-300] bump minimum pubnub version

patr1ckm:
From the pubnub changelog, they fixed all the keyword clashes (for async and await) since the v4.1.2 released in Sept 2018. 

I don't fully understand all the reasons I ran into this issue (why an older pubnub version was pulled in on my fresh install), but it's quick change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 
+### Changed
+- Bump minimum pubnub version to `4.1.12` (#397)
+
 ## 1.14.2 - 2020-06-03
 ### Added
 - Added support for Python 3.8 (#391)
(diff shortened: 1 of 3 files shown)


mheilman:
Would you mind also updating the line in the [requirements.txt](https://github.com/civisanalytics/civis-python/blob/3fdceb90ec8ffc1bf60f0a5014b5d5cb8209f3c7/requirements.txt#L7)?

jacksonllee:
One more spot to change:

https://github.com/civisanalytics/civis-python/blob/3fdceb90ec8ffc1bf60f0a5014b5d5cb8209f3c7/requirements.txt#L7

mheilman:
@jacksonllee, do you think we should only allow squash and merge? It's possible to disable the other 2 options

jacksonllee:
> do you think we should only allow squash and merge?

Yeah, allowing "squash and merge" only for this repo seems reasonable. If we did come across edge cases where we'd need the other merge options, we could always re-enable them...

PR #400, opened 2020-08-14
[APIC-305] Require sql_type for every column when table_columns is provided

izzy-zelichenko:
[APIC-305](https://civisanalytics.atlassian.net/browse/APIC-305)

Ensure data types are detected when the destination table does not exist or it is being dropped and table_columns are provided but do not contain a sql_type for every column.

### DIFF
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1027,11 +1027,11 @@ def civis_file_to_table(file_id, database, table, client=None,
     except ValueError:
         table_exists = False
 
-    table_columns_valid = True
+    sql_types_provided = True
     if table_columns:
         sql_type_cnt = sum(1 for col in table_columns if col.get('sql_type'))
         if sql_type_cnt == 0:
-            table_columns_valid = False
+            sql_types_provided = False
         elif sql_type_cnt != len(table_columns):
             error_message = 'Some table columns ' \
                            'have a sql type provided, ' \
@@ -1042,7 +1042,7 @@ def civis_file_to_table(file_id, database, table, client=None,
     # and perform necessary file cleaning
     need_table_columns = ((not table_exists or existing_table_rows == 'drop')
                           and
-                          (table_columns is None or not table_columns_valid))
+                          (table_columns is None or not sql_types_provided))
 
     cleaning_futures = _run_cleaning(file_id, client, need_table_columns,
                                      headers, delimiter, hidden)
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -662,10 +662,13 @@ def dataframe_to_civis(df, database, table, api_key=None, client=None,
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
     table_columns : list[Dict[str, str]], optional
-        A list of dictionaries corresponding to the columns in
-        the source file. Each dictionary should have keys
-        for column "name" and "sqlType". The import will only copy these
-        columns regardless if there are more columns in the table.
+        An array of hashes corresponding to the columns in the order
+        they appear in the source file. Each hash should have keys for
+        database column "name" and "sqlType". This parameter is
+        required if the table does not exist, the table is being dropped,
+        or the columns in the source file do not appear in the same order
+        as in the destination table. The “sqlType” key is not required
+        when appending to an existing table.
     headers : bool, optional [DEPRECATED]
         Whether or not the first row of the file should be treated as
         headers. The default, ``None``, attempts to autodetect whether
@@ -805,10 +808,13 @@ def csv_to_civis(filename, database, table, api_key=None, client=None,
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
     table_columns : list[Dict[str, str]], optional
-        A list of dictionaries corresponding to the columns in
-        the source file. Each dictionary should have keys
-        for column "name" and "sqlType". The import will only copy these
-        columns regardless if there are more columns in the table.
+        An array of hashes corresponding to the columns in the order
+        they appear in the source file. Each hash should have keys for
+        database column "name" and "sqlType". This parameter is
+        required if the table does not exist, the table is being dropped,
+        or the columns in the source file do not appear in the same order
+        as in the destination table. The “sqlType” key is not required
+        when appending to an existing table.
     delimiter : string, optional
         The column delimiter. One of ``','``, ``'\\t'`` or ``'|'``.
     headers : bool, optional
@@ -934,10 +940,13 @@ def civis_file_to_table(file_id, database, table, client=None,
     sortkey2 : str, optional
         The second column in a compound sortkey for the table.
     table_columns : list[Dict[str, str]], optional
-        A list of dictionaries corresponding to the columns in
-        the source file. Each dictionary should have keys
-        for column "name" and "sqlType". The import will only copy these
-        columns regardless if there are more columns in the table.
+        An array of hashes corresponding to the columns in the order
+        they appear in the source file. Each hash should have keys for
+        database column "name" and "sqlType". This parameter is
+        required if the table does not exist, the table is being dropped,
+        or the columns in the source file do not appear in the same order
+        as in the destination table. The “sqlType” key is not required
+        when appending to an existing table.
     primary_keys: list[str], optional
         A list of the primary key column(s) of the destination table that
         uniquely identify a record. These columns must not contain null values.
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1026,13 +1026,17 @@ def civis_file_to_table(file_id, database, table, client=None,
         table_exists = True
     except ValueError:
         table_exists = False
-    
+
     table_columns_valid = True
     if table_columns:
         table_columns_valid = True if table_columns[0]['sql_type'] else False
-        for i in range (0, len(table_columns)-1):
-            if not table_columns[i]['sql_type'] == table_columns[i+1]['sql_type']:
-                raise ValueError('Some table columns do not have a sql type provided')
+        for i in range(0, len(table_columns)-1):
+            thisSqlType = table_columns[i]['sql_type']
+            nextSqlType = table_columns[i+1]['sql_type']
+            if not thisSqlType == nextSqlType:
+                error_message = 'Some table columns do not ' \
+                                'have a sql type provided'
+                raise ValueError(error_message)
                 break
 
     # Use Preprocess endpoint to get the table columns as needed
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1038,6 +1038,7 @@ def civis_file_to_table(file_id, database, table, client=None,
                                 'have a sql type provided, ' \
                                 'but others do not.'
                 raise ValueError(error_message)
+        print(str(table_columns_valid))
 
     # Use Preprocess endpoint to get the table columns as needed
     # and perform necessary file cleaning
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1031,19 +1031,18 @@ def civis_file_to_table(file_id, database, table, client=None,
     if table_columns:
         table_columns_valid = True if table_columns[0]['sql_type'] else False
         for i in range(0, len(table_columns)-1):
-            thisSqlType = table_columns[i]['sql_type']
-            nextSqlType = table_columns[i+1]['sql_type']
-            if not thisSqlType == nextSqlType:
-                error_message = 'Some table columns do not ' \
-                                'have a sql type provided'
+            this_sql_type = table_columns[i]['sql_type']
+            next_sql_type = table_columns[i+1]['sql_type']
+            if not this_sql_type == next_sql_type:
+                error_message = 'Some table columns ' \
+                                'have a sql type provided, ' \
+                                'but others do not.'
                 raise ValueError(error_message)
-                break
 
     # Use Preprocess endpoint to get the table columns as needed
     # and perform necessary file cleaning
-    need_table_columns = (((not table_exists or existing_table_rows == 'drop')
-                          and table_columns is None)
-                          or not table_columns_valid)
+    need_table_columns = ((not table_exists or existing_table_rows == 'drop')
+                          and (table_columns is None or not table_columns_valid))
 
     cleaning_futures = _run_cleaning(file_id, client, need_table_columns,
                                      headers, delimiter, hidden)
@@ -1053,7 +1052,7 @@ def civis_file_to_table(file_id, database, table, client=None,
         cleaning_futures, client, headers, need_table_columns, delimiter
     )
 
-    table_columns = table_columns or cleaned_table_columns
+    table_columns = cleaned_table_columns if need_table_columns else table_columns
 
     source = dict(file_ids=cleaned_file_ids)
     destination = dict(schema=schema, table=table, remote_host_id=db_id,
--- a/civis/io/_tables.py
+++ b/civis/io/_tables.py
@@ -1026,19 +1026,20 @@ def civis_file_to_table(file_id, database, table, client=None,
         table_exists = True
     except ValueError:
         table_exists = False
+    
+    table_columns_valid = True
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range (0, len(table_columns)-1):
+            if not table_columns[i]['sql_type'] == table_columns[i+1]['sql_type']:
+                raise ValueError('Some table columns do not have a sql type provided')
+                break
 
     # Use Preprocess endpoint to get the table columns as needed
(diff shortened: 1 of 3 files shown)


civis/io/_tables.py
```diff
@@ -1029,8 +1029,16 @@ def civis_file_to_table(file_id, database, table, client=None,
 
     # Use Preprocess endpoint to get the table columns as needed
     # and perform necessary file cleaning
-    need_table_columns = ((not table_exists or existing_table_rows == 'drop')
+    table_columns_have_sql = True
```
uttercm, line 1032: nit: this seems a bit wordy maybe something like `table_columns_valid`, idk, thoughts?

civis/io/_tables.py
```diff
@@ -1029,8 +1029,16 @@ def civis_file_to_table(file_id, database, table, client=None,
...
+    if table_columns:
+        for col in table_columns:
+            if not col['sql_type']:
+                table_columns_have_sql = False
+                break
+
+    need_table_columns = (((not table_exists or existing_table_rows == 'drop')
```
uttercm, line 1039: I'm now wondering why we shouldn't throw an error when the table_columns are provided but don't have a `sql_type`, seems to me we might want to throw an error rather than ignore the provided ones?

izzy-zelichenko: @uttercm ahh do you mean because it'll override all of them, even though some of them might be provide?

civis/io/_tables.py
izzy-zelichenko: Yeah, that seems better. I wanted to be specific, but also thought it sounded awkward as is.

civis/io/_tables.py
uttercm: ya exactly, so they might be confused if the sql_types they did provide are different than the actual types we create since not all were provided

izzy-zelichenko: That's a good point, but that would be a change in approach, so let me touch base with John on that.

jkulzick: This makes sense. Just to be clear if no sql types are provided we will detect them but if some but not all are provided we will raise an error.

izzy-zelichenko: Ah okay, I misunderstood the criteria in the first place. Thanks!

civis/tests/test_io.py
```diff
@@ -398,6 +398,147 @@ def test_civis_file_to_table_table_doesnt_exist(
             **expected_kwargs
         )
 
+    @pytest.mark.civis_file_to_table
+    @mock.patch('civis.io._tables._process_cleaning_results')
+    @mock.patch('civis.io._tables._run_cleaning')
+    def test_civis_file_to_table_table_exists_all_sql_types_missing(
```
jkulzick, line 404: I think we want `table_does_not_exist` for this test. We're testing that if all sql types are missing that we run cleaning? But if the table exists then we don't need to run cleaning so we probably want to change this to the table not existing or the `existing_table_rows = 'drop'`.

civis/tests/test_io.py
```diff
@@ -398,6 +398,147 @@ def test_civis_file_to_table_table_doesnt_exist(
...
+        database = 'redshift-general'
+        mock_file_id = 1234
+        mock_cleaned_file_id = 1235
+        mock_import_id = 8675309
+
+        self.mock_client.imports.post_files_csv.return_value\
+            .id = mock_import_id
+        self.mock_client.get_database_id.return_value = 42
+        self.mock_client.default_credential = 713
+
```
jkulzick, line 420: Here we want to mock the api client response for `self.mock_client.get_table_id` based on whether or not the table exists. If the table exists, it should look like [this](https://github.com/civisanalytics/civis-python/blob/6d632e4e71888bdc63aba7655442bdc715c5f04d/civis/tests/test_io.py#L224). If the table does not exist then it should look like [this](https://github.com/civisanalytics/civis-python/blob/6d632e4e71888bdc63aba7655442bdc715c5f04d/civis/tests/test_io.py#L322).

civis/tests/test_io.py
```diff
@@ -398,6 +398,147 @@ def test_civis_file_to_table_table_doesnt_exist(
...
+        table_columns = [{'name': 'a', 'sql_type': ''},
+                         {'name': 'b', 'sql_type': ''}]
+        m_process_cleaning_results.return_value = (
+            [mock_cleaned_file_id],
+            True,  # headers
+            'gzip',  # compression
+            'comma',  # delimiter
+            table_columns  # table_columns
```
jkulzick, line 428: In this test, the `table_columns` provided as the parameter (to `civis.io.civis_file_to_table`) should be different than what is returned by `m_process_cleaning_results` or our cleaning job. This is because that job is detecting sql types since sql types were not provided and the table does not exist or is being dropped. So I would expect this `table_columns` to return valid `sql_type` entries.

civis/tests/test_io.py
```diff
@@ -398,6 +398,147 @@ def test_civis_file_to_table_table_doesnt_exist(
...
+            'name': expected_name,
+            'max_errors': None,
+            'existing_table_rows': 'truncate',
+            'hidden': True,
+            'column_delimiter': 'comma',
+            'compression': 'gzip',
+            'escaped': False,
+            'execution': 'immediate',
+            'loosen_types': False,
+            'table_columns': table_columns,
```
jkulzick, line 480: here we should expect the csv import job to be called with the `table_columns` returned by cleaning, not with the `table_columns` provided as the parameter to `civis_file_to_table`.

civis/tests/test_io.py
```diff
@@ -398,6 +398,147 @@ def test_civis_file_to_table_table_doesnt_exist(
...
+        table_columns = [{'name': 'a', 'sql_type': 'INT'},
+                         {'name': 'b', 'sql_type': ''}]
+        m_process_cleaning_results.return_value = (
+            [mock_cleaned_file_id],
+            True,  # headers
+            'gzip',  # compression
+            'comma',  # delimiter
+            table_columns  # table_columns
+        )
+        m_run_cleaning.return_value = [mock.sentinel.cleaning_future]
```
jkulzick, line 530: Review what we need to mock based on what gets called before we raise the `ValueError`. For example, we shouldn't hit `post_files_csv` or `m_process_cleaning_results` yet so we shouldn't need to mock those.

We will check if the table exists so we should be mocking the api client (based on the test name the table exists). See [here](https://github.com/civisanalytics/civis-python/blob/6d632e4e71888bdc63aba7655442bdc715c5f04d/civis/tests/test_io.py#L224).

civis/tests/test_io.py
izzy-zelichenko: Ah - I misread the condition statement. I was so confused.

civis/io/_tables.py
```diff
@@ -1027,10 +1027,23 @@ def civis_file_to_table(file_id, database, table, client=None,
     except ValueError:
         table_exists = False
 
+    table_columns_valid = True
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range(0, len(table_columns)-1):
+            thisSqlType = table_columns[i]['sql_type']
+            nextSqlType = table_columns[i+1]['sql_type']
+            if not thisSqlType == nextSqlType:
```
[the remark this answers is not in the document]
izzy-zelichenko: Yeah this might be confusing to read, but actually putting the `not` before a string evaluates it as a boolean. So what I am checking here is that if the next sql type exists but the previous one does not, or vice versa, then we throw an error. Because that means that there is a mix between existing and non-existing sql types. If the error isn't thrown then the original assignment of True or False remains, the former means they are all provided and the latter means none are provided. I'm pretty sure this logic achieves what we wants and I like that it exits the loop as early as necessary. But let me use a test just to be sure!

civis/io/_tables.py
```diff
@@ -1027,10 +1027,23 @@ def civis_file_to_table(file_id, database, table, client=None,
     except ValueError:
         table_exists = False
 
+    table_columns_valid = True
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range(0, len(table_columns)-1):
+            thisSqlType = table_columns[i]['sql_type']
```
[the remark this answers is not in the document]
izzy-zelichenko: Yes I know this, it was just out of habit haha

civis/io/_tables.py
```diff
@@ -1027,10 +1027,23 @@ def civis_file_to_table(file_id, database, table, client=None,
...
+                                'have a sql type provided'
+                raise ValueError(error_message)
+                break
+
     # Use Preprocess endpoint to get the table columns as needed
     # and perform necessary file cleaning
-    need_table_columns = ((not table_exists or existing_table_rows == 'drop')
+    need_table_columns = (((not table_exists or existing_table_rows == 'drop')
                           and table_columns is None)
+                          or not table_columns_valid)
```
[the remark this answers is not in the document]
izzy-zelichenko: Yes I see now! I misunderstood the logic at first.

civis/tests/test_io.py
izzy-zelichenko: Yeah I see now, since I had the logic statement wrong I miswrote the tests

civis/tests/test_io.py
izzy-zelichenko: yeah I was wondering that! Looking at the existing tests it looked like they were passing in the same table_columns so I was confused about that, but I copied them. I'll take a closer look!

civis/tests/test_io.py
izzy-zelichenko: Yeah, same confusion as above. Is it me or are these tests really hard to read...?

civis/tests/test_io.py
izzy-zelichenko: Ah yes, good catch. I meant to remove these.

civis/io/_tables.py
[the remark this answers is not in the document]
jkulzick: I think in Python we can arrive at a much simpler approach that is easier to read. Python has a lot of powerful built-ins.

For example - `sql_type_cnt = sum(1 for col in table_columns if col.get('sql_type'))`

civis/io/_tables.py
```diff
@@ -1027,10 +1027,23 @@ def civis_file_to_table(file_id, database, table, client=None,
...
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range(0, len(table_columns)-1):
+            thisSqlType = table_columns[i]['sql_type']
+            nextSqlType = table_columns[i+1]['sql_type']
+            if not thisSqlType == nextSqlType:
+                error_message = 'Some table columns do not ' \
+                                'have a sql type provided'
+                raise ValueError(error_message)
+                break
```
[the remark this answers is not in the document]
izzy-zelichenko: done

civis/io/_tables.py
```diff
@@ -1027,10 +1027,23 @@ def civis_file_to_table(file_id, database, table, client=None,
...
         table_exists = False
 
+    table_columns_valid = True
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range(0, len(table_columns)-1):
+            thisSqlType = table_columns[i]['sql_type']
+            nextSqlType = table_columns[i+1]['sql_type']
+            if not thisSqlType == nextSqlType:
+                error_message = 'Some table columns do not ' \
```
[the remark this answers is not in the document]
izzy-zelichenko: done

civis/io/_tables.py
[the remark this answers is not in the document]
izzy-zelichenko: done

civis/io/_tables.py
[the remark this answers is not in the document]
izzy-zelichenko: done

civis/tests/test_io.py
izzy-zelichenko: done

civis/tests/test_io.py
izzy-zelichenko: done

civis/tests/test_io.py
izzy-zelichenko: done

civis/tests/test_io.py
izzy-zelichenko: done

civis/tests/test_io.py
izzy-zelichenko: done

civis/io/_tables.py
```diff
@@ -1040,7 +1052,7 @@ def civis_file_to_table(file_id, database, table, client=None,
         cleaning_futures, client, headers, need_table_columns, delimiter
     )
 
-    table_columns = table_columns or cleaned_table_columns
+    table_columns = cleaned_table_columns if need_table_columns else table_columns
```
izzy-zelichenko, line 1055: This logic was flawed, since we may need to clean even when table_columns are provided, so needed to be updated.

civis/tests/test_io.py
```diff
@@ -320,7 +320,8 @@ def test_civis_file_to_table_table_doesnt_exist(
         self.mock_client.default_credential = 713
 
         self.mock_client.get_table_id.side_effect = ValueError('no table')
-        mock_columns = [{'name': 'foo', 'sql_type': 'INTEGER'}]
+        mock_columns = [{'name': 'foo', 'sql_type': 'INTEGER'},
+                        {'name': 'bar', 'sql_type': 'VARCHAR(42)'}]
```
izzy-zelichenko, line 324: I added a second, different sql type to show that this test passes with current logic

civis/io/_tables.py
```diff
@@ -1027,10 +1027,22 @@ def civis_file_to_table(file_id, database, table, client=None,
     except ValueError:
         table_exists = False
 
+    table_columns_valid = True
+    if table_columns:
+        table_columns_valid = True if table_columns[0]['sql_type'] else False
+        for i in range(0, len(table_columns)-1):
+            this_sql_type = table_columns[i]['sql_type']
+            next_sql_type = table_columns[i+1]['sql_type']
+            if (not this_sql_type) == (next_sql_type):
```
izzy-zelichenko, line 1036: This was not actually working as expected, had to add parentheses to have them evaluate as needed.

civis/tests/test_io.py
```diff
@@ -398,6 +398,242 @@ def test_civis_file_to_table_table_doesnt_exist(
             **expected_kwargs
         )
 
+    @pytest.mark.civis_file_to_table
+    @mock.patch('civis.io._tables._process_cleaning_results')
+    @mock.patch('civis.io._tables._run_cleaning')
+    def test_civis_file_to_table_table_doesnt_exist_valid_table_columns_provided(
```
izzy-zelichenko, line 404: I added a test to show that when the table does not exist, but all sql_types are provided, that even if they are different, it does not throw an error and evaluates them correctly. No cleaning happens.

izzy-zelichenko:
We now have a test for every case when table_columns are provided

civis/io/_tables.py
uttercm: See John's comments about simplifying this approach, this takes a bit of effort to understand what exactly we are looking for. I like the idea of using the counts of how many `sql_type` are provided.

civis/io/_tables.py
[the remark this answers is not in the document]
izzy-zelichenko: Yeah that is simpler, I'll go with this approach.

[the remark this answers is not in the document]
izzy-zelichenko: Done. That was easy!
I haven't used Python in a really long time, so it's hard for me to pull out these tools.

civis/io/_tables.py
```diff
@@ -1027,10 +1027,22 @@ def civis_file_to_table(file_id, database, table, client=None,
     except ValueError:
         table_exists = False
 
+    table_columns_valid = True
```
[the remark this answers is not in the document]
izzy-zelichenko: got it, yeah agree!

civis/tests/test_io.py
izzy-zelichenko: Yeah that makes sense!

civis/io/_tables.py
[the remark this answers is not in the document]
izzy-zelichenko: done. hows this?

izzy-zelichenko:
I removed the test I previously added because I realized it was nearly identical to the test `test_civis_file_to_table_table_doesnt_exist_provide_table_columns`, so I just updated that existing test to have two table_columns provided to be more thorough.

mheilman:
@izzy-zelichenko, would you mind making a followup PR to add an item to the changelog for this?

izzy-zelichenko:
@mheilman Sure thing.

PR #402, opened 2020-08-28
[APIC-313] retain specific sql types in civis_file_to_table

mheilman:
In the `civis.io.civis_file_to_table` function, the Python API client appears to tell platform to change SQL types like `VARCHAR(1)` in the `table_columns` argument to more generic values like `VARCHAR(1024)` when there are multiple files.  The behavior seems reasonable when data types are automatically inferred (i.e., when `need_table_columns` is true), as noted in the code comment, but if the user has specified a dtype with specific length like `VARCHAR(1)`, the import should just use that.

### DIFF
--- a/civis/tests/test_io.py
+++ b/civis/tests/test_io.py
@@ -536,102 +536,120 @@ class ImportTests(CivisVCRTestCase):
     @mock.patch('civis.io._tables._process_cleaning_results')
     @mock.patch('civis.io._tables._run_cleaning')
     def test_civis_file_to_table_table_doesnt_exist_provide_table_columns(
-            self,
-            m_run_cleaning,
-            m_process_cleaning_results,
-            _m_get_api_spec
+        self,
+        m_run_cleaning,
+        m_process_cleaning_results,
+        _m_get_api_spec
     ):
-        table = "scratch.api_client_test_fixture"
-        database = 'redshift-general'
-        mock_file_id = 1234
-        mock_cleaned_file_id = 1235
-        mock_import_id = 8675309
 
-        self.mock_client.imports.post_files_csv.return_value\
-            .id = mock_import_id
-        self.mock_client.get_database_id.return_value = 42
-        self.mock_client.default_credential = 713
+        # pytest.parametrize apparently doesn't work with unittest.TestCase
+        # objects, which this ultimately inherits from, so we'll define a
+        # function here and run subtests below.
+        # See https://docs.pytest.org/en/stable/unittest.html.
+        def run_subtest(mock_file_ids):
+            table = "scratch.api_client_test_fixture"
+            database = 'redshift-general'
+            mock_cleaned_file_ids = mock.Mock()
+            mock_import_id = 8675309
 
-        self.mock_client.get_table_id.side_effect = ValueError('no table')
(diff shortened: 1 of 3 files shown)


civis/tests/test_io.py
```diff
@@ -536,102 +536,120 @@ def test_civis_file_to_table_table_does_not_exist_some_sql_types_missing(
...
-
-            run_job_future = mock.MagicMock(
-                spec=civis.futures.CivisFuture,
-                job_id=123,
-                run_id=234
+        # pytest.parametrize apparently doesn't work with unittest.TestCase
+        # objects, which this ultimately inherits from, so we'll define a
+        # function here and run subtests below.
+        # See https://docs.pytest.org/en/stable/unittest.html.
+        def run_subtest(mock_file_ids):
```
mheilman, line 549: I didn't change much here, but it involved extra indentation, so git detected a lot of changed lines.

civis/tests/test_io.py
```diff
@@ -536,102 +536,120 @@ def test_civis_file_to_table_table_does_not_exist_some_sql_types_missing(
...
+            table = "scratch.api_client_test_fixture"
+            database = 'redshift-general'
+            mock_cleaned_file_ids = mock.Mock()
```
mheilman, line 552: I changed this to just be a mock rather than an int, to avoid having to add logic about how long the list was, etc.

civis/tests/test_io.py
mheilman: I think there are 4 (2x2) situations related to this PR that should to be tested: (one file id, multiple file ids) x (table_columns provided, table_columns not provided).

Previously, this test tested the "one file id and table_columns provided" case.  I added the `for mock_file_ids in` loop below to be able to also test the "multiple file ids and table_columns provided" case without duplicating a lot of code.

The other two cases appear to be covered: "one file ID and no table_columns provided" is tested by `test_civis_file_to_table_table_doesnt_exist`, and the "multiple file IDs and no table_columns provided" case is tested by `test_civis_file_to_table_multi_file`, which is the only one of these four cases where we want `loosen_types` to be `True`.

mheilman:
> @mheilman looks like the build errored

🤦  sorry about that. Fixed now

civis/tests/test_io.py
```diff
@@ -536,102 +536,119 @@ def test_civis_file_to_table_table_does_not_exist_some_sql_types_missing(
...
-                'primary_keys': None,
-                'last_modified_keys': None
-            },
-            True,
-            **expected_kwargs
-        )
+        # Check that things work with a single file ID or multiple IDs.
+        # In particular, we want to ensure that loosen_types is set to False
+        # in both situations.
+        for mock_file_ids in (1234, [1234], [1234, 1235]):
```
mheilman, line 646: I'm more used to using `pytest.mark.parametrize` for this sort of thing, but that seems to not be possible since this is a method of a `unittest.TestCase`, so I tried to follow [this SO answer](https://stackoverflow.com/questions/32899/how-do-you-generate-dynamic-parameterized-unit-tests-in-python).  The `subTest` context manager isn't super necessary, but I think it might improve the error messages a bit when tests fail.

PR #404, opened 2020-09-08
[APIC-284] remove python 3.5 support since it is reaching EOL

mheilman:
'Python 3.5 will reach its "end of life" at the end of September 2020.' (https://www.python.org/downloads/release/python-3510/)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: xenial
 language: python
 python:
-    - "3.5"
     - "3.6"
     - "3.7"
     - "3.8"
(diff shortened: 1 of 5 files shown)


jacksonllee:
Everything looks good. Should we also throw in the `python_requires=">=3.6"` kwarg in the `setup` call in `setup.py`? [Reference](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)

mheilman:
Hmm... that seems a bit heavy-handed to me since we aren't aware of a particular problem with using the client with python 3.5. I don't feel strongly about it, though. 🤷

mheilman:
Yeah, that seems like a good idea. At first, I was concerned about it breaking somebody's job, but if they're using pypi, then pypi will give them the previous version because the classifiers will say the next version doesn't support python 3.5.

jacksonllee (approved):
LGTM

PR #406, opened 2020-09-09
[APIC-76] include services and other missing resources in sphinx docs

mheilman:
This adds some missing API resources to the list of resources that the sphinx doc generation code uses to create https://civis-python.readthedocs.io/en/stable/api_resources.html.  Using the list to filter the resources is a way to avoid including, e.g., "admin" endpoints that users aren't going to be interested in. Maybe we should eventually just include all the resources in the doc, but I didn't want to make that larger change at the moment.


salilgupta1 (approved):
lgtm

PR #407, opened 2020-09-09
[APIC-302] Update API spec so that new endpoints are included in docs

mheilman:
Updated the API spec at `civis/tests/civis_api_spec.json` so that new endpoints are included (e.g., `/exports/files/csv`)


mheilman:
I built the docs locally by doing `export READTHEDOCS=True` and `cd docs && make html`, and the /exports/files/csv docs showed up on the API Resources page, which [isn't the case now](https://civis-python.readthedocs.io/en/stable/api_resources.html#exports).

patr1ckm (approved):
Looks like the relevant data is in there now:
Screen Shot 2020-09-09 at 3 33 30 PM

LGTM

PR #410, opened 2020-09-11
[APIC-306] Fixed a bug in parsing responses that included "update" as a key

mheilman:
This fixes a bug where a column named "update" would cause response parsing to fail for `/tables` endpoints.

The `civis.response.Response` provides convenient access to response attributes (e.g., `response.attrib` as opposed to `response['attrib']`). In retrospect, perhaps it'd have been better not to have that functionality recur all the way down the response structure (or at least not into user-provided table info), but that ship has sailed, and in order to minimize breaking changes, this PR should preserve the existing functionality and avoid the bug.

### DIFF
--- a/civis/response.py
+++ b/civis/response.py
@@ -118,6 +118,9 @@ class Response(dict):
             self.calls_remaining = headers.get('X-RateLimit-Remaining')
             self.rate_limit = headers.get('X-RateLimit-Limit')
 
+        # Keys to update for this response object.
+        self_updates = {}
+
         if json_data is not None:
             for key, v in json_data.items():
                 if snake_case:
@@ -131,8 +134,11 @@ class Response(dict):
                 else:
                     val = v
 
-                self.update({key: val})
-                self.__dict__.update({key: val})
+                self_updates[key] = val
+
+        self.update(self_updates)
+        # Update self.__dict__ at the end to avoid replacing the update method.
+        self.__dict__.update(self_updates)
 
 
 class PaginatedResponse:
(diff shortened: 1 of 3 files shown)


civis/tests/test_response.py
```diff
@@ -174,3 +174,15 @@ def test_convert_data_type_civis_list():
     assert isinstance(data[0], Response)
     assert data[0]['foo'] == 'bar'
     assert data[0].headers == {'header': 'val'}
+
+
+def test_avoid_parsing_column_names():
```
mheilman, line 179: this test would fail before

civis/tests/test_response.py
```diff
@@ -174,3 +174,15 @@ def test_convert_data_type_civis_list():
     assert isinstance(data[0], Response)
     assert data[0]['foo'] == 'bar'
     assert data[0].headers == {'header': 'val'}
+
+
+def test_parse_column_names():
```
mheilman, line 179: this test would fail before

civis/response.py
```diff
@@ -131,8 +134,11 @@ def __init__(self, json_data, snake_case=True, headers=None):
                 else:
                     val = v
 
-                self.update({key: val})
-                self.__dict__.update({key: val})
+                self_updates[key] = val
+
+        self.update(self_updates)
+        # Update self.__dict__ at the end to avoid replacing the update method.
+        self.__dict__.update(self_updates)
```
salilgupta1, line 141: so the issue is that the `update` key was overwriting the `dictionary` `update` method?

mheilman: yeah, and then the next `update` call would fail.

salilgupta1 (approved):
LGTM

PR #413, opened 2020-10-20
[CIVP-22369] - bug fix retry on iterate

byndcivilization:
fixed retry bug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 ### Fixed
-- fixes bug whereby calls with iterate=True do not retry
+- fixes bug whereby calls with iterate=True do not retry (#413)
 
 ## 1.15.0 - 2020-09-29
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
+
+### Fixed
+- fixes bug whereby calls with iterate=True do not retry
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+### Changed
+
+### Fixed
+- fixes bug whereby calls with iterate=True do not retry
```
mheilman, line 9: could you please put `(#413)` at the end of this?

mheilman (changes requested):
Would you mind adding/revising a test so that the previous version would fail?

byndcivilization:
@mheilman how does this look?

civis/tests/test_utils.py
```diff
@@ -260,3 +288,75 @@ def test_no_retry_on_connection_error(mock_session):
...
+    api_response = {'key': 'value'}
+    session_context = mock_session.return_value.__enter__.return_value
+    session_context.send.return_value.json.return_value = api_response
+
+    session_context.send.return_value.status_code = 429
+    session_context.send.return_value.headers = {
+        'Retry-After': str(retry_after)
+    }
+
+    for verb in ['head', 'trace', 'get', 'put', 'options', 'delete', 'post']:
```
mheilman, line 343: This function is the same as the one above except for the name and this line, right? Could you just put the lowercase verb strings in the list in the other function ([here](https://github.com/civisanalytics/civis-python/pull/413/files#diff-f2df2ec5c411a1d0c536fcfb23aa66d7196fa463abd9d22460617bb37d667db2R307))?  `pytest.mark.parametrize` would be another option, but that seems like overkill.

mheilman (approved):
LGTM

PR #415, opened 2021-02-09
[CIVP-24324] FIX remove `_get_headers` in `read_civis_sql`

madopal:
This PR removes the helper function `_get_headers` used by `civis.io.read_civis_sql`.

Originally, @madopal discovered a bug in this helper function. John Kulzick then [commented](https://github.com/civisanalytics/civis-python/pull/415#issuecomment-776867399) that this helper function should in fact be removed based on the current state of the internal Civis Platform code.

For reference, the following is @madopal 's original PR description:

---

- As it stands now, if an error is tripped in _get_headers, it silently fails. However, a simple fail can be invoked if the sql being checked is terminated with a semicolon. So a strip of the semicolon is added to cover a few more cases & get the
optimization of that routine. Because the entire failure is labeled NOQA, tests aren't really possible. However, it was run with some queries to verify it doesn't blow up.


salilgupta1:
Sorry, I'm a bit confused. You want to remove semi-colons because it will throw the error or you want to remove the semi-colon because it will further suppress the error?

madopal:
So, it seems like the benefit of looking up the header can be done regardless of whether or not the query is terminated by a semicolon, no? 
If so, then we need to either 1) strip the semicolon when we get it and try to get the header or 2) document that if any SQL query is semicolon terminated, it will silently fail & skip the optimization. I had assumed the former, as if it fails it's no worse than how it is now, but if we have a reason for not gaining the benefit of the header when the query is semicolon-terminated, then it should at least be mentioned in the higher level call, right?

salilgupta1:
ok, I get what you are saying. I have no opinion since i'm not familiar with this code. @mheilman do you have any thoughts?

jkulzick:
The `_get_headers` code should just be removed. It was added at a time when SQL Scripts did not support doing a parallel unload if headers were requested. At that time, rather than fixing it on the API side, it was easier to add this code on the client side. The API side has since been addressed making the `_get_headers` code redundant.

I'm happy to provide context to things in the Python API client like this, but we do need to expand the bus factor and get other people involved in CR.

madopal:
Happy to just close this if that's the solution @jkulzick, unless you'd rather I modify it to remove _get_headers?

jkulzick:
> Happy to just close this if that's the solution @jkulzick, unless you'd rather I modify it to remove _get_headers?

@madopal yeah, we should modify this to remove _get_headers

civis/tests/test_io.py
```diff
@@ -972,35 +973,6 @@ def test_check_column_types_passing(self, _m_get_api_spec):
 
         civis.io._tables._check_column_types(table_cols, file_cols, 42)
 
-    @pytest.mark.read_civis
-    @pytest.mark.skipif(not has_pandas, reason="pandas not installed")
-    @mock.patch(api_import_str, return_value=civis_api_spec)
-    def test_read_civis_pandas(self, *mocks):
```
jacksonlee-civis, line 978: These cassette-based tests wouldn't pass. Since we've been gradually removing/replacing these cassette tests with mock-based ones anyway (e.g, #328, #425), I'm doing the same here with the `read_civis(_sql)` tests. (`read_civis` is a thin wrapper of `read_civis_sql`, and so below I'm explicitly testing `read_civis_sql` only.)

jacksonlee-civis:
@skomanduri Tagging you for review here, as we'd need someone with knowledge of the Platform code. In particular, is [Kulzick's comment](https://github.com/civisanalytics/civis-python/pull/415#issuecomment-776867399) still accurate at this point given the current state of Platform? Thank you.

skomanduri:
@jacksonlee-civis Yep, the comment is still accurate!  This logic has been moved into Platform so it can be removed from the client.

skomanduri (approved):
LGTM

PR #416, opened 2021-02-11
ContainerFuture propagates error from logs

elsander:
The Civis API currently returns `'error': None` when a Python/R/container script fails, so API client users don't get any information about what caused the failure. CivisML works around this by returning error logs to the user in these cases. This PR moves this behavior from `ModelPipeline` to `ContainerFuture` so that users can get more informative errors when running jobs outside of CivisML as well.

This is something of a workaround, especially since R client users won't get the benefits. Longer-term, we'd like to get more informative errors directly from the API.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 ### Added
-- `ContainerFuture` returns error logs as an error message when a job
-  fails but no message is returned (#416)
+- `ContainerFuture` propagates error messages from logs (#416)
 - Added EmptyResultError to `civis.io.read_civis` docs (#412)
 
 ## 1.15.1 - 2020-10-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 ### Added
+- `ContainerFuture` returns error logs as an error message when a job
+  fails but no message is returned (#413)
 - Added EmptyResultError to `civis.io.read_civis` docs (#412)
 
 ## 1.15.1 - 2020-10-28
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Added
 - `ContainerFuture` returns error logs as an error message when a job
-  fails but no message is returned (#413)
+  fails but no message is returned (#416)
 - Added EmptyResultError to `civis.io.read_civis` docs (#412)
 
 ## 1.15.1 - 2020-10-28
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -299,8 +299,6 @@ class ContainerFuture(CivisFuture):
             fut._exception_handled = True
 
         if fut.failed():
-            import ipdb
-            ipdb.set_trace()
             # Container scripts do not return the error message,
             # so we override with an exception we can pull from the
             # logs
@@ -318,8 +316,6 @@ class ContainerFuture(CivisFuture):
         - MemoryError
         """
 
-        import ipdb
-        ipdb.set_trace()
         logs = self.client.scripts.list_containers_runs_logs(self.job_id,
                                                              self.run_id,
                                                              limit=nlog)
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -11,7 +11,7 @@ import threading
 import warnings
 
 from civis import APIClient
-from civis.base import CivisAPIError, DONE
+from civis.base import CivisAPIError, CivisJobFailure, DONE
 from civis.polling import PollableResult
 
 from pubnub.pubnub import PubNub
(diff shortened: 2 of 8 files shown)


civis/tests/test_parallel.py
```diff
@@ -376,8 +376,9 @@ def test_result_success(mock_civis):
     # Test that we can get a result back from a succeeded job.
     callback = mock.MagicMock()
     mock_civis.io.civis_to_file.side_effect = make_to_file_mock('spam')
-    fut = ContainerFuture(1, 2, client=mock.MagicMock())
-    fut.set_result(Response({'state': 'success'}))
+    mock_client = create_client_container_mock(1, 2, state='success',
```
elsander, line 379: `ContainerFuture` now has a callback during init, so we can't get away with initializing and setting a result directly during testing. I got around this by moving the container-focused client mock from CivisML to the general tests and using it when initializing the future.

civis/tests/test_parallel.py
```diff
@@ -416,17 +418,16 @@ def test_result_exception_no_result():
...
-    # patching the client tests that the _CivisBackendResult
-    # uses the client object on the input CivisFuture.
-    mock_client = create_client_mock()
-    mock_client.scripts.list_containers_runs_outputs.return_value = []
+    mock_client = create_client_container_mock(1, 2, state='failed',
+                                               run_outputs=[])
     fut = ContainerFuture(1, 2, client=mock_client)
-    fut._set_api_exception(Response({'state': 'failed'}))
     res = civis.parallel._CivisBackendResult(fut, callback)
+    fut._set_api_exception(CivisJobFailure(
```
elsander, line 425: h/t @stephen-hoover for noticing that the arguments to `_set_api_exception` were not actually the expected order/type here (although the test happened to pass anyway).

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 ### Added
+- `ContainerFuture` returns error logs as an error message when a job
```
mheilman, line 7: I'm a bit confused by this changelog message. It seems like it's describing the shortcoming this PR is intended to address rather than the change. Is that correct? I think it'd be better to use something along the lines of the PR title ("ContainerFuture propagates error from logs") for this changelog entry.

civis/tests/test_futures.py
```diff
@@ -534,3 +535,23 @@ def test_future_retry_error():
...
+            {'id': 222,
+             'created_at': 'def',
+             'message': msg,
+             'level': 'error'}]
+    mock_client = create_client_container_mock(1, 2, state='failed',
+                                               run_outputs=[],
+                                               log_outputs=logs)
+    fut = ContainerFuture(1, 2, client=mock_client)
+
+    assert msg in str(fut._exception.error_message)
```
mheilman, line 557: I feel like it'd be better to test exactly what `error_message` is here rather than using `in`.  I don't feel strongly about it, though.

civis/tests/mocks.py
```diff
@@ -37,6 +38,58 @@ def create_client_mock(cache=TEST_SPEC):
     return mock_client
 
 
+def create_client_container_mock(script_id=-10, run_id=100, state='succeeded',
```
mheilman, line 41: I feel like something like `create_client_mock_for_container_tests` might be a better name here. With this name, it wasn't clear to me that this would return a mocked API client object.  I don't feel strongly about this.

civis/futures.py
```diff
@@ -286,6 +286,53 @@ def __init__(self, job_id, run_id,
...
+        This function presently recognizes the following errors:
+
+        - MemoryError
+        """
+
+        logs = self.client.scripts.list_containers_runs_logs(self.job_id,
+                                                             self.run_id,
+                                                             limit=nlog)
+
+        # Check for memory errors
```
mheilman, line 323: I think there should be a test of this memory error flagging if this function is going to specialize the error type.

I'd also be OK with taking out the memory error logic and always returning the generic `CivisJobFailure`.

mheilman (changes requested):

civis/tests/mocks.py
elsander: It's a little longer, but I think it is clearer. I'll update.

civis/futures.py
```diff
@@ -286,6 +286,53 @@ def __init__(self, job_id, run_id,
                          client=client,
                          poll_on_creation=poll_on_creation)
         self._max_n_retries = max_n_retries
+        self._exception_handled = False
+        self.add_done_callback(self._set_job_exception)
+
+    @staticmethod
+    def _set_job_exception(fut):
```
stephen-hoover, line 293: It's worth documenting that this method is intended for use as a callback. You could add a docstring with that mention.

civis/tests/mocks.py
```diff
@@ -37,6 +38,58 @@ def create_client_mock(cache=TEST_SPEC):
...
+        List of Response objects returned as run outputs
+
+    Returns
+    -------
+    `unittest.mock.Mock`
+        With scripts endpoints `post_containers`, `post_containers_runs`,
+        `post_cancel`, and `get_containers_runs` set up.
+    """
+    c = mock.Mock()
+    c.__class__ = APIClient
```
stephen-hoover, line 64: How about starting with the auto-specced client mock? The autospecs can be very helpful with testing. There's also a couple of places where the return of this function is now replacing `create_client_mock()`, so those tests would lose their autospecs.
```suggestion
    c = create_client_mock()
```

civis/tests/mocks.py
stephen-hoover: I'm a bit concerned that this new function is too special-purpose to go here. Maybe it should be marked private, and kept for inside this library only? Or else the docstring should get more detail about the structure of the returned mock. As it is, I don't think the function is useful in someone else's tests without them first reading the source code.

stephen-hoover (changes requested):
Most of my comments about this change are about the `ModelFuture`. Let me know if it's not clear where in the code some of my comments are pointing.

civis/tests/mocks.py
elsander: It's true that it is a more specialized mock, but we need some of these outputs for any testing of a ContainerFuture, so I'm using it across a few test modules. I can add some detail about the structure though, that's a good idea.

civis/ml/_model.py
```diff
@@ -527,7 +502,7 @@ def _set_model_exception(fut):
             # If there's no metadata file
```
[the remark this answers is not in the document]
ggarcia-civis: Hi Stephen, I updated the name and took out the repeated callback attributes in the init. It seems to be working properly, but I had to move some of the other attributes being set in the init to above the `super().__init__()` since the `_set_job_exception` method relies on some of those properties being set.

civis/futures.py
ggarcia-civis: I added a test for the memory error.

ggarcia-civis:
@stephen-hoover @mheilman Sorry about the long wait period. I tried to address all of the remaining feedback.

stephen-hoover:
@ggarcia-civis , I'm unlikely to be able to re-review in a timely fashion, so I will leave this review to @mheilman .

civis/ml/tests/test_model.py
```diff
@@ -379,13 +379,16 @@ def test_set_model_exception_unknown_error():
     fut = _model.ModelFuture(1, 2, client=mock_client)
     with pytest.raises(CivisJobFailure) as err:
         fut.result()
+    print(str(err))
```
mheilman, line 382: Is this left over from debugging?

mheilman (changes requested):
It looks like there are some leftover print statements from debugging, but this seems good otherwise.

civis/ml/tests/test_model.py
ggarcia-civis: yup 😅 thanks

ggarcia-civis:
Hi @elliothevel, Mike is on PTO this week. Can you pick up on this PR review?

elliothevel (approved):
LGTM

PR #417, opened 2021-02-11
[HACK-465] Use default arguments from swagger when provided

ggarcia-civis:
Some of our users have found it misleading that default arguments are always set to 'DEFAULT'. Also some of our endpoints do use default arguments and it would be nice to be able to include that in the signature if specified in the swagger.

The 'default' property is included as swagger 2.0 format as optional (https://swagger.io/docs/specification/2-0/describing-parameters/), but it was not being read-in as part of the civis-python parsing process.

This PR uses that default value if present in the swagger, if not it uses the previous 'DEFAULT' string because not setting a default for kwargs will cause the signature to read it as required.

here is an example with our sdkclient.weights.list() method which takes some optional arguments to filter by.
Before (all set to 'DEFAULT'):
Screen Shot 2021-02-11 at 1 55 14 PM

After (real default values where specified):
Screen Shot 2021-02-11 at 1 40 03 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 ### Added
 - Added EmptyResultError to `civis.io.read_civis` docs (#412)
+- Added default values from swagger in client method's signature (#416)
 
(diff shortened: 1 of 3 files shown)


mheilman (approved):
LGTM

PR #419, opened 2021-03-02
BUG Correct sql_type doc casing and check input for typos

stephen-hoover:
The client is looking for the `table_columns` parameter to be a list of dictionaries, each of which has a "name" key and which optionally also have a "sql_type" key. If Civis Platform needs to know types, and there is no "sql_type" key in any of the dictionaries, the client will silently replace a user-provided `table_columns` with an inferred `table_columns`. The docstring claims that the expected formatting is "sqlType", but "sqlType" keys will be ignored.

Fix the docstrings to use snake case instead of camel case, and add a check on the `table_columns` input to prevent mistakes or typos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Added EmptyResultError to `civis.io.read_civis` docs (#412)
 - Added default values from swagger in client method's signature (#416)
 
+### Fixed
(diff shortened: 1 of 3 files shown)


pbuggana (commented):
@stephen-hoover it's gonna take some work for me to be able to test L535-570. Can you confirm that the test case passes on your side?
I have tested L1025-1041 with both valid and invalid cases and it passes.
Updated docstrings also look good.

stephen-hoover:
Thank you for taking a look, @pbuggana . I can confirm that the new test case passes. It runs in Travis with the other tests, and the build is green.

pbuggana (approved):
L G T M!

PR #425, opened 2021-10-29
[CIVP-24658] REF Remove PubNub

jacksonlee-civis:
This PR removes all PubNub-related code, as we've stopped using PubNub for quite a while. There's no effect to the users of civis-python, because the code being removed has not been triggered since PubNub usage was dropped from Civis Platform.


civis/futures.py
```diff
@@ -14,52 +13,17 @@
...
-from pubnub.enums import PNStatusCategory
-from pubnub.callbacks import SubscribeCallback
 
 log = logging.getLogger(__name__)
 
-# Pubnub connections can recover missed messages upon reconnecting for up to 10
-# minutes from the disconnect. Polling on a 9.5 minute interval is used as a
+# Polling on a 9.5 minute interval is used as a
 # fallback in case the job complete message is missed in an outage.
 _LONG_POLLING_INTERVAL = 9.5 * 60
```
skomanduri, line 21: I think these lines should be removed as well.

skomanduri (changes requested):
Looks great!  Thank you for tackling this and tracking down so much code.

I still see some references to pubnub/channels in cassettes and in `MANIFEST.in`.  Are you able to remove those lines/files as well?  Since you've removed the test code, I think you should be able to remove all the cassettes that are calling `https://api.civisanalytics.com/channels`.

jacksonlee-civis:
@skomanduri Thanks for catching more PubNub/Channels remnants! I've just removed those as well and the CI builds have passed. This PR is ready for another review.

skomanduri (approved):
LGTM

🎉  🎉  🎉

PR #426, opened 2021-11-01
[CIVP-19775] ENH add job and run ID to CivisJobFailure from CivisFuture

jacksonlee-civis:
This PR adds the job ID and run ID to the exception message of `CivisJobFailure` coming from a `CivisFuture` object.

### DIFF
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -118,9 +118,8 @@ class CivisFuture(PollableResult):
 
         - MemoryError
         """
-        logs = self.client.scripts.list_containers_runs_logs(self.job_id,
-                                                             self.run_id,
-                                                             limit=nlog)
+        logs = self.client.jobs.list_runs_logs(
+            self.job_id, self.run_id, limit=nlog)
 
         # Check for memory errors
         msgs = [x['message'] for x in logs if x['level'] == 'error']
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -105,7 +105,7 @@ class CivisFuture(PollableResult):
             # so we override with an exception we can pull from the
             # logs
             if isinstance(fut._exception, CivisJobFailure) and \
-               fut._exception.error_message == 'None':
+               fut._exception._original_err_msg in ('None', "", None):
                 fut._exception.error_message = ''
                 exc = fut._exception_from_logs(fut._exception)
                 fut.set_exception(exc)
@@ -128,16 +128,12 @@ class CivisFuture(PollableResult):
             exc = MemoryError(mem_err[0])
         else:
             # Unknown error; return logs to the user as a sort of traceback
-            job_run_ids_in_exc = _format_job_run_ids_in_exception(
-                self.job_id, self.run_id)
-            all_logs = (
-                f"{job_run_ids_in_exc} "
-                + '\n'.join([x['message'] for x in logs])
-            )
+            all_logs = '\n'.join([x['message'] for x in logs])
             if isinstance(exc, CivisJobFailure):
-                exc.error_message = all_logs + '\n' + exc.error_message
+                exc._update_error_message(all_logs + '\n' + exc.error_message)
             else:
-                exc = CivisJobFailure(all_logs)
+                exc = CivisJobFailure(
+                    all_logs, job_id=self.job_id, run_id=self.run_id)
         return exc
 
     @property
@@ -296,10 +292,6 @@ class ContainerFuture(CivisFuture):
                 self._invoke_callbacks()
                 return self.cancelled()
             return False
-
-
-def _format_job_run_ids_in_exception(job_id: int, run_id: int) -> str:
-    return f"(Job ID {job_id} / Run ID {run_id})"
 
 
 def _create_docker_command(*args, **kwargs):
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -234,12 +234,14 @@ class ContainerFuture(CivisFuture):
                  poll_on_creation=True):
         if client is None:
             client = APIClient()
+
+        self._max_n_retries = max_n_retries
+
         super().__init__(client.scripts.get_containers_runs,
                          [int(job_id), int(run_id)],
                          polling_interval=polling_interval,
                          client=client,
                          poll_on_creation=poll_on_creation)
-        self._max_n_retries = max_n_retries
 
     def _set_api_exception(self, exc, result=None):
         # Catch attempts to set an exception. If there's retries
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -131,19 +131,19 @@ class CivisFuture(PollableResult):
         mem_err = [m for m in msgs if m.startswith('Process ran out of its')]
         if mem_err:
             err_msg = _err_msg_with_job_run_ids(
-                MemoryError, mem_err[0], self.job_id, self.run_id)
+                mem_err[0], self.job_id, self.run_id)
             exc = MemoryError(err_msg)
         else:
             # Unknown error; return logs to the user as a sort of traceback
             all_logs = '\n'.join([x['message'] for x in logs])
             if isinstance(exc, CivisJobFailure):
                 err_msg = _err_msg_with_job_run_ids(
-                    CivisJobFailure, all_logs + '\n' + exc.error_message,
+                    all_logs + '\n' + exc.error_message,
                     self.job_id, self.run_id)
                 exc.error_message = err_msg
             else:
                 err_msg = _err_msg_with_job_run_ids(
-                    exc.__class__, all_logs, self.job_id, self.run_id)
+                    all_logs, self.job_id, self.run_id)
                 exc = CivisJobFailure(
                     "", job_id=self.job_id, run_id=self.run_id)
                 exc.error_message = err_msg
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -101,7 +101,7 @@ class CivisFuture(PollableResult):
             fut._exception_handled = True
 
         if fut.failed():
-            # Container scripts do not return the error message,
+            # Some platform script types do not return the error message,
             # so we override with an exception we can pull from the
             # logs
             if isinstance(fut._exception, CivisJobFailure) and \
--- a/civis/futures.py
+++ b/civis/futures.py
@@ -84,6 +84,63 @@ class CivisFuture(PollableResult):
                          client=client,
                          poll_on_creation=poll_on_creation)
 
+        self._exception_handled = False
+        self.add_done_callback(self._set_job_exception)
+
+    @staticmethod
+    def _set_job_exception(fut):
+        """Callback: On job completion, check the status.
+        If the job has failed, and has no pre-existing error message,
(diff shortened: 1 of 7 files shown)


civis/futures.py
```diff
@@ -228,7 +228,12 @@ def _exception_from_logs(self, exc, nlog=15):
             exc = MemoryError(mem_err[0])
         else:
             # Unknown error; return logs to the user as a sort of traceback
-            all_logs = '\n'.join([x['message'] for x in logs])
+            job_run_ids_in_exc = _format_job_run_ids_in_exception(
+                self.job_id, self.run_id)
+            all_logs = (
+                f"{job_run_ids_in_exc} "
+                + '\n'.join([x['message'] for x in logs])
+            )
```
jacksonlee-civis, line 236: These changes as-is are in `ContainerFuture` instead of the parent class `CivisFuture`. Because it seems to be much more common for users to write code like `fut = CivisFuture(...)` rather than using `ContainerFuture`, I'm wondering whether it's worth doing the following:

1. Move these changes from `ContainerFuture` to `CivisFuture`
2. Move the changes from #416 from `ContainerFuture` to `CivisFuture`

Hi @ggarcia-civis , I have questions for you:

* What do you think about 1 and 2 here? Are both or one of them worth the extra work? Or are there actually potential blockers? Speaking of blockers, my next question below...
* Do you recall why #416 was done in `ContainerFuture` but not `CivisFuture`?

ggarcia-civis: On the ticket Eli mentioned doing it on `ContainerFuture` instead of `CivisFuture` because there were some instances of `CivisFuture` that already had error handling and not wanting to mess with that.

civis/futures.py
```diff
@@ -84,6 +84,62 @@ def __init__(self, poller, poller_args,
...
+    def _exception_from_logs(self, exc, nlog=15):
+        """Create an exception if the log has a recognizable error
+
+        Search "error" emits in the last ``n_log`` lines.
+        This function presently recognizes the following errors:
+
+        - MemoryError
+        """
+        logs = self.client.jobs.list_runs_logs(
+            self.job_id, self.run_id, limit=nlog)
```
jacksonlee-civis, line 122: Switched from `clients.scripts.list_containers_runs_logs` to `clients.jobs.list_runs_logs` as this is not specific to `ContainerFuture` anymore. Also the relevant changes in tests.

**Question: Is there anything else I should also modify for this `ContainerFuture` -> `CivisFuture` migration?**

civis/futures.py
```diff
@@ -84,6 +84,62 @@ def __init__(self, poller, poller_args,
...
+        if mem_err:
+            exc = MemoryError(mem_err[0])
+        else:
+            # Unknown error; return logs to the user as a sort of traceback
+            job_run_ids_in_exc = _format_job_run_ids_in_exception(
+                self.job_id, self.run_id)
+            all_logs = (
+                f"{job_run_ids_in_exc} "
+                + '\n'.join([x['message'] for x in logs])
+            )
```
jacksonlee-civis, line 136: This is the original main mission of this PR: add the job ID and run ID to the error message.

civis/futures.py
```diff
@@ -110,14 +166,6 @@ def _check_message(self, message):
...
         return match
 
-    def _poll_and_set_api_result(self):
-        with self._condition:
-            try:
-                result = self.poller(*self.poller_args)
-                self._set_api_result(result)
-            except Exception as e:
-                self._set_api_exception(exc=e)
-
```
jacksonlee-civis, line 120: Some `CivisFuture` clean-up that should've happened in #425. This `_poll_and_set_api_result` method is no longer used by anything since I've removed all the PubNub code.

civis/ml/tests/test_model.py
```diff
@@ -67,53 +68,6 @@
 }
 
 
-def setup_client_mock(script_id=-10, run_id=100, state='succeeded',
-                      run_outputs=None):
```
jacksonlee-civis, line 71: Removing this mock client creator in favor of the newer one called `create_client_mock_for_container_tests` from #416

civis/futures.py
```diff
@@ -299,6 +298,10 @@ def cancel(self):
             return False
 
 
+def _format_job_run_ids_in_exception(job_id: int, run_id: int) -> str:
+    return f"(Job ID {job_id} / Run ID {run_id})"
```
ggarcia-civis, line 302: based on the ticket description, I'm not sure if the job id and run id were meant to be added to the error description or as an attribute to the `CivisJobFailure` class. On one had I think it would be easy to reference if stored in the error class. ie.
```
try:
...
except CivisJobFailure as exc:
  log.error(f'Job {exc.job_id} failed please re-run')
```
But that information should be stored and accessible on the future already, so maybe I don't understand the use case.

I still think this formatting might be better to have on the error class. Maybe format the `__str__` to prefix the error with `Job #{job_id} / Run #{run_id} failed. {self.error_message}`. What do you think?

civis/futures.py
```diff
@@ -84,6 +84,62 @@ def __init__(self, poller, poller_args,
...
+            return
+        else:
+            fut._exception_handled = True
+
+        if fut.failed():
+            # Container scripts do not return the error message,
+            # so we override with an exception we can pull from the
+            # logs
+            if isinstance(fut._exception, CivisJobFailure) and \
+               fut._exception.error_message == 'None':
```
ggarcia-civis, line 108: I'm not sure how likely it is, but maybe since this is covering a wider range of objects this check can be done on `'None'`, `None`, and `''`.
This could also be added as a property on the `CivisJobFailure` class since we know it is a job failure at this point. Something like `fut._exception.has_empty_error_message`

civis/futures.py
ggarcia-civis: it would also be nice to account for the case when run_id is None.

civis/tests/test_futures.py
```diff
@@ -33,6 +34,14 @@ def clear_lru_cache():
     generate_classes.cache_clear()
 
 
+def _create_poller_mock(state: str) -> mock.Mock:
+    poller = mock.Mock()
+    api_result = mock.Mock()
+    api_result.state = state
```
ggarcia-civis, line 40: not necessary but I think you can cut down on lines like this:
```
api_result = mock.Mock(state=state)
poller = mock.Mock(return_value=api_result)
```

civis/tests/test_futures.py
```diff
@@ -86,40 +95,31 @@ def test_check_message_when_job_is_running(self, *mocks):
...
-        poller = mock.Mock()
-        api_result = mock.Mock()
-        api_result.state = 'succeeded'
+    def test_set_api_result_succeeded(self, mock_api):
+        poller = _create_poller_mock("succeeded")
 
         result = CivisFuture(poller, (1, 2))
-        result._set_api_result(api_result)
-        assert poller.call_count == 0
+        assert poller.call_count == 1
```
ggarcia-civis, line 102: I don't understand why this call count changed?

ggarcia-civis (changes requested):
A few comments from me, but in general I think this looks good. Thanks for revisiting this work!

I don't see any obvious issues with moving this error message handling to the `CivisFuture` class, especially with all the checks taking place, but it may still be a good idea to test it out with a few different platform objects.

civis/futures.py
jacksonlee-civis: Good question re: job/run id as attributes of `CivisJobFailure` versus just part of the error message string. While power users would know the future already has the job/run id (like you've mentioned) and would therefore know what to do, less advanced users might want/hope to see an unhandled CivisJobFailure come with the job/run id in the logging to aid debugging. Which is why I've been leaning towards fiddling with the error msg string with the job/run id included in some way.

I like your suggestions. I'm going to make the following updates:
* Add job/run id as attributes to `CivisJobFailure`. No harm having these even though the future already does. Also, it'd be handy to have these attributes for the next bullet point...
* Move the job/run id string formatting code to under `CivisJobfailure`. I'm considering updating `self.error_message` directly so that it always has the job/run id embedded, which would mean `__str__` is also taken care of.
* If run_id is None, don't show the unwieldy `/ run ID None` bit in the error msg.

civis/tests/test_futures.py
jacksonlee-civis: I dig it -- I'll update this!

civis/tests/test_futures.py
jacksonlee-civis: I _think_ it's because now `CivisFuture` has the `self._set_job_exception` callback in `__init__`, and that's what calls the poller. I'll double check.

jacksonlee-civis:
Just switched this PR back to the "draft" status for now as I'll need to check a couple more things. Will tag for review when it's ready again

civis/futures.py
ggarcia-civis: thanks for the explanation, whatever you think is best! I'm also okay with the original way if it avoids too much complication with the error messages

civis/polling.py
```diff
@@ -173,8 +173,12 @@ def _set_api_result(self, result):
                     err_msg = str(result['error'])
                 except:  # NOQA
                     err_msg = str(result)
-                self._set_api_exception(exc=CivisJobFailure(err_msg, result),
-                                        result=result)
+                job_id = getattr(self, "job_id", None)
+                run_id = getattr(self, "run_id", None)
```
jacksonlee-civis, line 177: `CivisFuture` subclasses `PollableResult`, but only a `CivisFuture` object has the `job_id` and `run_id` properties (which I moved from `ContainerFuture` "up" to `CivisFuture` a long time ago, #290). Though I don't know when realistically we'd have a `PollableResult` object without job/run ID, I'm happy with the `getattr` treatment here for now.

civis/tests/test_futures.py
```diff
@@ -86,40 +92,41 @@ def test_check_message_when_job_is_running(self, *mocks):
...
+        CivisFuture(poller, (1, 2), poll_on_creation=True)
+        assert poller.call_count == 1
 
-        result = CivisFuture(poller, (1, 2))
-        result._set_api_result(api_result)
+    @mock.patch(api_import_str, return_value=civis_api_spec_base)
+    def test_poller_call_count_poll_on_creation_false(self, mock_api):
+        poller = _create_poller_mock("succeeded")
+        CivisFuture(poller, (1, 2), poll_on_creation=False)
         assert poller.call_count == 0
```
jacksonlee-civis, line 104: re: poller count, I've verified that `poll_on_creation=True` (= current default value) does make the poller called, and I've written both True vs False cases here to be explicit.

jacksonlee-civis:
@ggarcia-civis This PR is ready for review again. I've tested this PR branch for triggering a `CivisJobFailure` from a future object coming from different Platform scripts: Python, SQL, Container, and Custom scripts -- the logs look as expected and desired. The test job links are in the linked JIRA ticket (not posting them in a public repo here).

ggarcia-civis (approved):
LGTM!
Thanks for all the refactoring and sharing the links to all the testing you did.

PR #428, opened 2021-11-04
[CIVP-22464] BUG Better error message when CSV preprocessing fails in `civis_file_to_table`

jacksonlee-civis:
`civis.io.civis_file_to_table` assumes the internal CSV preprocessing jobs always return run outputs, but this assumption breaks down if CSV preprocessing fails with no run outputs and therefore leads to an uninformative "list index out of bounds" error. This PR handles this error and provides a better error message to aid debugging.


kurtenbachkyle (approved):
lgtm

sorry for the delay!

PR #429, opened 2021-11-05
[CIVIS-1508] DEP upgrade joblib dep

### LINKED ISSUE #423: Support Joblib 1.x
At this time, joblib is pinned at 0.14.x. Joblib is now on 1.1.x and has released several improvements to its memory caching which are key to ensuring threading works as expected. https://github.com/joblib/joblib/blob/master/CHANGES.rst.  

I would like to see this officially supported to avoid python version conflicts.

dleatherman-civis:
# Features

This PR upgrades the joblib dependency to allow versions 1.1.x which provides several enhancements and bug fixes around pickling which impacted previous runs.

Closes #423.


requirements.txt
```diff
@@ -3,6 +3,7 @@ click>=6.0,<=7.99
 jsonref>=0.1,<=0.2.99
 requests>=2.12.0,==2.*
 jsonschema>=2.5.1,<=3.99
-joblib>=0.11,<0.15
+pubnub>=4.1.12,<=4.99
```
jacksonlee-civis, line 6: Remove this line?

requirements.txt
```diff
@@ -3,6 +3,7 @@ click>=6.0,<=7.99
...
+joblib>=0.11,<=1.1.*
```
jacksonlee-civis, line 7: Instead of `<=1.1.*`, what do you think about `<2`? If joblib follows semantic versioning (it does, correct?), then anything below version 2 should be fine, and `<2` would mean we don't have to make a PR here every time joblib releases a new `1.*`.

setup.py
```diff
@@ -55,7 +55,8 @@ def main():
             'jsonref>=0.1.0,<=0.2.99',
             'requests>=2.12.0,==2.*',
             'jsonschema>=2.5.1,<=3.99',
-            'joblib>=0.11,<0.15',
+            'joblib>=0.11,<=1.1.*',
+            'pubnub>=4.1.12,<=4.99',
```
jacksonlee-civis, line 59: Remove this line?

requirements.txt
dleatherman-civis: That works for me. I would be okay with that.

dleatherman-civis:
> Thank you for the contribution! I have a high-level question:
> 
> * Assuming joblib follows semantic versioning, the jump from joblib version `0.*` to `1.*` could mean breaking changes from joblib. If there are any, would these changes affect civis-python in any way?
> 
> Also, it looks like you might have started this work before some of the most recent PRs got merged into master. Some of the no-longer-valid changes show up in your PR here and should be dropped.
> 
> I've taken the liberty to update this PR's title to link this work to the relevant JIRA ticket, and added `Closes #423` to the PR description so that GitHub will close the issue once this PR is merged.

The breaking change is identified [here](https://github.com/joblib/joblib/pull/1136). with joblib <1.x, pickling joblib.Memory objects was backwards compatible. Newer versions of numpy changed how numpy.dtypes was pickled which caused objects that should be the same to create different hashes. Essentially, if you're loading in joblib.Memory pickle objects, they are no longer compatible across versions (hence the major version bump).

jacksonlee-civis (approved):
Thank you for tracking down joblib's breaking changes. Looks like joblib 1.* (and numpy >= 1.20) have been out for about a year, which shouldn't cause issues for us.

LGTM!

PR #430, opened 2021-11-09
[CIVIS-1609] FIX no `assert` statements in non-test code

jacksonlee-civis:
`assert` statements shouldn't be used in the actual, non-test code, since they can be disabled by the `-O` flag: https://docs.python.org/3/using/cmdline.html#cmdoption-O It's not clear how likely folks would run Civis-related code with this flag on, but it's better to convert these `assert` statements into conditional statements with exception raising.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - Ability to use joblib 1.1.x (#423)
 
 ### Fixed
+- Converted `assert` statements in non-test code into proper error handling (#430)
 - Handled the index-out-of-bounds error when CSV preprocessing fails in `civis_file_to_table`
   by raising a more informative exception (#428)
 - Removed no-longer-used PubNub code (#425)
-- Converted `assert` statements in non-test code into proper error handling (#430)
 - Removed no-longer-supported Python 2 option for notebook creation in the CLI (#421)
 - Corrected camel to snake case for "sql_type" in `io` docstrings, and added an input check to catch misspellings in the `table_columns` input (#419).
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -17,13 +17,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
   from `ContainerFuture` to `CivisFuture` (#426)
 - Updated the docstrings for `file_to_civis` (for `buf` and `expires_at`),
   `dataframe_to_file` (for `expires_at`), and `json_to_file` (for `expires_at`). (#427)
-- Added default values from swagger in client method's signature (#416)
```
jacksonlee-civis, line 20: Just cleaning up -- this is a duplicate entry for the same one a couple lines above

jclairelopez (approved):
LGMT! Thanks for giving me insight!

civis/response.py
```diff
@@ -3,6 +3,9 @@
 from civis._utils import camel_to_snake
 
 
+_RETURN_TYPES = frozenset({'snake', 'raw', 'pandas'})
```
jclairelopez, line 6: What are the snake and raw return types? I couldn't figure out what this is. Thanks Jackson!

jacksonlee-civis: It's in the docstring of `APICilent`:

https://github.com/civisanalytics/civis-python/blob/6c7213ec0bd8b8e42203540597c001f6f530a7ef/civis/civis.py#L352-L361

The default value in `APIClient` is `"snake"`, since in the Python world you'd expect the attributes to be in `snake_case`. If you pick `"raw"` (which I've never used myself in practice), you'd get attributes in `camelCase` instead.

PR #431, opened 2021-11-10
[CIVIS-1642] FIX deprecate the inactive `retry_total` parameter of `civis.APIClient`

jacksonlee-civis:
civis-python v1.15.0, released in Sept 2020, reworked the retry logic (https://github.com/civisanalytics/civis-python/pull/401), which made the `retry_total` parameter of `civis.APIClient` actually not used by anything. This present PR officially deprecates this parameter.

### DIFF
--- a/civis/civis.py
+++ b/civis/civis.py
@@ -359,9 +359,11 @@ class APIClient(MetaMixin):
         - ``'pandas'`` Returns a :class:`pandas:pandas.DataFrame` for
           list-like responses and a :class:`pandas:pandas.Series` for single a
           json response.
-    retry_total : int, optional
+    retry_total : DEPRECATED int, optional
         A number indicating the maximum number of retries for 429, 502, 503, or
-        504 errors.
+        504 errors. This parameter no longer has any effect since v1.15.0,
+        as retries are automatically handled. This parameter will be removed
+        at version 2.0.0.
     api_version : string, optional
         The version of endpoints to call. May instantiate multiple client
         objects with different versions. Currently only "1.0" is supported.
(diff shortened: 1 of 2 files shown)


civis/civis.py
```diff
@@ -379,10 +381,16 @@ class APIClient(MetaMixin):
         a local cache of the specification may be passed as either an
         OrderedDict or a filename which points to a json file.
     """
-    @deprecate_param('v2.0.0', 'resources')
+    @deprecate_param('v2.0.0', 'retry_total', 'resources')
     def __init__(self, api_key=None, return_type='snake',
                  retry_total=6, api_version="1.0", resources="all",
                  local_api_spec=None):
+        if retry_total != 6:
```
jacksonlee-civis, line 388: If someone explicitly passes in the kwarg+value of `retry_total=6`, then this wouldn't trigger the warning -- I'm okay with not handling this, since few people if anyone would actually set this kwarg, and that if they did at all it'd be a different value than 6, which would trigger the warning.

civis/civis.py
```diff
@@ -359,9 +359,11 @@ class APIClient(MetaMixin):
         - ``'pandas'`` Returns a :class:`pandas:pandas.DataFrame` for
           list-like responses and a :class:`pandas:pandas.Series` for single a
           json response.
-    retry_total : int, optional
+    retry_total : DEPRECATED int, optional
         A number indicating the maximum number of retries for 429, 502, 503, or
-        504 errors.
+        504 errors. This parameter no longer has any effect since v1.15.0,
+        as retries are automatically handled. This parameter will be removed
+        at version 2.0.0.
```
margaretott, line 366: Q: what mechanism, if any, is there to ensure we remember to remove it with version 2.0.0?

margaretott (approved):
LGTM as far as I can tell. Just one question out of curiosity.

civis/civis.py
jacksonlee-civis: No mechanism other than the maintainers remembering to audit the codebase and handle it plus other API-breaking changes; the `deprecate_param` function sort-of helps us find these things more easily. There's no timeline for a v2.0.0 release yet.

PR #432, opened 2021-11-11
[CIVIS-1640] MAINT switch from TravisCI to CircleCI

jacksonlee-civis:
Switch from TravisCI to CircleCI, since CircleCI has better support for the more recent Python versions (we're about to try to turn on support for Python 3.9 and 3.10) as well as for CI build workflow features/design.


.circleci/config.yml
```diff
@@ -0,0 +1,90 @@
...
+            twine check dist/`ls dist/ | grep .tar.gz` && \
+            twine check dist/`ls dist/ | grep .whl`
+      - build-python:
+          requires:
+            - flake8
+            - sphinx-build
+            - twine
+          matrix:
+            parameters:
+              python-version: ["3.6", "3.7", "3.8"]
```
jacksonlee-civis, line 90: [This CI workflow structure](https://app.circleci.com/pipelines/github/civisanalytics/civis-python/10/workflows/24ef9db2-3ee0-40be-8276-6bb536a947e3) allows us to separate the test suite run that we need for every supported Python 3.x version from the other checks that we need to do only once per build (flake8, sphinx-build, twine). If any of these pre-build steps fail, then the "build-python" step won't run, thus speeding up our dev work and saving CI compute resources.

Screen Shot 2021-11-16 at 9 01 55 AM

madopal (approved):
LGTM! And great to see the extra checks for python versions.

PR #436, opened 2021-11-22
[CIVIS-1508] DEP support Python 3.9, drop 3.6, bump other dep version numbers

### LINKED ISSUE #190: Consolidate dependencies across requirements.txt and setup.py
I'm paraphrasing some ideas @stephen-hoover had and sent over email:

Repeating dependencies in both setup.py and requirements.txt is problematic.  There are a few solutions to this:

1. remove requirements.txt
2. Replace the contents of requirements.txt with

```
--index-url https://pypi.python.org/simple/

-e .
```
as recommended in [this article](https://caremad.io/posts/2013/07/setup-vs-requirement/).

We should also relax the dependencies in civis-python even further. We should put version requirements in places only where we know we require that version. For example, we require requests>=2.12. But are all of the other strict requirements necessary? I think we could remove the cloudpickle version restrictions entirely, for example. Maybe also for joblib.

jacksonlee-civis:
This PR updates dependencies of civis-python:

* Supported Python versions: Add 3.9, and drop 3.6.
* Packages that civis-python depends on: Bump their upper cap version number to allow their latest major release.

Closes #190 


setup.py
```diff
@@ -49,16 +55,7 @@ def main():
...
-            'pyyaml>=3.0,<=5.99',
-            'click>=6.0,<=7.99',
-            'jsonref>=0.1.0,<=0.2.99',
-            'requests>=2.12.0,==2.*',
-            'jsonschema>=2.5.1,<=3.99',
-            'joblib>=0.11,<2',
-            'cloudpickle>=0.2.0,<2',
-            'tenacity>=6.2,<7',
-        ],
+        install_requires=requirements,
```
jacksonlee-civis, line 58: I was debating if we should specify dependencies and their versions here in `setup.py` only or in `requirements.txt` only (that's what #190 is about). I've decided to keep `requirements.txt`, since the Python packaging world is very much still in flux. With all the emerging, competing tools (pipenv, Poetry, pip-tools, etc.) and standards (Pipfile, Pipfile.lock, pyproject.toml, etc.), there seems to be a slight preference for having a `requirements.txt` file laying around, since its content is more introspectable than what's embedded in `setup.py`. Though honestly, I don't know when, if ever, we'll move to one of these standards/tools.

README.rst
```diff
@@ -138,15 +138,6 @@ dependencies with
...
 
-.. start-include-marker-python-version-support-section
-
-Python version support
-----------------------
-
-Python 3.6, 3.7, and 3.8
-
-.. end-include-marker-python-version-support-section
-
```
jacksonlee-civis, line 149: `README.rst` already has the PyPI version badges. Removing this little section also means we have one fewer spot to potentially forget to update moving forward.

setup.py
```diff
@@ -4,12 +4,16 @@
 import setuptools
 from setuptools import find_packages, setup
 
+
+_THIS_DIR = os.path.dirname(os.path.realpath(__file__))
+
+PYTHON_REQUIRES = ">=3.7"
```
jacksonlee-civis, line 10: I almost missed updating the `python_requires` kwarg in the `setup()` call below. Moving this up here as a constant next to the other related trove classifiers, so that we hopefully won't forget about it next time.

margaretott (approved):
LGTM, with the caveat that since I have not looked at this repo much, I definitely wouldn't catch any other places you may have failed to update a python version number.

PR #438, opened 2021-11-30
[CIVIS-1614] DOC specify response type in Returns documentation.

gkcivis:
Ticket: https://civisanalytics.atlassian.net/browse/CIVIS-1614

These changes are to specify the civis response type (regular vs paginated) in https://civis-python.readthedocs.io/en/v1.15.1/api_resources.html

TODO: update CHANGELOG, if this implementation is good.

### DIFF
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -155,11 +155,11 @@ def doc_from_responses(responses, is_iterable):
     properties = get_properties(schema)
     if properties:
         if is_iterable:
-            resp_type = "civis.response.PaginatedResponse\n"
+            resp_type = ":class:`civis.response.PaginatedResponse`\n"
         else:
-            resp_type = "civis.response.Response\n"
+            resp_type = ":class:`civis.response.Response`\n"
         result_doc = resp_type + (
-            "\n".join(docs_from_properties(properties)))
+            "\n".join(docs_from_properties(properties, level=1)))
     else:
         description = response_object['description']
         result_doc_fmt = "None\n    Response code {}: {}"
--- a/civis/resources/_resources.py
+++ b/civis/resources/_resources.py
@@ -146,7 +146,7 @@ def deprecated_notice(deprecation_warning):
     return "Deprecation warning!\n------------------\n" + deprecation_warning
 
 
-def doc_from_responses(responses):
+def doc_from_responses(responses, is_iterable):
     """ Return a doc string element from a responses object. The
     doc string describes the returned objects of a function.
     """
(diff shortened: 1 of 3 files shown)


civis/resources/_resources.py
```diff
@@ -146,15 +146,20 @@ def deprecated_notice(deprecation_warning):
...
     """
     response_code, response_object = next(iter(responses.items()))
     schema = response_object.get('schema', {})
     properties = get_properties(schema)
     if properties:
-        result_doc = "\n".join(docs_from_properties(properties))
+        if is_iterable:
+            resp_type = "civis.response.PaginatedResponse\n"
+        else:
+            resp_type = "civis.response.Response\n"
```
jacksonlee-civis, line 160: I was compiling the Sphinx docs from this branch and saw `civis.response.Response` show up as plain text. If you put in (I need back ticks to actually show here and don't want to fight with Markdown rendering):

```
:class:`civis.response.Response`
```

then you should be able to turn it into a hyperlink that points to wherever the `Response` class's docstring appears in the compiled Sphinx docs. Same thing for `civis.response.PaginatedResponse`.

civis/resources/_resources.py
```diff
@@ -146,15 +146,20 @@ def deprecated_notice(deprecation_warning):
...
+        result_doc = resp_type + (
+            "\n".join(docs_from_properties(properties)))
```
jacksonlee-civis, line 162: Now that you're adding the regular/paginated Response class name, I wonder if we need to indent all the docs for the attributes/keys one level "down" (so with four more spaces) to make things look nicer in the compiled docs. What do you think about wrapping them with [textwrap.indent](https://docs.python.org/3/library/textwrap.html#textwrap.indent)?

civis/resources/_resources.py
```diff
@@ -431,7 +436,10 @@ def parse_method(verb, operation, path):
         return None
 
     args, param_doc = parse_params(params, summary, verb)
-    response_doc = doc_from_responses(responses)
+    elements = split_method_params(params)
+    _, _, _, query_params, _ = elements
+    is_iterable = iterable_method(verb, query_params)
```
jacksonlee-civis, line 441: So glad that the codebase already has this handy `iterable_method` function!

jacksonlee-civis (commented):
This is great! I've left in-line comments about Sphinx formatting.

Reminders:
* changelog entry needed (I also see you have this one in the PR description)
* add "DOC" to the PR title; see other merged commits/PRs for examples

civis/resources/_resources.py
gkcivis: great idea

civis/resources/_resources.py
gkcivis: makes sense. actually seems that I can just pass ```level=1``` to ```docs_from_properties```

gkcivis:
Added change to CHANGELOG.md; made Jackson's recommended changes to add link to response class and indenting response attributes by one level. Converting to formal PR

gkcivis:
I forgot to update the related test for the newest change to link the class and increase indentation.

jacksonlee-civis (approved):
LGTM!

PR #439, opened 2021-12-02
[CIVP-24576] FIX relax SQL type checking in `civis_file_to_table`

jacksonlee-civis:
This pull request relaxes SQL type checking in `civis.io.civis_file_to_table`, by casting to `VARCHAR` when type inconsistency is detected for a given column and at least one input file has `VARCHAR`.


civis/io/_tables.py
```diff
@@ -1316,168 +1299,148 @@ def _run_cleaning(file_ids, client, need_table_columns, headers, delimiter,
...
-                               ' but detected compression {}. Please '
-                               'ensure all imported files have the same '
-                               'compression.'.format(
-                                   compression,
-                                   detected_info['compression'])
-                               )
-
-
 def _process_cleaning_results(cleaning_futures, client, headers,
                               need_table_columns, delimiter):
```
jacksonlee-civis, line 1303: I ended up having to rewriting `_process_cleaning_results` and its dependent helper functions. Before this PR, the logic of `_processing_cleaning_results` was to use the first input file as the reference and then check the second and remaining files against this first file for type consistency. Simply rewriting the helper function `_check_column_types` that checked two files' type info wouldn't be sufficient, because it'd still crash for a situation like three input files with [INT, FLOAT, VARCHAR] for a given column, where checking (INT, FLOAT) would raise an error before the code would know about a possible upcoming VARCHAR. For what we want now, we need to collect all the detected SQL types across all the files before we can apply the checking + relaxing logic.

civis/tests/test_io.py
```diff
@@ -912,66 +925,112 @@ def mock_preprocess(fid):
...
+                             {"name": "col2", "sql_type": "VARCHAR(42)"}]),
+            self._test_file([{"name": "col1", "sql_type": "INT"},
+                             {"name": "col2", "sql_type": "VARCHAR(47)"}]),
+        ]
+        actual, _ = civis.io._tables._check_column_types(files)
+        expected = [{'name': 'col1', 'sql_type': 'INT'},
+                    {'name': 'col2', 'sql_type': 'VARCHAR(42)'}]
+        assert actual == expected
+
+    def test_check_column_types_coerce_to_varchar(self, _m_get_api_spec):
```
jacksonlee-civis, line 995: Actual new test here for coercing to VARCHAR

jacksonlee-civis (commented):
@skomanduri Tagging you for review, since there's a small chance that this PR requires internal Platform knowledge that you'd know better than anyone else

civis/io/_tables.py
```diff
@@ -1310,168 +1292,148 @@ def _run_cleaning(file_ids, client, need_table_columns, headers, delimiter,
...
-        table_columns = _replace_null_column_names(table_columns)
+        table_columns, allow_inconsistent_headers = _check_column_types(files)
+    else:
+        table_columns = None
+        allow_inconsistent_headers = False
+
+    try:
+        headers = _check_detected_info(files, "includeHeader", headers)
+    except CivisImportError:
+        if allow_inconsistent_headers:
```
skomanduri, line 1322: How do you feel about this logic?  I see what you are trying to address but not sure that it covers all cases.  If you feel good about it, we can keep it!  Otherwise another option is to only do the headers and delimiter checks when the provided value is `None`.  Also happy to discuss offline.

skomanduri (approved):
This is very comprehensive!!! Just have some minor style suggestions and then one question about the inconsistent-headers logic.

civis/io/_tables.py
jacksonlee-civis: My potential blind spot here is that I don't have much background about the internal CSV preprocessing code, whose behavior I attempted to understand through my own testing via this civis-python layer (hence my long comment in `_check_column_types`). If you have potential problematic cases in mind, I'd be curious about what they might be and I'd like to address them. Otherwise I'd feel comfortable to merge this in for now and come back with future PRs if the bugs show up. :-)

civis/io/_tables.py
```diff
@@ -1310,168 +1292,144 @@ def _run_cleaning(file_ids, client, need_table_columns, headers, delimiter,
...
+    # https://stackoverflow.com/q/6473679
+    cols_by_col: List[List[Dict[str, str]]]
+    cols_by_col = list(map(list, zip(*cols_by_file)))
+
+    table_columns: List[Dict[str, str]] = []
+    allow_inconsistent_headers = False
+    err_msgs: List[str] = []
+
+    for i, cols in enumerate(cols_by_col, 1):
+        col_name = cols[0].get("name") or f"column_{i}"
```
skomanduri, line 1382: I found an issue where this didn't take the correct column name when the first file happened to be the one with only strings.  This code is a little ugly, but will take the first truthy value or default to the `column_{i}` form if there are no truthy values.

```suggestion
        col_name = next((c.get("name") for c in cols if c.get("name")), f"column_{i}")
```

skomanduri (changes requested):
@jacksonlee-civis Got it.  Thanks for explaining your process.  I was able to do some verification myself using the same methodology and script you provided and I couldn't get this to break with the cases I tried, so I think this is good to go!  I just found one small issue where the column name might not get pulled from the file but might be auto-generated instead.  Feel free to code it differently if you want.

jacksonlee-civis:
@skomanduri Thank you for catching the column name issue -- this PR is ready for another pass!

skomanduri (approved):
LGTM

PR #461, opened 2023-02-08
[CIVIS-1822] ENH `file_id_from_run_output` works for all job types

KasiaHinkson:
Replace `client.scripts.list_containers_runs_outputs` with `client.jobs.list_runs_outputs` in `civis.io.file_id_from_run_output` so it works for all job types, not just container scripts. 


KasiaHinkson:
Done, that worked. Thank you!

civis/parallel.py
```diff
@@ -710,7 +710,7 @@ def _fetch_result(fut):
 
             try:
                 # Find the output file ID from the run outputs.
-                run_outputs = client.scripts.list_containers_runs_outputs(
+                run_outputs = client.jobs.list_runs_outputs(
```
jacksonlee-civis, line 713: This `_CivisBackendResult` class is designed to work with only _container_ scripts, so I don't think we should modify `_CivisBackendResult`'s implementation.

If you take back this change and have both `scripts.list_containers_runs_outputs` and `jobs.list_runs_outputs` mocked out at the `create_client_mock_for_container_tests` function below, then the test suite will pass.

civis/tests/mocks.py
```diff
@@ -75,7 +75,7 @@ def create_client_mock_for_container_tests(
         mock_container_run['error'] = 'None'
     c.scripts.post_containers_runs.return_value = mock_container_run_start
     c.scripts.get_containers_runs.return_value = mock_container_run
-    c.scripts.list_containers_runs_outputs.return_value = (run_outputs or [])
+    c.jobs.list_runs_outputs.return_value = (run_outputs or [])
```
jacksonlee-civis, line 78: Related to the comment above -- try mocking both `scripts.list_containers_runs_outputs` and `jobs.list_runs_outputs`.

jacksonlee-civis (commented):
@KasiaHinkson Thank you for improving civis-python! Two small changes are needed (see my comments below), then this PR should be good to go.

civis/tests/mocks.py
KasiaHinkson: That makes complete sense, thank you!

jacksonlee-civis (approved):
LGTM!

PR #464, opened 2023-06-27
[CIVIS-6409] Change civis_file_to_table to not use table ids to validate that a table exists

mjziolko:
The `civis_file_to_table` relies on a table having an id to validate its existence.  Big query tables do not have ids and are not stored in RDS.  Instead, use the new `/databases/:id/schemas/:schema_name/tables/:table_name` endpoint to validate whether or not a table exists.  This endpoint works for redshift, postgres, and big query tables.


civis/cli/_cli_commands.py
```diff
@@ -240,7 +240,7 @@ def notebooks_download_cmd(notebook_id, path):
     """Download a notebook to a specified local path."""
     client = civis.APIClient()
     info = client.notebooks.get(notebook_id)
-    response = requests.get(info['notebook_url'], stream=True)
+    response = requests.get(info['notebook_url'], stream=True, timeout=60)
```
mjziolko, line 243: Bandit got mad at us for not having timeouts in our request methods:

```
Issue: [B113:request_without_timeout] Requests call without timeout
   Severity: Medium   Confidence: Low
   CWE: CWE-400 (https://cwe.mitre.org/data/definitions/400.html)
   More Info: https://bandit.readthedocs.io/en/1.7.5/plugins/b113_request_without_timeout.html
```

Console has a 60 second timeout so I added this to each of the requests we're making.

mjziolko:
@jacksonlee-civis mind reviewing?

civis/tests/test_io.py
```diff
@@ -186,7 +186,7 @@ def test_civis_file_to_table_table_exists(m_run_cleaning,
     mock_civis.get_database_id.return_value = 42
     mock_civis.default_credential = 713
 
-    mock_civis.get_table_id.return_value = 42
+    mock_civis.databases.get_schemas_tables.return_value = 42
```
jacksonlee-civis, line 189: It probably wouldn't affect this test, but for completeness, since `client.databases.get_schemas_tables` is supposed to return a `civis.response.Response` object, maybe the return value here should match that?

civis/resources/civis_api_spec.json
jacksonlee-civis, line 1: Out of curiosity, how did you update this file? IIRC, there are multiple processes folks have used over the years. I was thinking of settling on one "official" way and documenting it in this repo later.

mjziolko: I used the `tools/update_civis_api_spec_json.py` and it worked without any issues!  It did take me a while to find it though, so I definitely think it would be good to document.

mattyb:
Duplicates #467

PR #474, opened 2024-05-03
[CIVIS-5287] FIX `find` should handle boolean kwargs like other data types

### LINKED ISSUE #227: `civis.find` has unintuitive behavior with boolean kwargs
The `civis.find` (and `civis.find_one`) function allows users to filter a list of `Response` objects for user-supplied attribute/value pairs. However, the behavior when given a keyword argument with a boolean value is different than when given a keyword argument with any other value. If for example, one calls `civis.find(objs, spam='eggs')`, then the returned list contains all objects which have an attribute "spam" with value "eggs". However, `civis.find(objs, spam=True)` will return all objects which have an attribute "spam", regardless of value.

Perhaps it's worthwhile to have a way to filter on presence or absence of an attribute, but this means that there's no way to filter objects based on attributes with boolean values. I suggest that we treat booleans the same way as any other value type.

Regardless of whether we treat booleans the same way as other values, there's also a bug in the way we handle them now. If a user supplies a keyword argument with a value of `False`, then the returned list will always be empty, no matter what's in the objects.

I noticed this issue when reviewing #224 .

jacksonlee-civis:
This pull request fixes `civis.find` for its unintuitive behavior with boolean kwarg filters. As suggested in #227, boolean kwargs are now treated the same way as kwargs of other data types, for value comparison instead of presence or absence of an attribute.

Resolves #227.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   Also dropped the deprecated methods in `ServiceClient`. (#471)
 - The `return_type` parameter of a `civis.response.Response` object
   no longer has the `"pandas"` option. (#472)
+- When `civis.find` uses kwargs as filters, boolean values are now treated in the same
+  way as other data types for value equality comparison, rather than the presence or
+  absence of the key in question. (#473)
 
 ### Added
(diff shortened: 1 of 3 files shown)


Drew-Pappas (approved):
LGTM! Very clean and easy to read change

PR #477, opened 2024-05-09
[CIVIS-2564] ENH raise a warning for deprecated API endpoint methods

jacksonlee-civis:
After this pull request, a `FutureWarning` is raised when we call a deprecated API endpoint method.

---


civis/base.py
```diff
@@ -135,7 +137,12 @@ def _make_request(self, method, path=None, params=None, data=None,
 
         return response
 
-    def _call_api(self, method, path=None, params=None, data=None, **kwargs):
+    def _call_api(self, method, path=None, params=None, data=None,
+                  deprecation_warning=None, **kwargs):
+        if deprecation_warning:
+            # stacklevel=3 to point to the call just outside civis-python
+            warnings.warn(deprecation_warning, FutureWarning, stacklevel=3)
```
jacksonlee-civis, line 144: To see this action, one of the deprecated endpoint method is `client.tables.post_refresh` (as of May 2024). If we run the following code:

```python
import civis
from civis.base import CivisAPIError
from civis.resources import API_SPEC_PATH

fake_table_id = 123

# Use the local API spec to avoid making actual Civis API calls.
client = civis.APIClient(local_api_spec=API_SPEC_PATH)

try:
    client.tables.post_refresh(fake_table_id)
except CivisAPIError as e:
    if e.status_code == 404:
        # The table {fake_table_id} doesn't exist.
        # This 404 error is irrelevant for our test of FutureWarning
        # for a deprecated endpoint method.
        pass
    else:
        raise
```

we get this warning that points to where the line of code with `client.tables.post_refresh` is:

```
<some-local-path-and-line-number>: FutureWarning: Warning: The tables/:id/refresh endpoint is deprecated. Please use tables/scan from now on.
  client.tables.post_refresh(fake_table_id)
```

civis/base.py
```diff
@@ -103,7 +105,7 @@ class Endpoint:
 
     _lock = threading.Lock()
 
-    def __init__(self, session_kwargs, client, return_type='civis'):
+    def __init__(self, session_kwargs, client, return_type="raw"):
```
jacksonlee-civis, line 108: Fixing this hidden bug that never got exposed. This `return_type` is for the `civis.response.Response` objects, where {"raw", "snake"} are the only options. We've never run into an exception of an unrecognized return type "civis", because this `Endpoint` class is only used dynamically (from [here](https://github.com/civisanalytics/civis-python/blob/6e18575234ecf5bdca2e801ab710041b312146c9/civis/resources/_resources.py#L472) onwards) to generate the individual endpoint classes. These endpoint class instances are then created when a `civis.APIClient` instance is created, where `return_type` is set to "snake" by default ([here](https://github.com/civisanalytics/civis-python/blob/a884416673a358436703732aa1b8127bbde48213/civis/civis.py#L387)).

Drew-Pappas: Good catch!

Drew-Pappas (approved):
LGTM

PR #483, opened 2024-05-16
[CIVIS-8476] DOC configure "readthedocs" builds and refresh docs

jacksonlee-civis:
This pull request configures the "readthedocs" doc builds by adding `.readthedocs.yaml`. Also added a requirements file to pin all transitive dependencies for doc builds, as well as refreshed all the docs in general.

---


jacksonlee-civis:
Needed to actually create a pull request to test/update the "readthedocs" doc builds. This PR is not ready for review yet. If you're watching this PR, expect noisy commits as I iterate (note that you can unsubscribe from this PR).

jacksonlee-civis:
@Drew-Pappas I went back and forth about these doc updates, but apparently the readthedocs builds have actually kicked in for PR branches. For this PR, see https://civis-python--483.org.readthedocs.build/en/483/ -- let me know if you don't like anything that I can fix before making the v2.0.0 release!

Drew-Pappas (approved):
A couple of things that are non-blocking, but would be "nice to haves with the docs"

1. Reading some of the longer methods can be pretty cumbersome. See `client.scripts.get()` [here](https://civis-python--483.org.readthedocs.build/en/483/api_scripts_endpoint.html#civis.resources._resources.Scripts.get). Is there any way to modify the spacing such that it takes up less vertical space? Alternatively it'd be nice if the method could "pin" to the top as you're scrolling, but given we're working with auto formatted docs that probably wouldn't be the easiest to get working.
2. In the [user guide](https://civis-python--483.org.readthedocs.build/en/483/user_guide.html), logging looks a little bit awkward being an openable menu when the others aren't.
3. An "examples" section under "running jobs and templates" or within the user guide might be pretty nice to have for people looking into the package for the first time. However, this probably isn't necessary for a 2.0 release.

Otherwise, LGTM! I really like that the API resources section is no longer just one big page. It should be significantly easier to ctrl + f within a page now!

PR #486, opened 2024-05-23
[CIVIS-8647] ENH response objects: add `.json()`, preserve key casing under 'arguments'

jacksonlee-civis:
`civis.response.Response` objects turn keys into snake case from Civis API's camel-case key names. There are keys that shouldn't be converted for casing, specifically those under `"arguments"` for scripts, as the argument names are from environmental variables. This pull request fixes this issue. Also adding `.json()` to `civis.response.Response` if an actual dict of the response is desired.

Also refreshed the Civis API spec and the corresponding stub file.

---


Drew-Pappas (approved):
LGTM!

PR #490, opened 2024-06-07
[CIVIS-8658] PERF speed up startup time

jacksonlee-civis:
The startup time when just running `import civis` in a fresh Python interpreter session appears to be long. This is especially noticeable in an interactive environment (the regular Python interactive interpreter, IPython, or a Jupyter notebook), which isn't great for user experience. This pull request applies lazy imports to the top-level `civis.*` modules, for a speed boost of about 5x. The relatively long startup time was due to importing dependencies like pandas through `civis.io` and other pydata packages (numpy, scipy) through `civis.ml`.

We can use `python -X importtime -c 'import civis'` to get the import time details. Before this pull request (for brevity, just showing the final line for the top-level `civis` import):

```
import time: self [us] | cumulative | imported package
...
import time:     14290 |    3144744 | civis
```

After this PR:

```
import time: self [us] | cumulative | imported package
...
import time:      5978 |     606370 | civis
```

It's potentially possible to speed things up further by coating a few more things along the import paths with lazy imports. I tried that with some success, but I started to have to make _other_ changes as well to have the test suite pass (e.g., `ModuleNotFoundError` for existing working imports, where tweaking the import syntax here and there might resolve the issue). That suggested to me that it's not worth the effort. Keeping the lazy import implementation simple for a 5x speed-up is good enough.

---

### DIFF
--- a/src/civis/_utils.py
+++ b/src/civis/_utils.py
@@ -1,6 +1,5 @@
 import logging
 import os
-import re
 import time
 import uuid
 from random import random
@@ -28,10 +27,6 @@ def maybe_get_random_name(name):
     if not name:
         name = uuid.uuid4().hex
     return name
-
-
-def to_camelcase(s):
-    return re.sub(r"(^|_)([a-zA-Z])", lambda m: m.group(2).upper(), s)
 
 
 def get_api_key(api_key):
--- a/src/civis/_utils.py
+++ b/src/civis/_utils.py
@@ -16,8 +16,7 @@ from tenacity.wait import wait_base
 
 
 log = logging.getLogger(__name__)
-UNDERSCORER1 = re.compile(r"(.)([A-Z][a-z]+)")
-UNDERSCORER2 = re.compile("([a-z0-9])([A-Z])")
+
 MAX_RETRIES = 10
 
 _RETRY_CODES = [429, 502, 503, 504]
@@ -29,12 +28,6 @@ def maybe_get_random_name(name):
     if not name:
         name = uuid.uuid4().hex
     return name
-
-
-def camel_to_snake(word):
-    # https://gist.github.com/jaytaylor/3660565
-    word = UNDERSCORER1.sub(r"\1_\2", word)
-    return UNDERSCORER2.sub(r"\1_\2", word).lower()
 
 
 def to_camelcase(s):
--- a/src/civis/resources/_resources.py
+++ b/src/civis/resources/_resources.py
@@ -11,7 +11,8 @@ import requests
 from requests import Request
 
 from civis.base import Endpoint, get_base_url, open_session
-from civis._utils import camel_to_snake, get_api_key, retry_request, MAX_RETRIES
+from civis._camel_to_snake import camel_to_snake
+from civis._utils import get_api_key, retry_request, MAX_RETRIES
 
 
 API_VERSIONS = frozenset({"1.0"})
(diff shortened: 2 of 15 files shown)


src/civis/_utils.py
```diff
@@ -31,16 +29,6 @@ def maybe_get_random_name(name):
     return name
 
 
-def camel_to_snake(word):
```
jacksonlee-civis, line 34: The `_utils.py` module here uses `tenacity` for retries, which has a fair amount of overhead in imports. Moving `camel_to_snake` to its own module so that other modules needing this function no longer have to import `_utils.py`.

src/civis/_utils.py
```diff
@@ -31,16 +29,6 @@ def maybe_get_random_name(name):
...
-    # https://gist.github.com/jaytaylor/3660565
-    word = UNDERSCORER1.sub(r"\1_\2", word)
-    return UNDERSCORER2.sub(r"\1_\2", word).lower()
-
-
-def to_camelcase(s):
```
jacksonlee-civis, line 40: The only module that uses `to_camelcase` is `service_client.py`, so I'm moving this function there.

src/civis/resources/_resources.py
```diff
@@ -22,8 +23,6 @@
     os.path.dirname(os.path.realpath(__file__)),
     "civis_api_spec.json",
 )
-with open(API_SPEC_PATH) as f:
-    API_SPEC = json.load(f, object_pairs_hook=OrderedDict)
```
jacksonlee-civis, line 26: Having to load a file for JSON parsing at the module level incurs quite a bit of import overhead. `API_SPEC` is only used in the test suite, so I'm moving it there.

src/civis/resources/_resources.py
```diff
@@ -334,7 +335,7 @@ def raise_for_unexpected_kwargs(
 
 
 def bracketed(x):
-    return re.search("^{.*}$", x)
+    return _BRACKETED_REGEX.search(x)
```
jacksonlee-civis, line 338: On the theme of speeding things up and as I saw regex from `camel_to_snake` above, I've noticed we aren't using compiled regex expressions where possible, but we should. Many of these little functions in this module are heavily used in handling an API spec, so we should optimize them:

```python
In [1]: import re

In [2]: %timeit re.search(r"^{.*}$", "foobar")
521 ns ± 4.13 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [3]: _BRACKETED_REGEX = re.compile(r"^{.*}$")

In [4]: %timeit _BRACKETED_REGEX.search("foobar")
159 ns ± 1.84 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

src/civis/resources/_resources.py
```diff
@@ -457,7 +458,7 @@ def parse_method_name(verb, path):
     verb = "list" if verb == "get" and (not bracketed(final_elem)) else verb
     path_name = "_".join(name_elems)
     method_name = "_".join((verb, path_name)) if path_name else verb
-    return re.sub("-", "_", method_name)
+    return method_name.replace("-", "_")
```
jacksonlee-civis, line 461: When we don't actually need regex, don't use it:

```python
In [1]: import re

In [2]: %timeit "foobar".replace("-", "_")
36.2 ns ± 0.58 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [3]: %timeit re.sub("-", "_", "foobar")
510 ns ± 7.49 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

Drew-Pappas (approved):
LGTM! +1 for using timeit benchmarks

PR #492, opened 2024-06-12
[CIVIS-1949] PERF more efficient polling at future objects

jacksonlee-civis:
This pull request makes polling at future objects more efficient, in that a short-running job's `future.result()` should return faster than before.

Before this PR, the default polling interval was a constant of 15 seconds. After this PR, the default polling interval starts at 1 second, increases geometrically for subsequent polls, and is capped at a maximum of 15 seconds. If the user provides a number to the `polling_interval` argument, the polling interval will be a constant of this number (so no change from before this PR).

---

### DIFF
--- a/src/civis/base.py
+++ b/src/civis/base.py
@@ -173,39 +173,7 @@ class CivisAsyncResultBase(futures.Future):
     the result. The `_result_` object needs to be set to an object with a
     ``state`` attribute. Alternatively the `_check_result` method can be
     overwritten to change how the state of the object is returned.
-
-    Parameters
-    ----------
-    poller : func
-        A function which returns an object that has a ``state`` attribute.
-    poller_args : tuple
-        The arguments with which to call the poller function.
-    polling_interval : int or float
-        The number of seconds between API requests to check whether a result
-        is ready.
-    client : :class:`civis.APIClient`, optional
-        If not provided, an :class:`civis.APIClient` object will be
-        created from the :envvar:`CIVIS_API_KEY`.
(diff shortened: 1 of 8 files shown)


src/civis/base.py
```diff
@@ -173,40 +173,8 @@ class CivisAsyncResultBase(futures.Future):
...
-        client=None,
-        poll_on_creation=True,
-    ):
-        super().__init__()
-        self.poller = poller
-        self.poller_args = poller_args
-        self.polling_interval = polling_interval
-        self.client = client
-        self.poll_on_creation = poll_on_creation
-
```
jacksonlee-civis, line 209: The polling logic is in `PollableResult`, which subclasses `CivisAsyncResultBase` here. Since `CivisAsyncResultBase` isn't supposed to have any knowledge of polling, I've moved the initialization of these attributes to `PollableResult` below.

src/civis/ml/_model.py
```diff
@@ -555,6 +559,8 @@ def __setstate__(self, state):
         self._condition = threading.Condition()
         self.client = APIClient()
         self.poller = self.client.scripts.get_containers_runs
+        self._next_polling_interval = 1
+        self._use_geometric_polling = True
```
jacksonlee-civis, line 563: `ModelFuture` objects are designed to be pickle-able, so we need to restore the default values of these polling-related attributes upon un-pickling.

Drew-Pappas (approved):
LGTM. The time savings on this change will add up very quickly across all code bases using civis-python!

PR #493, opened 2024-06-14
[CIVIS-8695] ENH SQL I/O functions accept SQL parameters and arguments

jacksonlee-civis:
This pull request adds a new feature to the `civis.io.*` functions that call `client.scripts.post_sql` under the hood, for passing in SQL parameters and arguments. This new feature lets the user take advantage of Civis Platform's parameterized SQL scripts with functionality to mitigate security risks of SQL injections. See the linked internal ticket for testing this new feature.

Also going to make a new release of v2.3.0 after this pull request, so I'm bundling in here a few small fixes as well -- they're noted both in the changelog and comments on this PR.

---

### DIFF
--- a/src/civis/__init__.py
+++ b/src/civis/__init__.py
@@ -1,6 +1,7 @@
 import importlib
 import sys
 from importlib.metadata import version
+from typing import TYPE_CHECKING
 
 from civis.client import APIClient
 from civis.loggers import civis_logger
@@ -19,11 +20,14 @@ def _lazy_import(name):
     return module
 
 
-futures = _lazy_import("civis.futures")
-io = _lazy_import("civis.io")
-ml = _lazy_import("civis.ml")
-parallel = _lazy_import("civis.parallel")
-utils = _lazy_import("civis.utils")
+if TYPE_CHECKING:
+    from civis import futures, io, ml, parallel, utils
+else:
+    futures = _lazy_import("civis.futures")
+    io = _lazy_import("civis.io")
+    ml = _lazy_import("civis.ml")
+    parallel = _lazy_import("civis.parallel")
+    utils = _lazy_import("civis.utils")
 
 __version__ = version("civis")
 __all__ = [
--- a/src/civis/__init__.py
+++ b/src/civis/__init__.py
@@ -31,6 +31,7 @@ __all__ = [
     "APIClient",
     "find",
     "find_one",
+    "futures",
     "io",
     "civis_logger",
     "ml",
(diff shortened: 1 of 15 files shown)


src/civis/__init__.py
```diff
@@ -19,18 +20,22 @@ def _lazy_import(name):
...
-parallel = _lazy_import("civis.parallel")
-utils = _lazy_import("civis.utils")
+if TYPE_CHECKING:
+    from civis import futures, io, ml, parallel, utils
+else:
+    futures = _lazy_import("civis.futures")
+    io = _lazy_import("civis.io")
+    ml = _lazy_import("civis.ml")
+    parallel = _lazy_import("civis.parallel")
+    utils = _lazy_import("civis.utils")
```
jacksonlee-civis, line 30: After #490, one thing I've noticed locally was that my IDE had become blind for type hints and code autocompletes. Fixing this issue by using real imports for tools of static type checks while keeping the lazy imports for actual runtime.

src/civis/client.pyi
jacksonlee-civis, line 1: Lots of lines of code changes, but they're all just about `list -> List` to appease static type checks.

src/civis/response.py
```diff
@@ -230,9 +223,7 @@ def __hash__(self):
 
     def _to_dataclass_repr_pprint(self):
         """Convert the response to a dataclass for repr and pprint purposes."""
-        # Only show the "Response" class name at the outermost level.
-        class_name = "" if self._inner_response else "Response"
-        klass = dataclasses.make_dataclass(class_name, self._data_snake.keys())
+        klass = dataclasses.make_dataclass("Response", self._data_snake.keys())
```
jacksonlee-civis, line 226: I was trying to make the repr look nicer with less indentation by not showing the class name "Response" in embedded objects, but in my actual usage, not showing it for embedded objects had actually make it harder to read :-(

src/civis/response.py
```diff
@@ -262,7 +253,7 @@ def __eq__(self, other):
         elif isinstance(other, Response):
             return self._data_snake == other._data_snake
         else:
-            raise TypeError(f"Response and {type(other)} can't be compared")
+            return False
```
jacksonlee-civis, line 256: This is borderline between a behavior change and a bug. In a way, I'm loosening the strictness of the equality check here. I ran into multiple cases where a `Response` object was compared to a non-`Response` object, and making it return `False` here instead of raising an error would have made life a lot easier.

Drew-Pappas (approved):
LGTM!

PR #498, opened 2024-08-09
[CIVIS-9155] ENH support setting a Civis file's new `description` attribute

jacksonlee-civis:
The Civis API has been updated with a new `description` attribute for Civis file objects. This PR updates the `civis.io.*` functions that create a Civis file with a new `description` kwarg to support setting a Civis file's description.

---


src/civis/io/_files.py
```diff
@@ -196,7 +187,7 @@ def _upload_part_base(item, file_path, part_size, file_size):
     return file_response.id
 
 
-def file_to_civis(buf, name=None, client=None, **kwargs):
+def file_to_civis(buf, name=None, expires_at="DEFAULT", description=None, client=None):
```
jacksonlee-civis, line 190: The `**kwargs` before this PR would only accept the `expires_at` key. It's clear by now that we don't really add new features to Civis files that often, so I'm switching to explicitly listing the kwargs here (currently only `expires_at` plus the new `description`) one by one. This update doesn't break existing code calling this function (unless someone has specified the `client` kwarg as a _positional_ arg? seems unlikely, since we always recommend using the kwarg syntax of `client=client` to specify `client`). It also has the advantage of making the function signatures consistent between `file_to_civis`, `dataframe_to_file`, and `json_to_file`.

src/civis/resources/civis_api_spec.json
jacksonlee-civis, line 1: Updating `civis_api_spec.json` for the new `description` attribute under the `/Files` endpoint, or else the relevant unit tests using the mock client object wouldn't work.

src/civis/client.pyi
jacksonlee-civis, line 1: Nothing to review here practically. `client.pyi` is auto-generated / updated when we run `python tools/update_civis_api_spec.py` to update `civis_api_spec.json`.

margaretott (approved):
LGTM!

PR #499, opened 2024-08-13
[CIVIS-8846] FIX validation checks of a workflow YAML definition; DEP drop Python 3.9 support

jacksonlee-civis:
This pull request is a follow-up of #497, where I added the new `civis.workflows.validate_workflow_yaml` function and now I'm updating/fixing it. In #497, I missed a few things for validation:

* A workflow YAML definition must contain only ASCII characters (a constraint that apparently inherits from the Mistral engine). I've verified that both the Civis Platform GUI and Civis API would raise an error for non-ASCII characters, so `civis.workflows.validate_workflow_yaml` should check for that as well.
* For task transition, the Mistral workflow language apparently allows an array of tasks to be specified with a mixture of task names or a dict/object mapping a task name to a YAQL expression (which I only learned about from a real workflow YAML definition in production in one of the internal projects).

I'm also bumping up the versions of the dev and documentation dependencies. Since the latest `scipy` v1.14.0 allows only Python >= 3.10 and I'm always more than happy to help folks move things forward, I'm dropping support for Python 3.9 in civis-python.

---


src/civis/resources/civis_api_spec.json
jacksonlee-civis, line 1: This `civis_api_spec.json` file is the basis of the Civis API information on the "API Resources" page of the Sphinx / Read The Docs site. We update this file by running `python tools/update_civis_api_spec.py` (explained in `README.rst`) whenever the upstream Civis API information has been updated. Especially relevant for this pull request is that the new dbt job type on Civis Platform is now generally available to all standard Civis Platform users, and so we need to update `civis_api_spec.json` for workflow YAML validation related to the task action of "civis.scripts.dbt".

src/civis/client.pyi
jacksonlee-civis, line 1: Nothing to review here practically. `client.pyi` is auto-generated / updated when we run `python tools/update_civis_api_spec.py` to update `civis_api_spec.json`.

tests/test_workflows/test_validate.py
```diff
@@ -29,32 +29,35 @@ def test_valid_workflow_yaml():
 
 
 @pytest.mark.parametrize(
-    "replace_to_break",
+    "replacee, replacer, error_message_contains",
     [
-        # "version" is required
-        ('version: "2.0"', ""),
```
jacksonlee-civis, line 35: Rewriting this test so that the comments that explain the offense are now formally part of the test (by checking the error message in the `pytest.raises` call).

gkcivis (approved):
LGTM

PR #479, opened 2024-05-13
[CIVIS-8439] ENH add stub file for API endpoints

jacksonlee-civis:
The API endpoints are auto-generated attributes attached to a `civis.APIClient` instance. Since dynamically generated objects don’t play nice with static type checkers (such as those embedded in an IDE as well as tools like mypy), writing code to hit the Civis API via a `civis.APIClient` instance has been somewhat prone to errors due to the lack of IDE support. This pull request adds a [stub file](https://typing.readthedocs.io/en/latest/source/stubs.html) `src/civis/client.pyi` for the API endpoints as attributes attached to `civis.APIClient`, so that our IDE can now help us catch potential coding errors -- see the screen captures below for PyCharm and VS Code.

Related changes:
* Given the relatively large file size of the stub file, we can somewhat mitigate the increase in package size by adopting the repo structure of `src/` and `tests/` at the repo root level, so that the test suite isn’t included in the distributed package.
* The module `civis.py` with a less descriptive name, where the `APIClient` class is defined, has been renamed `client.py`. This is why the corresponding stub file needs to be named `client.pyi`.
* Having the stub file for the API endpoints has made it much easier to fix several minor long-standing API doc issues, mostly about docstring formatting.

---

PyCharm:

pycharm

VS Code:

vscode

---

- [x] (For Civis employees only) Reference to a relevant ticket in the pull request title
- [x] Changelog entry added to `CHANGELOG.md` at the repo's root level
- [x] Description of change in the pull request description
- [x] If applicable, unit tests have been added and/or updated
- [x] The CircleCI builds have all passed

PR #501, opened 2024-10-28
[CIVIS-7940] [CIVIS-8862] ENH API response objects: typing, repr, fixes for key and value handling

jacksonlee-civis:
This pull request improves `Response` and `PaginatedResponse` in various ways.

New features:
* To facilitate testing and local development, `Response` objects are now fully typed through the attribute syntax. (This update is a follow-up of #479, where `civis.APIClient` has since been fully typed for the API endpoints and methods.) Typing `response.` in an IDE such as VS Code now triggers autocomplete suggestions for available attributes. In testing, static type checkers such as `mypy` can now catch possible code issues based on response attribute access. See the screen capture below.
* For simpler code, both `Response` and `PaginatedResponse` are now directly available under the `civis` namespace. This means that `civis.Response` and `civis.PaginatedResponse` (in addition to the existing `civis.response.Response` and `civis.response.PaginatedResponse`) are now valid, and that for imports, the statement `from civis import Response, PaginatedResponse` is also valid (in addition to the existing `from civis.response import Response, PaginatedResponse`).

Changed:
* For an API call from `civis.APIClient` that returns a `PaginatedResponse` object (with `iterator=True`), the user-specified kwarg `limit` is now honored to facilitate speeding up the pagination.

Bug fixes:
* The repr form of `Response` objects now looks like the dict-based `Response({‘spam’: 123})` instead of the dataclass-based `Response(spam=123)`, because response object keys can be invalid Python identifiers (e.g., containing dashes or forward slashes, or even just an empty string).
* In `Response` object instantiation, object keys that originate from environment variables (e.g., under `environmentVariables` in the `/services` endpoint) are now preserved for their (customarily upper-) case even in the default snake-case setting.
* In `Response` object instantiation, an API response that represents a JSONValue object now has its `value` attribute unmodified as the Python object representation of the deserialized JSON form (as opposed to being converted to a `Response`-based form).

---

VS Code screen capture for the attribute knowledge of an API response object:

civis-python-response-typing

---


docs/source/responses.rst
jacksonlee-civis, line 1: All Sphinx doc changes can be reviewed through the built page for this PR at https://civis-python--501.org.readthedocs.build/en/501/

Drew-Pappas (approved):
LGTM! Great changes + fixes all around. The attribute knowledge capture will help encourage API endpoint exploration and remove a ton of the dev headaches of flipping back and forth between the readthedocs information and local IDE

PR #509, opened 2025-02-20
[CIVIS-10251] Add `civis.utils.job_logs()` function

mheilman:
This does two things:
* Adds `civis.utils.job_logs()` function to return a generator of log messages for a job run
* Revises the CLI commands `civis jobs follow-log` and `civis jobs follow-run-log` to not skip log messages for running jobs

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Added `civis.utils.job_logs()` function to return a generator of log messages for a job run (#509)
-
 ### Changed
-
-- Revised the CLI commands `civis jobs follow-log` and `civis jobs follow-run-log` to not skip log messages for running jobs (#509)
 
 ### Deprecated
 
@@ -21,6 +17,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 ### Security
+
+## 2.5.0 - 2025-02-21
+
+### Added
+
+- Added `civis.utils.job_logs()` function to return a generator of log messages for a job run (#509)
+
+### Changed
+
+- Revised the CLI commands `civis jobs follow-log` and `civis jobs follow-run-log` to not skip log messages for running jobs (#509)
 
 ## 2.4.3 - 2025-01-13
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `civis.utils.job_logs()` function to return a generator of log messages for a job run (#509)
+
 ### Changed
+
+- Revised the CLI commands `civis jobs follow-log` and `civis jobs follow-run-log` to not skip log messages for running jobs (#509)
 
 ### Deprecated
 
@@ -114,7 +118,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   top-level and nested response objects. (#493)
 
 ### Security
-- Bumped the minimum required version of `requests` to the latest v2.32.3, 
+- Bumped the minimum required version of `requests` to the latest v2.32.3,
   due to a security vulnerability for < v2.32.0
   ([CVE-2024-35195](https://nvd.nist.gov/vuln/detail/CVE-2024-35195)). (#488)
 
@@ -181,7 +185,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 - Updated references from 'master' to 'main' (#460)
-- Clarified the usage example for `civis.io.civis_to_multifile_csv`. Updated 
+- Clarified the usage example for `civis.io.civis_to_multifile_csv`. Updated
   CircleCI config so dev-requirements is only used when needed. (#452)
 - Removed unneeded `time.sleep` calls and `pytest.mark` calls and mocked `time.sleep` calls to optimize tests. (#453)
 - Refactored tests to remove dependency on the vcr library. (#456)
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,8 +8,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
+- Added `civis.utils.job_logs()` function to return a generator of log messages for a job run (#509)
```
mheilman, line 11: Should I make a 2.5.0 heading or leave that for a followup PR?

src/civis/utils/_jobs.py
```diff
@@ -1,10 +1,18 @@
...
+import operator
+import time
+from datetime import datetime
 
 from civis import APIClient
 from civis.futures import CivisFuture
 
 log = logging.getLogger(__name__)
 
+_FOLLOW_POLL_INTERVAL_SEC = 5
```
mheilman, line 11: I increased the interval from 3 to 5 seconds here since there's an extra API call to check whether the job is done. Also, 3 seemed faster than necessary anyway.

src/civis/utils/_jobs.py
```diff
@@ -96,3 +104,149 @@ def run_template(id, arguments, JSONValue=False, client=None):
...
+    except ValueError:
+        try:
+            # This is the format that Civis Platform returns.
+            return datetime.strptime(s, "%Y-%m-%dT%H:%M:%S.%f%z").timestamp()
+        except ValueError:
+            # Another format, just in case.
+            return datetime.strptime(s, "%Y-%m-%dT%H:%M:%S%z").timestamp()
+
+
+def _compute_effective_max_log_id(logs):
```
mheilman, line 126: There's fairly similar logic in our internal react code for the log display component of the scripts interface.

mheilman:
Here's a little snippet that I put in a python job for testing: 

```python
import time
for i in range(30):
    print(i)
    time.sleep(1)
```

And then I ran this locally:

```python
import civis
job_id = ...
civis.utils.run_job(job_id)
for msg in civis.utils.job_logs(job_id, finished_timeout=20):
    print(msg)
```

src/civis/utils/_jobs.py
```diff
@@ -96,3 +104,143 @@ def run_template(id, arguments, JSONValue=False, client=None):
...
+
+    Yields
+    ------
+    dict
+        A log message dictionary with "message", "createdAt" and other attributes
+        provided by the job logs endpoint. Note that this will block execution
+        until the job has stopped and all log messages are retrieved.
+    """
+    if raw_client is None:
+        raw_client = APIClient(return_type="raw")
```
jacksonlee-civis, line 196: This requirement of a "raw" client seems odd. Is it possible to update this function to use the regular default / "snake" client instead? I know this "raw" client comes from the previously internal-only code , but now that we're exposing this argument in a public function like this, it would seem odd for this and only this place to require a "raw" client, while everywhere else in civis-python the regular "snake" client is required. To use this function as-is while needing to explicitly pass in an API client object (e.g., when handling both robot and non-robot API keys in the same execution environment -- not too uncommon these days in internal work), one would have to instantiate two client objects for a given API key, the "raw" one just for this function and the regular one for everything else, which would be strange.

The default "snake" `client` would get us a list of `civis.Response` objects returned by `client.jobs.list_runs_logs(...)`. The `headers` dict is [available at an attribute by the same name](https://github.com/civisanalytics/civis-python/blob/d5153897914ed799164637672bd18a2492a023f1/src/civis/response.py#L132) at the first `civis.Response` object in this list (all objects in this list [has the same `headers` content](https://github.com/civisanalytics/civis-python/blob/d5153897914ed799164637672bd18a2492a023f1/src/civis/response.py#L92-L96)). When looping through each of these response objects, calling `.json()` on this object would return [a dict with "snake-case" keys by default](https://github.com/civisanalytics/civis-python/blob/d5153897914ed799164637672bd18a2492a023f1/src/civis/response.py#L169), which has the added benefit of using/expecting "created_at" not "createdAt" etc. to conform to Python's preference for snake-case styling.

mheilman: Oh, I didn't realize `headers` was part of the `Response` object. Perhaps it wasn't when this code was originally written.  I'll give that a try.

mheilman: Hmm, it looks like the normal API client will return an empty list without the headers if there are no new logs, so the raw client is needed to check `civis-cache-control` when that happens. I don't immediately see a way around this. Thoughts?

jacksonlee-civis: I see! How about if the user provides a regular "snake" client, the function internally creates a "raw" one based on the user-provided one? I was only 50-50 for whether this approach might work, so I wanted to be more certain -- [this](https://github.com/civisanalytics/civis-python/compare/add-log-generation-function...add-log-generation-function-use-regular-client) seems to work?

If you think this direction is a no-go, i.e., if a user-provided client must be a "raw" one, I'd recommend the following:
* Name the argument `client` instead of `raw_client`, to anticipate that we may one day be able to allow a user-provided regular "snake" client in a future release, without having to deal with a breaking change from `raw_client` to `client`.
* If a "raw" client is required, it should be formally checked (in addition to documentation) and we should raise an exception if this requirement is not met (in which case we likely still would need to keep a `self._return_type` directly at a client object for this purpose).

mheilman: Hmm... I'm concerned there might be issues if someone used multithreading with this approach where `_return_type` is changed and then changed back. We could add a lock of some kind, but that seems overly complex.

The main reason I was passing in `raw_client` was to avoid an extra API call to `/endpoints`, but `generate_classes_maybe_cached` should help avoid that. How about I just not have a `client` or `raw_client` argument, which is how the CLI command was before this?

mheilman: OK. I removed the arg in [41c381a](https://github.com/civisanalytics/civis-python/pull/509/commits/41c381a0a9b808c1e6709783ac8c2e8e8a680027)

jacksonlee-civis (approved):
LGTM

PR #521, opened 2025-07-02
[CIVIS-11076] ENH add `user_agent` kwarg to `civis.APIClient`

jacksonlee-civis:
This pull request adds a new `user_agent` kwarg to `civis.APIClient`.

---

### DIFF
--- a/src/civis/client.py
+++ b/src/civis/client.py
@@ -59,11 +59,10 @@ class APIClient:
         please note that you should leave the ``retry`` attribute unspecified,
         because the conditions under which retries apply are pre-determined
         -- see :ref:`retries` for details.
-    headers : dict, optional
-        Additional headers to include in requests made by this client.
-        If a "User-Agent" header is provided, it will be prepended to information
-        including the Civis Python client version, Python version, and the requests
-        package version.
+    user_agent : str, optional
+        A custom user agent string to use for requests made by this client.
+        The user agent string will be appended with the Python version,
+        Civis Python client version, and requests package version.
     """
 
     def __init__(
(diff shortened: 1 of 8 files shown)


src/civis/client.py
```diff
@@ -69,14 +74,19 @@ def __init__(
         local_api_spec: collections.OrderedDict | str | None = None,
         force_refresh_api_spec: bool = False,
         retries: tenacity.Retrying | None = None,
+        headers: dict | None = None,
```
jacksonlee-civis, line 77: I've decided to add a `headers` kwarg for generality (in case in the future a user asks if they can provide custom headers) rather than just `user_agent` here. I think the risk of cluttering our internal bookkeeping due to custom headers should be low, but let me know if you have concerns and/or want me to switch to just allowing `user_agent`.

jacksonlee-civis (commented):
@mattyb Would you be able to test drive this civis-python branch for your use case?

Install from this branch:

```
pip install --force-reinstall git+https://github.com/civisanalytics/civis-python.git@civis-11076-headers-at-APIClient
```

Update your code with something like this:

```python
client = civis.APIClient(
    ...,
    headers={"User-Agent": "your custom user agent here"},
)
```

jacksonlee-civis:
Self reminders if we move forward with this implementation:

* [ ] Update changelog
* [ ] Bump version to v2.7.0 for a new civis-python release

jacksonlee-civis:
@mattyb Thank you for your feedback! I've updated this PR:

* Switched to just exposing a `user_agent` kwarg (not a generic `headers`) at `civis.APIClient`.
* Updated the string appended to the custom user agent provided, so that if you provide `"test script"`, the user agent string looks like `test script (Python 3.13.2; civis-python 2.7.0; requests 2.32.4)`, from which we could target substrings like "civis-python".

This pull request is ready for review. To test it, the following code snippet has logging turned on for headers:

```python
import http.client
import civis

http.client.HTTPConnection.debuglevel = 1
client = civis.APIClient(user_agent="test script")
client.users.list_me()
```

The logs from this code should show two API calls, first for `GET /endpoints` to get the upstream Civis API spec for this platform user, and then for `GET /users/me` due to `client.users.list_me()`. For both calls, the user agent was `test script (Python 3.13.2; civis-python 2.7.0; requests 2.32.4)`.

PR #530, opened 2026-01-26
[CIVIS-12338] DEP Fix "list" versus "get" API methods; release for v2.8.0

jacksonlee-civis:
Some of the `list*` methods from various Civis API endpoints should have been named `get*`, because they return a singleton `civis.Response` object, as opposed to an array of such objects. These `list*` methods are now deprecated. The corresponding equivalent `get*` methods are added and should be preferred. The full list of these methods are in the new changelog entry.

For most users, the impact is expected to be minimal. The vast majority of the `list*` methods in question are not commonly used, with the notable exception of `client.users.list_me`.

After this pull request is merged, I'll make a new civis-python release for v2.8.0.

---

### DIFF
--- a/src/civis/base.py
+++ b/src/civis/base.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import os
 import platform
 import threading
-import warnings
 from concurrent import futures
 from json.decoder import JSONDecodeError
 from posixpath import join
@@ -193,13 +192,8 @@ class Endpoint:
         path=None,
         params=None,
         data=None,
-        deprecation_warning=None,
         **kwargs,
     ):
-        if deprecation_warning:
-            # stacklevel=3 to point to the call just outside civis-python
-            warnings.warn(deprecation_warning, FutureWarning, stacklevel=3)
-
         iterator = kwargs.pop("iterator", False)
 
         if iterator:
--- a/src/civis/ml/_helper.py
+++ b/src/civis/ml/_helper.py
@@ -61,7 +61,7 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
     template_id_str = ", ".join(str(tmp) for tmp in set(template_id_list))
 
     if author is SENTINEL:
-        author = client.users.list_me().id
+        author = client.users.get_me().id
 
     # default to showing most recent models first
     kwargs.setdefault("order_dir", "desc")
--- a/src/civis/resources/_resources.py
+++ b/src/civis/resources/_resources.py
@@ -576,7 +576,6 @@ def parse_param_body(parameter):
     return arguments
 
 
-# TODO: Write tests for this function.
 def get_response_is_array(path, operation) -> bool:
     schema = operation["responses"].get("200", {}).get("schema", {})
     if schema.get("type") == "array":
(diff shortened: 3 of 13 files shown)


src/civis/resources/_resources.py
```diff
@@ -67,6 +68,31 @@
 
 _BRACKETED_REGEX = re.compile(r"^{.*}$")
 
+# Civis API says these paths for a GET call should return an object,
+# but they actually return an array.
+# See the internal discussion linked through the pull request that added this list.
```
jacksonlee-civis, line 73: I've created an internal ticket for the maintainers of the upstream Civis API to fix this issue.

src/civis/resources/_resources.py
```diff
@@ -582,28 +622,47 @@ def parse_method_name(verb, path):
...
+    if use_legacy_names:
+        # When releasing civis-python v3.0.0, this `if` block can be removed,
+        # leaving the `else` block as the only code path.
+
+        final_elem = path_elems[-1] if path_elems else ""
+        verb = "list" if verb == "get" and (not bracketed(final_elem)) else verb
+    else:
+        verb = (
+            "list" if verb == "get" and get_response_is_array(path, operation) else verb
+        )
```
jacksonlee-civis, line 634: This line `verb = "list" if verb == "get" and (not bracketed(final_elem)) else verb` is the source of the problem that this pull request attempts to fix. Making the distinction of "legacy" versus "preferred" method names here leads to the other changes throughout this `_resources.py` file.

src/civis/base.py
```diff
@@ -193,13 +192,8 @@ def _call_api(
         path=None,
         params=None,
         data=None,
-        deprecation_warning=None,
         **kwargs,
     ):
-        if deprecation_warning:
-            # stacklevel=3 to point to the call just outside civis-python
-            warnings.warn(deprecation_warning, FutureWarning, stacklevel=3)
```
jacksonlee-civis, line 201: Instead of calling `warnings.warn` on our own, I'm wrapping a deprecated, auto-generated API call method using the officially supported `warnings.deprecated` / `typing_extensions.deprecated` decorator [above](https://github.com/civisanalytics/civis-python/pull/530/changes#diff-2cc8d40592739264b17fd8b6095200df09e982e66769e5ce04c447aec39e977aR474).

tests/test_resources.py
```diff
@@ -56,7 +56,7 @@ def test_create_method_iterator_kwarg():
         "get",
         "mock_name",
         "/objects",
-        "deprecation",
+        None,  # no deprecation
```
jacksonlee-civis, line 59: Dropping the deprecation in these tests that aren't actually directly about testing the deprecation warning. Not caught by `pytest.warns`, these deprecation messages would show up in the test report as warnings (though not a test failure). There already exists `test_create_method_deprecation_warning` below (which I'm updating slightly).

src/civis/ml/_helper.py
```diff
@@ -61,7 +61,7 @@ def list_models(job_type="train", author=SENTINEL, client=None, **kwargs):
     template_id_str = ", ".join(str(tmp) for tmp in set(template_id_list))
 
     if author is SENTINEL:
-        author = client.users.list_me().id
```
jacksonlee-civis, line 64: One added benefit of the official `deprecated` decorator is that standard static type checkers also recognize it. For instance, VS Code's Pylance would now nicely mark `client.users.list_me` as deprecated with strikethrough and show the deprecation message:

Image

Relatedly, I've also fixed the type annotations for attributes of the return object of these deprecated `list*` methods, e.g. (cc @mheilman):

Image

Drew-Pappas (approved):
LGTM! Straightforward name replacement with an easy to understand changelog entry for this upgrade

PR #520, opened 2025-07-02
[CIVIS-11072] FIX switch to concurrent.futures API in file multipart upload

jacksonlee-civis:
This pull request updates the file multipart upload implementation to use the `concurrent.futures` API for threading. This change removes the need for explicitly instantiating a "pool" object, thereby (hopefully) resolving #517.

---

- [x] (For Civis employees only) Reference to a relevant ticket in the pull request title
- [ ] Changelog entry added to `CHANGELOG.md` at the repo's root level
- [x] Description of change in the pull request description
- [ ] If applicable, unit tests have been added and/or updated
- [x] The CircleCI builds have all passed

PR #537, opened 2026-04-27
[CIVIS-13165] FIX surface exceptions from a CivisFuture callback

### LINKED ISSUE #536: Exception not being raised from civis.io.civis_to_csv when decompression fails
I am using civis.io.civis_to_csv to sync a table to a file. The decompression is failing yet the future does not raise an exception and when I check `fut.exception()` it is none. Is there something I am doing wrong here? I am running this in a lambda so disk space was the issue. Regardless I would like the exception to be raised so it can be properly handled. Thanks!

Example code:

```python
 fut = civis.io.civis_to_csv(
        file_path, query, database="ZZZ", client=get_civis_api_client()
    )
    fut.result()
    with open(file_path) as f:
        row_count = sum(1 for _ in f) - 1  # subtract header
    logger.info(f"CSV file saved to {file_path} ({row_count} rows)")
```

In the logs I can see that there was an exception but this is not correctly handled and raised to my current thread. I tried adding a check for the exception but still nothing.

```python
    if fut.exception() is not None:
        raise fut.exception()
```

```
exception calling callback for <CivisFuture at 0xffff84d1c6e0 state=succeeded returned Response>
Traceback (most recent call last):
  File "/var/lang/lib/python3.13/concurrent/futures/_base.py", line 340, in _invoke_callbacks
    callback(self)
    ~~~~~~~~^^^^^^
  File "/opt/python/lib/python3.13/site-packages/civis/io/_tables.py", line 1488, in callback
    return _download_file(url, filename, headers, compression)
  File "/opt/python/lib/python3.13/site-packages/civis/io/_tables.py", line 1454, in _download_file
    _decompress_stream(response, fout)
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File "/opt/python/lib/python3.13/site-packages/civis/io/_tables.py", line 1431, in _decompress_stream
    buf.write(d.decompress(to_decompress))
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 28] No space left on device
CSV file saved to /tmp/database_name.table_name.csv (12345678 rows)
```

jacksonlee-civis:
This pull request fixes [#536](https://github.com/civisanalytics/civis-python/issues/536), where a download/decompression error from `civis.io.civis_to_csv` (e.g., out of disk space when running in AWS Lambda) was silently swallowed instead of surfacing through the returned future's `.result()` and `.exception()` methods.

The root cause is in Python's standard library: `concurrent.futures.Future._invoke_callbacks` (and the immediate-fire path in `Future.add_done_callback`) wraps each callback in `try/except Exception` and only logs the error -- so any exception raised inside a done-callback never becomes the future's exception. `civis.io.civis_to_csv` attaches its download step as a done-callback, which is why the reporter's `OSError` was logged but never raised from `fut.result()`.

The fix lives on `CivisAsyncResultBase` (the parent of `CivisFuture`, `ContainerFuture`, `ModelFuture`, and `PollableResult`) so it benefits every future, not just `civis_to_csv`:

- `add_done_callback` and `_invoke_callbacks` now route each callback through a wrapper that captures the first exception onto the future as `_callback_exception`.
- `result()` and `exception()` now block until done-callbacks have finished running and surface any captured callback exception. This also closes a latent race: `Future` notifies waiters *before* invoking callbacks, so previously a caller could see `.result()` return before a callback-driven side effect (e.g., the download) had completed.
- A `threading.get_ident()` check lets a callback call `.result()` on its own future without deadlocking on the new completion event.

Note: this is a behavior change. Exceptions that were previously only logged from done-callbacks will now propagate from `.result()`/`.exception()`. In practice this is the behavior most users expect, and there is no known caller that relies on the silent-swallow behavior.

Only `threading` primitives (`threading.Event`, `threading.get_ident`) are used -- no `multiprocessing` -- so this is safe in AWS Lambda, unlike the issue that motivated #520.

Two regression tests are added in `tests/test_io.py` covering the OSError-surfacing path and the success path.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 ### Deprecated
 ### Removed
-### Fixed
 ### Security
+
+## 2.9.1 - 2026-04-27
+
+### Fixed
+- Exceptions raised by `add_done_callback` callbacks on a `CivisFuture`
+  (and any other `CivisAsyncResultBase` subclass) are now captured onto
+  the future and surfaced through `.result()` and `.exception()`, instead
+  of being silently swallowed by `concurrent.futures.Future._invoke_callbacks`. (#537)
 
 ## 2.9.0 - 2026-04-22
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "civis"
-version = "2.9.0"
+version = "2.9.1"
 description = "Civis API Python Client"
 readme = "README.rst"
 requires-python = ">= 3.11"
--- a/src/civis/base.py
+++ b/src/civis/base.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import logging
 import os
 import platform
 import threading
@@ -12,6 +13,8 @@ import requests
 import civis
 from civis.response import PaginatedResponse, convert_response_data_type, Response
 from civis._retries import retry_request
+
+log = logging.getLogger(__name__)
 
 FINISHED = ["success", "succeeded"]
 FAILED = ["failed"]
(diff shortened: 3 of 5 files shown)


Drew-Pappas (approved):
LGTM!

PR #538, opened 2026-04-27
[CIVIS-13170] FIX support legacy list* endpoints for shares and search

jacksonlee-civis:
The upstream Civis API is going to be updated for the "shares" and "search" endpoints, for similar list*/get* issues as handled in #530. This pull request adds the relevant endpoints to the set of endpoints for which we will need to auto-generate the (preferred) get* methods and mark the list* ones as deprecated. To clarify, as of this moment, these "shares" and "search" aren't updated yet from upstream, but we have to update civis-python first and propagate a new Python client throughout Civis Platform, so that user code still calling list* would not break.

---


src/civis/resources/civis_api_spec.json
jacksonlee-civis, line 1: Updating the API spec and client.pyi as there have been other API updates (unrelated to what this PR aims to handle)

Drew-Pappas (approved):
LGTM!

PR #539, opened 2026-07-28
[CIVIS-13902] DOC Split an API endpoint page into individual method pages

jacksonlee-civis:
The Civis API is documented via the Python client’s Read The Docs site. The API is documented by endpoint, but certain endpoints have a very long HTML page (particularly the `/scripts` endpoint, see https://civis-python.readthedocs.io/en/v2.9.1/api_scripts_endpoint.html), which likely causes issues for AI tools / LLMs to parse and extract info due to token limits etc.

To solve this problem, this pull request breaks up all the methods into their own individual HTML pages, while keeping the existing per-endpoint pages (so that any existing URLs will still work). The “Methods” table in each endpoint page has hyperlinks, which now each point to a new page for each method, instead of using the same endpoint page for all methods.

---


jacksonlee-civis:
@Drew-Pappas Please check out the built site by Read The Docs: https://civis-python--539.org.readthedocs.build/en/539/ -- let me know if you see anything off.

Drew-Pappas (approved):
LGTM!

